PR #303, opened 2015-01-01
Switch to MathJax rendering

skirpichev:


fredrik-johansson:
Output looks good, and it builds much faster :-)

PR #321, opened 2016-01-22
Better error message for int(inf) or int(nan)

asmeurer:
Before:

    >>> int(mpf('inf'))
    ...
    ValueError: cannot convert 0 to int

After:

    >>> int(mpf('inf'))
    ...
    ValueError: cannot convert inf or nan to int


fredrik-johansson:
Very good.

PR #329, opened 2016-08-12
Consistency of log base in error computation

nbruin:
It struck me as odd that the errors D1 and D2, which are clearly base-10 log, get compared on equal footing with D3, which is base-2 log. Is that correct? it would seem the precision is a bit overestimated here (it should hardly ever matter, because by the time D3 is dominant, it doesn't seem we'd be able to do much to get a more accurate result anyway)

### DIFF
--- a/mpmath/calculus/quadrature.py
+++ b/mpmath/calculus/quadrature.py
@@ -196,7 +196,7 @@ class QuadratureRule(object):
             D2 = self.ctx.log(abs(results[-1]-results[-3]), 10)
         except ValueError:
             return epsilon
-        D3 = -prec
+        D3 = -prec*log(10)/log(2)
         D4 = min(0, max(D1**2/D2, 2*D1, D3))
         return self.ctx.mpf(10) ** int(D4)
 


fredrik-johansson:
Shouldn't it be log(2,10) and not log(10,2)? The change apparently causes some tests to hang though.

clslgrnc:
Python complains that is does not know 'log'.
Anyway, wouldn't it be simpler to just put everything in base 2 by replacing 10 by 2 on lines 195, 196, and 201?

skirpichev:
> Python complains that it does not know 'log'.

@clslgrnc, you could use math.log or (better) - a precomputed value of log(2)/log(10).

> Shouldn't it be log(2,10) and not log(10,2)? 

@fredrik-johansson, I think it should be just the `prec_to_dps(prec)` value, isn't?

skirpichev (commented):
This PR seems to be abandoned.  There are also test failures if you use a precomputed value of `log(2,10)` (I didn't run the whole test suite).

But feel free to reopen PR if you wish to continue on this.

PR #332, opened 2016-10-18
Minor edit in functions/bessel.py

paulmasson:
While experimenting with cythonizing all of mpmath, this file produced an error arising from the edited line. Of four instances of `nan` in this file, this one is treated differently.


fredrik-johansson:
Thanks!

PR #334, opened 2016-10-28
lbmpi - mpi_mul by zero

clslgrnc:
My understanding of the documentation is that iv.mpf represent closed intervals of R:
"The iv.mpf type represents a closed interval [a,b]; that is, the set {x:a≤x≤b}"
In particular if b is 'inf', iv.mpf(a,'inf') represents the set {x:a≤x}.
Then [0,0] \* [a,inf] = {0 \* x : a≤x} = {0}.
Contrary to floats where a lot of small numbers collapse to zero, iv.mpf([0,0]) is the singleton 0.
Thus, mpi_mul should return [0,0] if one of its arguments is [0,0].
This would be consistent with the fact that mpi_mul([0, 1], [1, 'inf']) and mpi_mul([-1, 0], [1, 'inf']) currently returns [0, 'inf'] and [-'inf', 0] respectively, and not [-'inf', 'inf'].

### DIFF
--- a/mpmath/libmp/libmpi.py
+++ b/mpmath/libmp/libmpi.py
@@ -136,14 +136,10 @@ def mpi_mul(s, t, prec=0):
     tas = mpf_sign(ta)
     tbs = mpf_sign(tb)
     if sas == sbs == 0:
-        # Should maybe be undefined
-        if ta == fninf or tb == finf:
-            return fninf, finf
+        # for any x in R, 0*x=0
         return fzero, fzero
     if tas == tbs == 0:
-        # Should maybe be undefined
-        if sa == fninf or sb == finf:
-            return fninf, finf
+        # for any x in R, x*0=0
         return fzero, fzero
     if sas >= 0:
         # positive * positive


clslgrnc:
Obviously this makes the following assert fail:
E       assert (mpi('0.0', '0.0') * mpf('+inf')) == mpi('-inf', '+inf')

skirpichev:
> Then [0,0] * [a,inf] = {0 * x : a≤x} = {0}.

Hmm, why is this?  Second set include x==inf and 0*inf != 0.

clslgrnc:
> > Then [0,0] * [a,inf] = {0 * x : a≤x} = {0}.
> 
> Hmm, why is this? Second set include x==inf and 0*inf != 0.

As I said:
> My understanding of the documentation is that iv.mpf represent closed intervals of R

inf is not a real number and thus not included in the second set, if `[0, 0]` represents the singleton {0} and `[a,inf]` is the set of real bigger than `a`: `{x : a≤x}`, then `[0,0] * [a,inf] = {0 * x : a≤x} = {0}`.
If you multiply any real number bigger than a (or any real for that matter) with 0 you get 0.

But then `mpi(inf, inf)̀` in the test suite does not really make sense and should return an error if the intent is to work with closed intervals of reals.

Anyway, I opened this merge request a while ago and do not use mpmath anymore.

Feel free to  close this merge request.

skirpichev:
> [a,inf] is the set of real bigger than a: {x : a≤x}

If you don't include inf - this set is not closed.

Anyway, from [the docs](https://mpmath.org/doc/current/contexts.html#arbitrary-precision-interval-arithmetic-iv) it's clear that interval endings are arbitrary floating-point values, including `inf`.

> Feel free to close this merge request.

I can't.  You can.  Or you can wait for review from the repo owner, if you feel that the pr does make sense.

clslgrnc:
I may be wrong but I feel that you are convinced that this PR does _not_ make sense and are not looking for clarification.
I'll give some anyway, in case you (or anyone else reading this) are interested.

I'll try to be as precise as possible because it seems we are not talking about the same things. Please forgive me if it is too obvious.

First, [interval arithmetic](https://en.wikipedia.org/wiki/Interval_arithmetic) is usually understood as interval arithmetic over the reals and defined as:
```
[a, b] = { x ∈ ℝ ∣ a ≤ x ≤ b }
```
`[a, b]` is the set of real numbers `x` bigger or equal to `a` and smaller or equal to `b`.
As noted in the linked wikipedia page `a` can be `-∞` and `b` can be `∞` resulting in unbounded sets.

Functions over the reals are extended to functions over intervals in the following way:
```
f([a, b], [c, d]) ⊇ { f(x,y) ∣ x ∈ [a, b] and y ∈ [c, d] }
```

Now to your specific comments:

> > [a,inf] is the set of real bigger than a: {x : a≤x}
> 
> If you don't include inf - this set is not closed.
> 

First, if we are talking about intervals over ℝ, inf is not real and cannot be included in the interval.
Second, if we use the usual definition of [closed set](https://en.wikipedia.org/wiki/Closed_set) for metric spaces, this set is closed:
If you take any `x` in the complement of `{ x ∈ ℝ  ∣ a≤x }` then the ball of radius `(a-x)/2 > 0` centered on `x` is included in the complement of `{ x ∈ ℝ  ∣ a≤x }`.

> Anyway, from [the docs](https://mpmath.org/doc/current/contexts.html#arbitrary-precision-interval-arithmetic-iv) it's clear that interval endings are arbitrary floating-point values, including `inf`.
> 

Obviously when doing actual computation on interval arithmetic you need to be able to represent efficiently the endings on a computer.
Floating points, or better, arbitrary precision floating points, are usually a good choice.
The issue is not about the endings but about what is `{ x : a ≤ x ≤ b }` mentionned in the documentation.
From the description it seems that the intent is to use the usual definition of interval arithmetics over real numbers.
If not, this should be clarified.


> > Feel free to close this merge request.
> 
> I can't. You can. Or you can wait for review from the repo owner, if you feel that the pr does make sense.

I am sorry I assumed you were a [collaborator](https://docs.github.com/en/github/setting-up-and-managing-your-github-user-account/permission-levels-for-a-user-account-repository#collaborator-access-for-a-repository-owned-by-a-user-account) on this repository, you are just the second contributor.


Anyway, I feel indeed that keeping this PR open does not make sense:

- It was opened more than four years ago
- The only reactions I got are the ones above
- I do not use mpmath anymore
- This was just a proposal to improve precision and an outdated offer to contribute more on this part of the code

skirpichev:
> you are convinced that this PR does not make sense

That's true.

Returning nan (like Mathematica does) - seems more appropriate.

> not looking for clarification

I hope, not.

> As noted in the linked wikipedia page a can be -∞ and b can be ∞ resulting in unbounded sets.

That means, that ending points are extended reals.

> First, if we are talking about intervals over ℝ, inf is not real and cannot be included in the interval.

But it is included.  See referenced docs or the wikipedia article.

> If you take any x in the complement of { x ∈ ℝ ∣ a≤x } then the ball of radius (a-x)/2 > 0 centered on x is included in the complement of { x ∈ ℝ ∣ a≤x }.

Yes, here I was wrong.  But it seems, mpmath uses definition of intervals, which include `inf`'s.  I.e. - over extended reals.

> I am sorry I assumed you were a collaborator on this repository, you are just the second contributor.

Yes, I'm nobody here)

> It was opened more than four years ago
> The only reactions I got are the ones above

Sorry for that.  But see #550.

clslgrnc:
> > you are convinced that this PR does not make sense
> 
> That's true.
> 
> Returning nan (like Mathematica does) - seems more appropriate.
> 

Maybe you should have led with that instead of asking question you were not interested in.

> > not looking for clarification
> 
> I hope, not.
> 
> > As noted in the linked wikipedia page a can be -∞ and b can be ∞ resulting in unbounded sets.
> 
> That means, that ending points are extended reals.
> 

Almost, you may notice that they do not allow a to be ∞  or b to be -∞


> > First, if we are talking about intervals over ℝ, inf is not real and cannot be included in the interval.
> 
> But it is included. See referenced docs or the wikipedia article.
> 

You are aware of reals and extended reals, yet you insist on the fact that ∞ is in a subset of ℝ.
If I wrote A = { n ∈ ℕ ∣ 2 ≤ n ≤ 3.5 } would you say that 3.5 is in A ?

Obviously the comparison is done over extended reals, but x is taken in ℝ as written in the usual definition:
`[a, b] = { x ∈ ℝ ∣ a ≤ x ≤ b }`

> > If you take any x in the complement of { x ∈ ℝ ∣ a≤x } then the ball of radius (a-x)/2 > 0 centered on x is included in the complement of { x ∈ ℝ ∣ a≤x }.
> 
> Yes, here I was wrong. But it seems, mpmath uses definition of intervals, which include `inf`'s. I.e. - over extended reals.
> 

This is unusual and should be in the documentation.

> > I am sorry I assumed you were a collaborator on this repository, you are just the second contributor.
> 
> Yes, I'm nobody here)
> 
> > It was opened more than four years ago
> > The only reactions I got are the ones above
> 
> Sorry for that. But see #550.

Congratulation I guess.

Maybe you could have just said that you recently took over the project, did not have the permission yet to close pull requests, and ask me to close it as you were not interested. That would have felt more welcoming.

skirpichev:
> ask me to close it as you were not interested

That's not true.

clslgrnc:
My bad, I misunderstood your comments then.
Let's try to have a more constructive conversation in #591.

Issue #339, opened 2016-12-16
findroot bisect solver should dynamically set maxsteps


asmeurer:
The default `maxsteps` for the bisect solver is not high enough to find roots for higher precision. Consider the following:

```
In [37]: import mpmath as mp

In [38]: mp.mp.dps = 15

In [39]: mpmath.findroot(lambda x: mp.cos(x) - x, [0, 1], solver='bisect')
Out[39]: mpf('0.73908513321516064')
```

Works fine, but

```
In [40]: mp.mp.dps = 100

In [41]: mpmath.findroot(lambda x: mp.cos(x) - x, [0, 1], solver='bisect')
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-41-3f13b0529a9a> in <module>()
----> 1 mpmath.findroot(lambda x: mp.cos(x) - x, [0, 1], solver='bisect')

/Users/aaronmeurer/anaconda3/lib/python3.5/site-packages/mpmath/calculus/optimization.py in findroot(ctx, f, x0, solver, tol, verbose, verify, **kwargs)
    973                              '(%g > %g)\n'
    974                              'Try another starting point or tweak arguments.'
--> 975                              % (norm(f(*xl))**2, tol))
    976         return x
    977     finally:

ValueError: Could not find root within given tolerance. (1.00455e-63 > 1.39525e-104)
Try another starting point or tweak arguments.
```

With a little bisection, I find that the `maxsteps` needs to be at least 172 for 100 digits of precision:

```

In [42]: mpmath.findroot(lambda x: mp.cos(x) - x, [0, 1], solver='bisect', maxsteps=171)
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-42-196333f2d7bd> in <module>()
----> 1 mpmath.findroot(lambda x: mp.cos(x) - x, [0, 1], solver='bisect', maxsteps=171)

/Users/aaronmeurer/anaconda3/lib/python3.5/site-packages/mpmath/calculus/optimization.py in findroot(ctx, f, x0, solver, tol, verbose, verify, **kwargs)
    973                              '(%g > %g)\n'
    974                              'Try another starting point or tweak arguments.'
--> 975                              % (norm(f(*xl))**2, tol))
    976         return x
    977     finally:

ValueError: Could not find root within given tolerance. (5.31455e-104 > 1.39525e-104)
Try another starting point or tweak arguments.

[...]

PR #341, opened 2016-12-29
Use %s instead of %g in findroot error

asmeurer:
This fixes incorrect cutoff for tolerances that don't fit in a Python float.

This causes the error message to be longer, especially when the precision is
set high. Perhaps there is a better way to only print a few digits without
using %g.

Fixes an issue mentioned in #339.


asmeurer:
I fixed the failing test.

PR #348, opened 2017-02-14
Add numerical inverse Laplace transform functionallity

klkuhlm:
This branch implements three common numerical inverse Laplace transform algorithms (Talbot, Stehfest, and de Hoog, Knight & Stokes). The default algorithm is de Hoog. Algorithms have been "tuned" based on testing a suite of typical functions with known inverses (Table 9.4 of Cohen, 2007).

Doctests and main tests pass.


fredrik-johansson:
There is a small markup bug in the docstring for invertlaplace: the double colon on line 546 should be a single colon for the following equation to render correctly.

This looks good, and I'll be happy to merge it. Feel free to add a credit line for yourself in the source file and/or the documentation page.

Issue #349, opened 2017-02-24
Inconsistency in hyper() evaluation with "eliminate" keyword enabled


skirpichev:
Lets take an example:
```python
In [6]: mpmath.__version__
Out[6]: '0.19'

In [7]: mpmath.hyper((-1, 1), (-1,), 2)
Out[7]: mpf('-1.0')

In [8]: mpmath.hyper((-1, 1), (-1,), 2, eliminate=False)
Out[8]: mpf('3.0')
```

Second result also equivalent to the Mathematica output:
```
In[1]:= HypergeometricPFQ[{-1, 1}, {-1}, 2]

Out[1]= 3
```

@fredrik-johansson, please correct me, but I believe the result must be same with or without eliminating common parameters.  Is that wrong?

This seems to be an issue for SymPy and Diofant.  See sympy/sympy#10531 (real technical discussion was started only [here](https://github.com/sympy/sympy/pull/10531#issuecomment-281806140)) and diofant/diofant#446.

fredrik-johansson:
This is the very purpose of "eliminate" option. Eliminating common parameters is only unambiguous when the parameters are not nonpositive integers. At nonpositive integers, there are two possible interpretations, and mpmath allows you to choose.

Note that Mathematica is inconsistent with itself about which rule to use:

    In[1]:= HypergeometricPFQ[{-1, 1}, {-1}, 2]                                    
    Out[1]= 3
    In[2]:= HypergeometricPFQ[{a, 1}, {a}, 2] /. a -> -1                           
    Out[2]= -1

fredrik-johansson:
> @fredrik-johansson, this seems to be very hackish way of choosing different hyper() definitions.

I don't see what is hackish about it.

> In this case, shouldn't ``eliminate=False`` - be default?

It would probably be fine to change it, though in rare cases it will break compatibility. There are some functions like Jacobi and Gegenbauer polynomials, incomplete beta function, etc. where one would have to check that the expected results are obtained.

It was made default for performance reasons, but it could be changed to eliminate only safe parameters by default, with eliminate=True/False simply controlling elimination of nonpositive integers. Or maybe there should be a completely separate flag for nonpositive integers.

> You also have a documentation issue at least. (eliminate kwarg is not documented in docstring).

True.

skirpichev:
FYI: it seems for me that @fredrik-johansson was positive about technical changes in #350, related to this bug.  Anyone willing to fix this issue?

[...]

PR #350, opened 2017-02-24
[wip] Set eliminate=False per default in hyper()

skirpichev:
Not sure how common cancellation is to affect performance badly.  Hence, here is a simplest solution for #349.

Except one test, nothing seems to be affected in your test suite.  But please note, that moderate amount of docstrings - not tested.

Also readme update about Code of Conduct.

### DIFF
--- a/mpmath/function_docs.py
+++ b/mpmath/function_docs.py
@@ -2816,9 +2816,9 @@ series terminates before the division by zero occurs)::
     >>> hyper([1,1,1,-1],[-2,5],1)
     1.1
 
-With the keyword argument ``eliminate=True``, function degree reduced
+With the keyword argument ``eliminate=True``, the function degree is reduced
 first by cancellation of common parameters.  For nonpositive parameters
-that may lead to the different function definition::
+that may lead to a different function definition::
 
     >>> hyper([-1, 1],[-1],2)
     3.0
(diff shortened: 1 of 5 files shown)


fredrik-johansson:
Could you please revert the addition to the README? I don't think it needs to be said that contributions will be judged by their technical merit, and a link to someone's opinion piece on codes of conduct (even if I largely agree with the sentiment behind it) is plainly off topic.

fredrik-johansson:
The performance issue matters, but I can easily fix it later if you don't want to bother with it.

mpmath/function_docs.py
```diff
@@ -2816,6 +2816,15 @@
     >>> hyper([1,1,1,-1],[-2,5],1)
     1.1
 
+With the keyword argument ``eliminate=True``, function degree reduced
```
fredrik-johansson: "the function degree is"

mpmath/function_docs.py
```diff
@@ -2816,6 +2816,15 @@
...
+first by cancellation of common parameters.  For nonpositive parameters
+that may lead to the different function definition::
```
fredrik-johansson: "to a different"

skirpichev:
> I don't think it needs to be said that contributions will be judged by their technical merit

So, you wont formalize your words (quoted in the commit message) in project docs, in any from?

fredrik-johansson:
> So, you wont formalize your words (quoted in the commit message) in project docs, in any from?

I don't think it belongs.

skirpichev:
Ok, sorry.  Lets wait if someone else will come with better solution.

PR #360, opened 2017-06-18
Fix rendering of brackets in math

paulmasson:
This set of equations doesn't render properly in Sphinx with literal brackets. Technically only the left brackets need to be replaced, but I changed them all for consistency.


fredrik-johansson:
Thanks.

PR #362, opened 2017-07-02
Clarify documentaton for re() and im()

paulmasson:


paulmasson:
The documentation was unclear to me here, since `mpc` has attributes `real` and `image`. Hopefully the clarification is what was originally intended.

PR #364, opened 2017-07-27
fredrik-johansson/mpmath#344

abrys:
Rewrite the python major version selection to make it work if something else has redefined xrange.


fredrik-johansson:
Yes, this looks cleaner.

PR #369, opened 2017-09-21
Add an optional argument to polyroots() function

kagalenko-m-b:
Sometimes polyroots() fail to converge, but lower-precision estimates of the roots are available. This proposed change lets a user to supply an array of initial guesses of the roots to be refined by the Durand–Kerner method.


fredrik-johansson:
This looks like a sensible addition, but the code is broken. Please add a test case.

kagalenko-m-b:
The test I added to mpmath/tests/test_calculus.py  seems to be working fine.

fredrik-johansson:
Seems ok. Thanks!

Issue #394, opened 2018-06-04
Tests fail with Python 3.7


casevh:
When run the test suite with Python 3.7, the following error occurs:

> qfunctions                
> TEST FAILED!
> 
> Traceback (most recent call last):
>   File "/home/case/src/mpmath-1.0.0/mpmath/functions/qfunctions.py", line 126, in factors
>     raise StopIteration
> StopIteration
> 
> The above exception was the direct cause of the following exception:
> 
> Traceback (most recent call last):
>   File "runtests.py", line 126, in testit
>     module.__dict__[f]()
>   File "/home/case/src/mpmath-1.0.0/mpmath/tests/test_functions2.py", line 1449, in test_qfunctions
>     assert qp(2,3,100).ae('2.7291482267247332183e2391')
>   File "/home/case/src/mpmath-1.0.0/mpmath/functions/qfunctions.py", line 129, in qp
>     return ctx.mul_accurately(factors)
>   File "/home/case/src/mpmath-1.0.0/mpmath/ctx_base.py", line 391, in mul_accurately
>     for factor in factors():
> RuntimeError: generator raised StopIteration

The failure is due to implementation of PEP 479 which changes the behavior of StopIteration inside a generator. The new behavior was first available in Python 3.5 via an import from \_\_future\_\_. It became the default in Python 3.7.

The fix is easy: change the line `raise StopIteration` in qfunctions.py to just `return`. I've tested the change with Python 2.7, 3.5, 3.6, and 3.7. There are no other occurrences of `raise StopIteration`.

PR #398, opened 2018-07-20
Fix PEP-479 compatibility (python 3.7)

keszybz:
'raise StopIteration' is turned into a RuntimeError under python3.7
when the generator function stack is exited because of it (and causes
a warning under python3.6). It is equivalent to plain 'return', which
works under both python2.x and python3.x.

### DIFF
--- a/mpmath/functions/qfunctions.py
+++ b/mpmath/functions/qfunctions.py
@@ -123,7 +123,7 @@ def qp(ctx, a, q=None, n=None, **kwargs):
             r *= q
             k += 1
             if k >= n:
-                raise StopIteration
+                return
             if k > maxterms:
                 raise ctx.NoConvergence
     return ctx.mul_accurately(factors)


fredrik-johansson:
@skirpichev yes, that is a good idea

skirpichev:
this fixes #394

But single doctest fail on py3.7:
```
034     | ``mpc``    | Complex float  |
035     +------------+----------------+
036     | ``matrix`` | Matrix         |
037     +------------+----------------+
038 
039 The following section will provide a very short introduction to the types ``mpf`` and ``mpc``. Intervals and matrices are described further in the documentation chapters on interval arithmetic and matrices / linear algebra.
040 
041 The ``mpf`` type is analogous to Python's built-in ``float``. It holds a real number or one of the special values ``inf`` (positive infinity), ``-inf`` (negative infinity) and ``nan`` (not-a-number, indicating an indeterminate result). You can create ``mpf`` instances from strings, integers, floats, and other ``mpf`` instances:
042 
043     >>> mpf(4)
Expected:
    mpf('4.0')
Got:
    4.0
```

fredrik-johansson:
It fails only on py3.7? Is there a difference in the test order that leads to mp.pretty being set?

skirpichev:
> It fails only on py3.7?

On py3.6 too.

skirpichev:
Oh, I see
```
doc/build/_sources/basics.txt .                                          [  0%]
doc/build/_sources/contexts.txt .                                        [  0%]
doc/build/_sources/matrices.txt .                                        [  0%]
doc/build/_sources/setup.txt .                                           [  0%]
doc/build/_sources/technical.txt .                                       [  1%]
doc/build/_sources/calculus/inverselaplace.txt .                         [  1%]
doc/build/_sources/functions/constants.txt .                             [  1%]
doc/build/_sources/functions/gamma.txt .                                 [  1%]
doc/source/basics.txt F                                                  [  1%]
```
Probably, that was orphaned documentation build in the source tree.  Sorry for noise, I seems tests pass on py3.6 and py3.7.

PR #403, opened 2018-08-14
Fix several invalid escapes

asmeurer:
Future versions of Python will make invalid escapes in strings like `'\s'` a syntax error if the string isn't a raw string. This makes several docstrings in mpmath with backslashes raw. I didn't check but potentially some of these even had valid escapes, meaning things like `\t` or `\n` which would just produce the corresponding whitespace character in the parsed string. 

These cause issues for upstream SymPy, as we have our tests set to fail if there are invalid escapes. See https://github.com/sympy/sympy/pull/15094. 

I am also setting up testing of 3.6 and 3.7 in Travis, and testing of invalid escapes there so they won't reappear.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,7 +15,7 @@ install:
 script:
   # Guard against invalid escapes in strings, like '\s'. If this fails, a
   # string or docstring somewhere needs to be a raw string.
-  - if [ "$TRAVIS_PYTHON_VERSION" == "3.6" ];
+  - if [ "$TRAVIS_PYTHON_VERSION" == "3.6" ]; then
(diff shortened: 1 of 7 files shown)


asmeurer:
Worth noting here that the compileall test is a little non-user friendly. It doesn't stop when there is an error, so if the command fails, you may have to scroll up to see it. Unfortunately, it's the best way I know to actually test this.

asmeurer:
I guess 3.7 doesn't actually work in Travis yet (I had thought that might be the case).

fredrik-johansson:
Thanks!

PR #409, opened 2018-10-06
Householder reflections for complex matrices

kagalenko-m-b:
This corrects bug in householder() function that resulted in incorrect solution of linear systems with complex left hand side matrix.


fredrik-johansson:
Great, thanks!

PR #410, opened 2018-10-07
Improved support for numerical types

warnerjon12:
I updated the convert function to work with numpy's numerical types as well as Fraction and Decimal. I added a test and all tests passed on Python 2.7 and 3.5. Also fixed an issue with findroot


skirpichev:
It would be better, if you will do changes on the dedicated branch, not in master.  Also, unrelated changes, probably, should go to different pull requests.

mpmath/libmp/libmpf.py
```diff
@@ -426,6 +426,37 @@ def from_float(x, prec=53, rnd=round_fast):
...
+    """Create a raw mpf from a Decimal, rounding if necessary.
+    If prec is not specified, use the equivalent bit precision
+    of the number of significant digits in x."""
+    if x.is_nan(): return fnan
+    if x.is_infinite(): return fninf if x.is_signed() else finf
+    sgn, digs, e = x.as_tuple()
+    if prec is None:
+        prec = int(len(digs)*math.log(10, 2))
+    n = int(('-' if sgn > 0 else '').join(list(map(str, digs))))
+    if e < 0: d = 10**-e
```
fredrik-johansson: This is bad for numbers with large exponents. Just use from_str instead.

mpmath/ctx_mp_python.py
```diff
@@ -365,8 +365,8 @@ class behaves fairly similarly to Python's complex type.
 
     def __new__(cls, real=0, imag=0):
         s = object.__new__(cls)
-        if isinstance(real, complex_types):
-            real, imag = real.real, real.imag
+        if isinstance(real, complex_types) or isinstance(real, numbers.Complex):
+            real, imag = real.real, (real.imag if imag == 0 else imag)
```
fredrik-johansson: What is the logic here?

I think the extra comparison with 0 is going to needlessly slow down the common case. Perhaps make a separate if-case for the isinstance(real, numbers.Complex) case if this is needed.

The new isinstance call will also slow down the common case where the input is real, but that perhaps hard to avoid if we want to support this feature.

warnerjon12:
As was mentioned, I will need to do some branching and cleanup before pulling this in. As such, I am closing this to reopen from a different branch.

Issue #382, opened 2018-01-12
mp.mpf(numpy.int64(1)) fails with Python 3


nschloe:
```python
from mpmath import mp
import numpy

mp.sqrt(numpy.int64(1))
```
fails with
```
TypeError: cannot create mpf from 1
```
under Python 3 (not 2). A similar bug exists in sympy, https://github.com/sympy/sympy/issues/13618.

warnerjon12:
Finished and pushed. Pull request in.

warnerjon12:
I am reorganizing my branches, will put in a new pull request for this shortly.

PR #412, opened 2018-10-14
Improved numpy type support

warnerjon12:
The convert function now supports numpy types, Fraction, and Decimal. Fixes  issue #382.


fredrik-johansson:
The code needs to work even if numpy is not available.

In the test code, you can put

    try:
        ...
    except Import Error:
        pass

around the code requiring numpy.



from_npfloat looks improvable. I think we should do something like y = float(x), if y == x: return from_float(y, ...) and only otherwise fall back to code requiring long ints. The test code doesn't look strong enough -- you want to check that the precision is really preserved.


The new convert is problematic. First, as noted above, you can't import numpy here.

Second, we shouldn't do x = np.asscalar(x). We *don't* want to silently convert arrays to scalars.

Next, can we check for NumPy floating and complex types in some way without importing numpy, for instance by looking for some numpy-only attribute?

I would also prefer a way detect Decimals that does not require importing the decimal module.

A final comment about this method is that these conversions should be further down, to avoid slowing down the common cases of converting from Python int, float and complex types.

warnerjon12:
FYI, np.asscalar(x) will not convert arrays to scalars if x is a scalar. This was checked with isinstance(x, np.generic). Although this would have been a problem with float96 or float128 potentially. In any case, I made some changes.

asmeurer:
You might have a look at what SymPy does https://github.com/sympy/sympy/blob/99c70916a4e6d6a827a55b3e4d07151e9120e935/sympy/core/sympify.py#L54

warnerjon12:
Yeah, but asscalar has the potential to return an array for the bigger types on some platforms I believe. The current method should work fine. I am moving on to bigger things now. Hence my new branch on my fork

warnerjon12:
> What is the purpose of the
> 
> ```
> try: return ctx.convert(str(x))
> ```
> 
> fallback? This looks too fragile; we should be explicit about what gets (or doesn't get) converted.

Yeah, that was intended as a fallback. In theory it should never execute, but you never know. The other option would be to raise an exception there without falling back.

fredrik-johansson:
> The other option would be to raise an exception there without falling back.

Please do that instead!

warnerjon12:
Are we waiting to merge until some other stuff is done, or what?

warnerjon12:
Done and ready for testing

fredrik-johansson:
OK, it looks reasonable now. Thanks!

PR #423, opened 2019-01-03
Optimized python_trailing function

adhoc-king:
Included a look-up table which reduces time by 50% for numbers which have a non-zero bit in the lowest byte

### DIFF
--- a/mpmath/libmp/libintmath.py
+++ b/mpmath/libmp/libintmath.py
@@ -12,6 +12,9 @@ from bisect import bisect
 from .backend import xrange
 from .backend import BACKEND, gmpy, sage, sage_utils, MPZ, MPZ_ONE, MPZ_ZERO
 
+small_trailing = [i and max(int(not i % 2**j) and j for j in range(1, 8))
+    for i in range(256)]
+
 def giant_steps(start, target, n=2):
     """
     Return a list of integers ~=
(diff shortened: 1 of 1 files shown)


fredrik-johansson:
Wow, nice find! I can't believe I overlooked such a crucial optimization when writing this code.

Could you make a few further improvements?

To precompute the table, the following is 140 times faster:

    small_trailing = [0] * 256
    for j in range(1,8):
        small_trailing[1<<j::1<<(j+1)] = [j] * (1<<(7-j))

In the fallback loop, instead of taking 1 bit at a time, the loop could mask out 8 bits at a time until the lookup table can be used.

Actually, this is still an O(n^2) algorithm for powers of two which isn't really ideal. I wonder if using hex() in the fallback case would be better?

fredrik-johansson:
Yes, exactly. (Perhaps a micro-optimization you could also start with t = 8, n >>= 8 since the first loop iteration always succeeds.)

It's O(n^2) because each n >>= 8 operation is O(n) (relevant only for huge numbers).

fredrik-johansson:
Of course, O(n) and O(n^2) should be O(log(n)) and O(log(n)^2) above.

adhoc-king:
Made a pull request #430 
Thanks a lot! I learnt many things :)

PR #425, opened 2019-01-09
Fixed scalar-matrix multiplication in iv module

warnerjon12:


warnerjon12:
I had accidentally rebased one commit too far, so the history shows an extra commit. I am not sure how to remove me from that while keeping the commit, would be nice to know.

fredrik-johansson:
Seems ok.

fredrik-johansson:
Reverted because of failing tests. I don't get why -- this PR was green before the merge.

warnerjon12:
I will have to look at that when I have more time

PR #426, opened 2019-01-09
Implemented float and complex conversions for ivmpf

warnerjon12:
Response to issue fredrik-johansson#266


fredrik-johansson:
Thanks!

PR #436, opened 2019-02-04
Restore support for CPython 3.4

### LINKED ISSUE #434: Restore Python 3.4 support?
We just dropped Python 3.4 support since Travis was having issues with it, but a user has pointed out in private communication that mpmath is being actively used in some environments that only have Python 3.4. We shouldn't need any Python >= 3.5 features, so it's just a question of testing compatibility.

skirpichev:
Closes #434

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,6 @@ language: python
 sudo: false
 python:
   - 2.7
-  - 3.4
   - 3.5
   - 3.7-dev
   - pypy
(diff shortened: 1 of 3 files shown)


skirpichev:
@fredrik-johansson, please restart this build several times.

fredrik-johansson:
Thanks!

skirpichev:
@fredrik-johansson, this may be a wrong solution.  To be sure it's working - you had to restart travis build for several times.

Issue #443, opened 2019-02-26
mpmath should not hard depend on pkg_resources


asmeurer:
It looks like https://github.com/fredrik-johansson/mpmath/pull/422 added a hard dependency on pkg_resources. 

I think this is a bad idea. It forces a dependency on setuptools. pkg_resources is also known to have a lot of issues. 

I would recommend using something like versioneer if you want to have a version based on git tags.

asmodehn:
To add a concrete use-case where that is a problem : I m trying to import `mpmath` from source (a cloned worktree on my filesystem, via sys.path hacking, to easily work on my project, yet also be able to modify my fork of mpmath), and I get : 
```
  File "/home/alexv/Projects/crypy/crypy/euc/__init__.py", line 51, in <module>
    import mpmath
  File "/home/alexv/Projects/crypy/submodules/mpmath/mpmath/__init__.py", line 2, in <module>
    __version__ = pkg_resources.get_distribution(__name__).version
  File "/home/alexv/.local/share/virtualenvs/crypy-PletnYfd/lib/python3.7/site-packages/pkg_resources/__init__.py", line 479, in get_distribution
    dist = get_provider(dist)
  File "/home/alexv/.local/share/virtualenvs/crypy-PletnYfd/lib/python3.7/site-packages/pkg_resources/__init__.py", line 355, in get_provider
    return working_set.find(moduleOrReq) or require(str(moduleOrReq))[0]
  File "/home/alexv/.local/share/virtualenvs/crypy-PletnYfd/lib/python3.7/site-packages/pkg_resources/__init__.py", line 898, in require
    needed = self.resolve(parse_requirements(requirements))
  File "/home/alexv/.local/share/virtualenvs/crypy-PletnYfd/lib/python3.7/site-packages/pkg_resources/__init__.py", line 784, in resolve
    raise DistributionNotFound(req, requirers)
pkg_resources.DistributionNotFound: The 'mpmath' distribution was not found and is required by the application
```
because the package has not been installed.
I havent tested but I'd probably get the same error, if I just cloned the repository on the side and open it with my IDE (Pycharm), which will add it to the PYTHONPATH.

So in my usecase, I need a way to import directly from source, that means a way to get a version number without installing, without involving setuptools at all.

I *think* this means that the version number should be in the source (as it was before #422), to minimize dependencies to other systems around (installer, version control, etc.)

fredrik-johansson:
I'd be happy to have it in the source again.

[...]

PR #446, opened 2019-03-08
reverting to use a static version string in code

asmodehn:


asmodehn:
Attempting to address #443

fredrik-johansson:
I think I'd rather just have the static __version__ without any magic at all.

asmodehn:
Done.

asmeurer:
Should the changes to setup.py from https://github.com/fredrik-johansson/mpmath/pull/422/files be reverted too?

fredrik-johansson:
It makes no difference to me how setup.py is written as long as it works. I believe that commit did fix some of the metadata, so better to patch it as needed than to revert, unless it's somehow broken in other ways now.

asmeurer:
setup.py does require setuptools now, perhaps unnecessarily. That could matter, though it isn't nearly as big of a deal as the library runtime depending on it.

asmodehn:
IMHO dependency on setuptools is not a big issue. however the fact that
there is now effectively 2 versions (the string in code and the git tag) is
likely to become a problem eventually...

asmeurer:
I've maintained SymPy with the version duplicated in the tag and the code for years with no problems. You just remember to update the version in the code when you release, and then git tag after making the release.

On the other hand, a run dependency on setuptools creates many problems. It also forces upstream projects like SymPy to depend on setuptools by proxy, which we would like to avoid. 

If you want to version based on tags, there are projects like versioneer, which I use myself in several projects, which let you do this without any additional runtime dependencies. The main advantage is getting git version version numbers that include the commit hash. The actual nonduplication of the version number isn't a huge deal. The advantage to having them separate is you don't actually tag until after the release is made (though that generally isn't a big deal as tags aren't "permanent" until you push them up).

skirpichev:
> I've maintained SymPy with the version duplicated in the tag and the code for years with no problems.

And I have seen several project, where duplicated versions were out of sync...  Maybe even the mpmath in past.  This only means, if you did no error yet - you will do error, eventually.

> On the other hand, a run dependency on setuptools creates many problems.

For example?  As I said in the issue thread - setuptools is the recommended tool for package management in the python ecosystem.  Using setuptools allows simple tracking of hard and optional requirements and many users are already familiar with pip/setuptools (less installation instructions, etc).

> It also forces upstream projects like SymPy to depend on setuptools by proxy, which we would like to avoid.

Why at all you should care about dependencies of the project, which you require?  This may harm SymPy, if it will install own requirements, which has conflicts with other requirements of SymPy itself (pip will prevent such a situation).  But I fail to see how setuptools may introduce such a conflict...

PS: SymPy is a good example of bad practice, where you fork almost everything you depend on.  Should mpmath encourage this?  I doubt.

PR #453, opened 2019-03-18
Improve slicing and add matrix multiply

tchen01:
The general theme of this pull request is adding some functionality to make the syntax more similar to numpy. Hopefully this will make it easier to port code from numpy to mpmath. 

I don't believe that any of the changes implemented here will change any of the behaviour of existing code. That is, all operations arithmetic operations will be performed in exactly the same order. 

- Add `__matmul__` method with multiply which takes advantage of sparsity
  - Note that since the order of some of the scalar multiplications is changed the exact output will vary slightly from the original `__mul__` method but the norms of the matrices should not differ by more the machine precision times the product of the norms of the matrices.

- Update `__getitem__` and `__setitem__` methods to allow indexing with boolean and integer lists
  - I have simplified the syntax for the case handling 
  - Note that using a slice which is out of bounds will behave like slicing a python list (i.e. it gets mapped in bounds). So just as `[1,2,3,4][2:10]` returns  `[3,4]`, `mp.matrix([1,2,3,4])[2:10]` will return a matrix with the elements 3 and 4. This means one test from the matrix tests will fail since it wants this to raise an error.
  - if the `key` provided is not a tuple, then these functions check if you are working with a vector, in which case it assumes you want to slice along the non-trivial axis. Otherwise it will assume the rows should be sliced and take all columns

 - add `.shape` property

Note: `test_quad.py` fails, but it failed before I made any changes (as far as I can tell).

### DIFF
--- a/mpmath/matrices/matrices.py
+++ b/mpmath/matrices/matrices.py
@@ -437,18 +437,18 @@ class _matrix(object):
         
         '''
         if isinstance(key_1d,int):
-            return xrange(key_1d,key_1d+1)
+            return xrange(key_1d%depth,key_1d%depth+1)
         elif isinstance(key_1d,slice):
             return xrange(*key_1d.indices(depth))
         elif isinstance(key_1d,list):
             if isinstance(key_1d[0],bool):
                 if len(key_1d) != depth:
                     raise IndexError('boolean mask does not match dimension size')
-                return [i for i,x in enumerate(key_1d) if x] # return list so that we can take length late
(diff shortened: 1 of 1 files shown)


tchen01:
I closed this pull request because it does not work if the user indexes with a numpy array or a numpy integer type. Strictly speaking, this doesn't break the specs for mpmath, but my guess is that a lot of users have code where they might directly use the output from something like `np.where` to access an element from a mpmath matrix. As such, I have closed this until I can think about this a bit more.

PR #454, opened 2019-03-20
Enable testing on CPython 3.8

skirpichev:


mpmath/ctx_mp_python.py
```diff
@@ -889,7 +889,7 @@ def fsum(ctx, terms, absolute=False, squared=False):
             s = ctx.make_mpc((s, mpf_sum(imag, prec, rnd)))
         else:
             s = ctx.make_mpf(s)
-        if other is 0:
+        if other == 0:
```
skirpichev: Hmm, I think that all logic, related to the `other` local variable - is inaccessible.  Same for fdot function.  @fredrik-johansson, does this code assume anything, except mpf/mpc?

skirpichev:
Beware, the last commit - remove the mentioned inaccessible logic.

fredrik-johansson:
Thanks!

PR #462, opened 2019-07-25
Polynomial doc rendering

sylee957:
http://mpmath.org/doc/current/calculus/polynomials.html
Polynomial doc is using inline latex for some strings.
I've changed them to inline literals.

image

image


fredrik-johansson:
Thanks!

PR #468, opened 2019-08-21
Prevent division by zero

thepith:
Problem: `rnd.getrandbits` can result in 0, so that `b` could equal to 0.

This fix makes it that b is a pseudorandom number close, but not equal,
to zero.


fredrik-johansson:
Thanks!

PR #470, opened 2019-09-13
WIP: Python 3 rounding

cbm755:
Python 2 rounding always made floats but Python 3 has more general
semantics [1].  To take advantage of those, we have `__round__` return
an `mpf` intead of a `float`.

[1] https://docs.python.org/3/library/functions.html#round

### TODO List
- [x] basic implementation(s)
- [x] some basic tests
- [ ] should we always return an mpf or out an `int` or `mpz` be appropriate?
- [ ] add a Python 2 fallback
- [ ] more tests
- [ ] add `mpc.__round__`
- [ ] support other rounding modes.  Is there something like `mp.dps` but for rounding mode?  I notice each mpf knows its rounding mode inside `._ctxdata` but where do I set that?

### algorithms

I've tried two rather embarrassing implementations but not happy with either.  Do we need to modify/generalize `mpf_round_int`?

Help very much welcome!

### DIFF
--- a/mpmath/ctx_mp_python.py
+++ b/mpmath/ctx_mp_python.py
@@ -228,8 +228,50 @@ class _mpf(mpnumeric):
     def to_fixed(self, prec):
         return to_fixed(self._mpf_, prec)
 
-    def __round__(self, *args):
-        return round(float(self), *args)
+    def __round__(self, n=0):
+        """
+        Round to `n` digits, respecting current rounding flags and precision.
+
+        The Python docs [1] say:
+        > For the built-in types ... toward the even choice
+
+        Thus, as we are not a built-in type, we can round using the current
+        rounding flag.
+
+        This implementation always returns an `mpf` object.
+        TODO: would we like `round(foo, 0)` to return `int`?
+
+        [1] https://docs.python.org/3/library/functions.html#round
+        """
+        if False:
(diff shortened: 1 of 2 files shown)


mpmath/ctx_mp_python.py
```diff
@@ -228,8 +228,50 @@ def ae(s, t, rel_eps=None, abs_eps=None):
...
+        Round to `n` digits, respecting current rounding flags and precision.
+
+        The Python docs [1] say:
+        > For the built-in types ... toward the even choice
+
+        Thus, as we are not a built-in type, we can round using the current
+        rounding flag.
+
+        This implementation always returns an `mpf` object.
+        TODO: would we like `round(foo, 0)` to return `int`?
```
skirpichev, line 242: "The return value is an integer if ndigits is omitted or None. Otherwise, the return value has the same type as number." (c) CPython
But I think we also can ignore this:)

skirpichev:
> add a Python 2 fallback

No need.

> I notice each mpf knows its rounding mode inside ._ctxdata but where do I set that?

I don't think there is a public interface for  `_prec_rounding[1]`.

skirpichev:
There are test failures and no progress for a long time.

Feel free to reopen the pr if you would like to continue this work.

PR #492, opened 2019-11-19
Travis CI: Use production versions of CPython, pypy, pypy3

cclauss:
Remove the hardcoding of out of date versions


fredrik-johansson:
Seems OK.

PR #519, opened 2020-03-12
elliprc has incorrect imaginary part for negative y-axis

paulmasson:
Comparing values above and below the negative y-axis,

>>> elliprc(-1,-2+.001j,pv=False)
mpc(real='0.00014269906542846742', imag='-0.78539811887314348')
>>> elliprc(-1,-2,pv=False)
mpc(real='0.0', imag='0.78539816339744828')
>>> elliprc(-1,-2-.001j,pv=False)
mpc(real='0.0014280962796194981', imag='2.3561933566202637')

the imaginary part on the axis has the wrong sign. This pull request should fix that along with minor cleanup.


fredrik-johansson:
Travis reports test failures.

paulmasson:
Not understanding what is failing here. The branch runs fine on my computer and returns the expected result, now that the sign has been reversed.

paulmasson:
Finally figured out where the Travis error was and fixed that test. The [line after](https://github.com/fredrik-johansson/mpmath/blob/master/mpmath/functions/elliptic.py#L810) has a comparison to direct integration that appears to be wrong because the choice of contour changes the sign of the imaginary part. Since I could see no justification for which choice to make, I assume you did it to match the existing answer for the Carlson integral. If there is another reason, please let me know.

There are no longer Travis errors for this PR, but there is some other code coverage error I do not understand. Perhaps it is not important.

skirpichev:
Fork was removed, unfortunately.

PR #525, opened 2020-03-20
Support @ for matrix multiplication

MaxGaukler:
The meaning of ``*`` is ambiguous: Some libraries such as NumPy use it for elementwise multiplication, others for matrix multiplication. Since Python3 and [PEP465](https://www.python.org/dev/peps/pep-0465/), there is a new operator ``@`` for matrix multiplication. 

This pull request adds ``@`` as an alias for ``*`` and includes some workarounds so that the tests using ``@`` can be skipped under ~~Python 2~~ Python <= 3.4, where it is a syntax error. (Not sure how long you plan to support Python 2, which has reached its end-of-life, see https://python3statement.org/ ).

### DIFF
--- a/mpmath/matrices/matrices.py
+++ b/mpmath/matrices/matrices.py
@@ -205,15 +205,25 @@ class _matrix(object):
          ['2.0', '3.0']])
 
     Of course you can perform matrix multiplication, if the dimensions are
-    compatible::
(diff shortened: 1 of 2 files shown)


fredrik-johansson:
Seems fine. I have nothing against removing Python 2 support at this point.

PR #526, opened 2020-03-20
Implement unary plus for matrices (+M == M)

MaxGaukler:


fredrik-johansson:
For scalars, the meaning is actually slightly different: +x rounds to the current precision. This should probably be consistent for matrices.

MaxGaukler:
fixed that.

MaxGaukler:
If there is anything missing to get this merged, please let me know.

fredrik-johansson:
No, this is indeed fine. Thanks!

PR #527, opened 2020-03-20
Support randmatrix() for mp.iv and mp contexts

MaxGaukler:
Before, `mpmath.iv.randmatrix()` caused an error. Now it works correctly:

```
>>> import mpmath
>>> mpmath.iv.randmatrix(3)
matrix(
[['[0.66382280562735496154, 0.66382280562735496154]', '[0.016861908353603016764, 0.016861908353603016764]', '[0.73734056129022396142, 0.73734056129022396142]'],
 ['[0.36647301100616180847, 0.36647301100616180847]', '[0.62220372249504873885, 0.62220372249504873885]', '[0.48761902415903068952, 0.48761902415903068952]'],
 ['[0.81322606973113087392, 0.81322606973113087392]', '[0.15903271707414323721, 0.15903271707414323721]', '[0.090983058161586249923, 0.090983058161586249923]']])
>>> type(mpmath.fp.randmatrix(3)[0,0])
<type 'float'>
>>> type(mpmath.mp.randmatrix(3)[0,0])
>>> type(mpmath.iv.randmatrix(3)[0,0])
<class 'mpmath.ctx_iv.ivmpf'>
```

- [ ] Is it okay that I moved `import random` to the top level? AFAIK it should even be faster this way.

### DIFF
--- a/mpmath/ctx_base.py
+++ b/mpmath/ctx_base.py
@@ -1,4 +1,5 @@
 from operator import gt, lt
+import random
 
 from . import libmp
 from .calculus.calculus import CalculusMethods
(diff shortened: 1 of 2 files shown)


fredrik-johansson:
There is a small problem: random.random() returns a random double, so this will not give full precision random numbers.

mgmax:
Should be better now. I added tests, but am not sure how `workprec()` is supposed to work for `iv`. Is that a bug or by design?

MaxGaukler:
By the way, please let me know as soon as you consider it ready for merging so that I can provide a rebased version of this pull request as a single commit.

skirpichev:
@MaxGaukler could you please fix merge conflicts?

MaxGaukler:
Now squashed and rebased. Happy to see that there is interest in my suggestion :-)

mpmath/ctx_fp.py
```diff
@@ -230,10 +231,6 @@ def harmonic(ctx, x):
     def to_fixed(ctx, x, prec):
         return int(math.ldexp(x, prec))
 
-    def rand(ctx):
```
skirpichev, line 233: I don't know why this was replaced with the code below: new `ctx.convert` call is just a source of an additional slowdown.

I think this should be reverted.

skirpichev (changes requested):
There are also merge conflicts.  Please fix them.

Except for comment in the ctx_fp.py and above remark - seems ok for me.

mpmath/ctx_fp.py
MaxGaukler: You're right. I reverted this.

skirpichev (approved):

PR #528, opened 2020-03-20
Matrix equality: Implement M == None, raise exception for unknown types

MaxGaukler:
The following statements now no longer give a cryptic error message:
```
>>> import mpmath, numpy
>>> mpmath.eye(2) == None
False
>>> mpmath.eye(2) == numpy.eye(2)   
...
NotImplementedError: Unsupported type for testing equality: mpmath.matrix == <type 'numpy.ndarray'>
```

### DIFF
--- a/mpmath/matrices/matrices.py
+++ b/mpmath/matrices/matrices.py
@@ -690,8 +690,19 @@ class _matrix(object):
         return -self + other
 
     def __eq__(self, other):
-        return self.__rows == other.__rows and self.__cols == other.__cols \
-               and self.__data == other.__data
+        if other is None:
+            return False
+        try:
+            return self.__rows == other.__rows and self.__cols == other.__cols \
+                   and self.__data == other.__data
+        except AttributeError:
+            return NotImplemented
+
+    def __ne__(self, other):
+        v = self.__eq__(other)
+        if v is NotImplemented:
+            return v
+        return not v
 
     def __len__(self):
         if self.rows == 1:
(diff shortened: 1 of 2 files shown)


fredrik-johansson:
Isn't the recommended way to handle unsupported comparisons to return NotImplemented?

MaxGaukler:
My intention was to keep behavior backwards-compatible and fail loudly instead of silently, especially for the case of numpy matrices. If the working precision is not too large, mpmath and numpy matrices can be semantically equal, so I think it should not be that `mpmath.matrix(list) == np.array(list)` is `False`. Unfortunately, as far as I understand, Python will return `False` for the comparison of two objects that each return `NotImplemented` in `__eq__`.

MaxGaukler:
I now pushed a new version that uses `NotImplemented`. The semantics are now more consistent with the Python standard, and - surprisingly - comparison with `numpy` works. However, some results are counterintuitive:
```
>>> import numpy
>>> import mpmath
>>> A=numpy.array([42])
>>> B=mpmath.matrix([42])
>>> A == B
array([ True], dtype=bool)
>>> A == 42
array([ True], dtype=bool)
>>> B == 42
False
>>> any(B==42)
TypeError: 'bool' object is not iterable
>>> A == [42]
array([ True], dtype=bool)
>>> B == [42]
False
>>> any(A==42)
True
```
See that `[42] == A` and `A == B` but not `[42] == B`.

skirpichev:
Please fix merge conflicts an reopen, if you wish

MaxGaukler:
rebased, please reopen (I don't see a button for reopening) :-)

skirpichev:
Hmm, "Reopen and comment" button looks as gray for me...  Feel free to open a new pr, if this seems relevant to you.

MaxGaukler:
continued in https://github.com/mpmath/mpmath/pull/686

PR #529, opened 2020-03-22
fixed indexing errors in inverse laplace transform algorithm, as pointed out by Mark Bakker

klkuhlm:
The fix made a tiny change to the near-zero imaginary results in a couple doctests, so I updated those too.

There appear to be some other failing doctests (e.g., missing "xrange"), but they have nothing to do with inverselaplace.py, so I left them.


fredrik-johansson:
Thanks!

PR #533, opened 2020-03-24
Fix scalar-matrix-multiplication for mpi * iv.matrix

MaxGaukler:
Fixes this error:
```
>>> mpmath.mpi(1,2) * mpmath.iv.eye(1)
...
TypeError: cannot unpack non-iterable NoneType object
```

And adds a test for the currently failing case of `mpmath.mp.matrix * mpmath.iv.matrix`. That test is marked as expected failure.


MaxGaukler:
Should be ready to merge - please let me know if you suggest any change.

fredrik-johansson:
Sure, it makes sense to allow this.

Issue #542, opened 2020-06-21
Determinant function raises TypeError on simple matrix


sonntagsgesicht:
determinant function `det(A)` on matrix `A = matrix([[0, 0, 4], [2, 0, 3], [0, 0, -3]])` raises 
`TypeError: '>=' not supported between instances of 'NoneType' and 'int'`.

Due to issue in `LU_decomp` line `143` at `ctx.swap_row(A, j, p[j])` according `p[j]`being none, since `A` has a zero column.

Many be `det` should check first if there is any zero column. If so `det` is zero and `LU_decomp` is not required any more.

```
   if any(all(almosteq(0,x) for x in A[:,c]) for c in range(A.cols)):
       return 0*A[0,0]
```

skirpichev:
Fixed by 79d99ff

PR #543, opened 2020-06-21
fixing issue: Determinant function raises TypeError on simple matrix #542

sonntagsgesicht:


fredrik-johansson:
The solution looks like a bit of a hack, and it's problematic that it doesn't account for the scaling of the matrix. For example, it will return 0 if you input the identity matrix times 1e-30. It would be fine if you check any column for being exactly zero, but perhaps that isn't general enough.

sonntagsgesicht:
Yes, you are right. Had a closer look and fixed it directly inside LU_comp where pivoting failed. 
Added a LU_comp, too. But, since I have no `pytest` experience on testing exceptions, this test looks more like a hack, too.

mgmax:
If you don't like the compact `pytest.raises(Exception, function)` notation, you can also use `pytest.raises` like this:
```
    with pytest.raises(MyException):
        do_something_wrong()
```
This checks that the expected exception is thrown somewhere within the "with"-block, and is actually the recommended notation in the [pytest docs](https://docs.pytest.org/en/latest/reference.html#pytest-raises).

vks:
The updated fix looks good to me.

vks (approved):

PR #552, opened 2020-09-16
Fix Travis CI build error

MaxGaukler:
Travis CI for the current master branch gives errors for some Python versions:
```
pluggy.manager.PluginValidationError: Plugin 'pytest_cov' could not be loaded: (pytest 4.1.0 (/home/travis/virtualenv/python3.5.6/lib/python3.5/site-packages), Requirement.parse('pytest>=4.6'))!
```
https://travis-ci.org/github/MaxGaukler/mpmath/builds/727727144
The error may be due to https://github.com/z4r/python-coveralls/issues/66

Solution: drop CI support for Python 3.4 and 3.5, and apply a workaround for pypy. Note that Python 3.4 and 3.5 are EOL, so other projects like SymPy don't support them either (see [SymPy python version support policy](https://github.com/sympy/sympy/wiki/Python-version-support-policy))

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,6 @@
 language: python
 matrix:
+# When changing the supported python versions, please keep the README, .travis-ci.yml and setup.cfg in sync!
   include:
     - arch: arm64
       python: 2.7
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,5 @@
 language: python
 matrix:
-# When changing the supported python versions, please keep the README, .travis-ci.yml and setup.cfg in sync!
   include:
     - arch: arm64
       python: 2.7
@@ -16,14 +15,8 @@ matrix:
       python: 3.8
     - arch: amd64
       python: pypy
-      before_install:
-        # Workaround for dependency errors, probably due to https://github.com/z4r/python-coveralls/issues/66
-      - pip install --upgrade --force-reinstall pytest-cov==2.5.0 
     - arch: amd64
       python: pypy3
-      before_install:
-        # Workaround for dependency errors, probably due to https://github.com/z4r/python-coveralls/issues/66
-        - pip install --upgrade --force-reinstall pytest-cov==2.5.0 
     - arch: arm64
       python: 2.7
       env: COVERAGE='on'
@@ -37,7 +30,8 @@ matrix:
       python: 3.6
       env: COVERAGE='on'
 install:
-  - pip install -U -e .[develop]
+  - pip install 'pip>=20.2.2'
+  - pip install --use-feature=2020-resolver -U -e .[develop]
 script:
   # Guard against invalid escapes in strings, like '\s'.
   - python -We:invalid -m compileall -f mpmath -q
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,10 +6,6 @@ matrix:
     - arch: amd64
       python: 2.7
     - arch: arm64
-      python: 3.5
-    - arch: amd64
-      python: 3.5
-    - arch: arm64
       python: 3.7
     - arch: amd64
       python: 3.7
@@ -19,8 +15,14 @@ matrix:
       python: 3.8
     - arch: amd64
       python: pypy
+      before_install:
+        # Workaround for dependency errors, probably due to https://github.com/z4r/python-coveralls/issues/66
+      - pip install --upgrade --force-reinstall pytest-cov==2.5.0 
     - arch: amd64
       python: pypy3
+      before_install:
+        # Workaround for dependency errors, probably due to https://github.com/z4r/python-coveralls/issues/66
+        - pip install --upgrade --force-reinstall pytest-cov==2.5.0 
     - arch: arm64
       python: 2.7
       env: COVERAGE='on'
@@ -33,14 +35,6 @@ matrix:
     - arch: amd64
       python: 3.6
       env: COVERAGE='on'
-    - arch: arm64
-      python: 3.4
-      before_install:
-        - pip install --upgrade --force-reinstall setuptools
-    - arch: amd64
-      python: 3.4
-      before_install:
-        - pip install --upgrade --force-reinstall setuptools
 install:
   - pip install -U -e .[develop]
 script:
--- a/setup.cfg
+++ b/setup.cfg
@@ -6,6 +6,7 @@ author = Fredrik Johansson
 author_email = fredrik.johansson@gmail.com
 url = http://mpmath.org/
 license = BSD
+# When changing the supported python versions, please keep the README, .travis-ci.yml and setup.cfg in sync!
 classifiers = License :: OSI Approved :: BSD License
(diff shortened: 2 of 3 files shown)


skirpichev:
>  drop CI support for Python 3.4 and 3.5

@MaxGaukler, probably you should reflect this in docs.  E.g. in the README (which was already wrong about testing on CPython 3.4)

BTW, @fredrik-johansson I didn't see the Travis-CI check in the list of successful checks.  Is it intentional?

MaxGaukler:
Thanks for the hint! Done.

.travis.yml
```diff
@@ -19,8 +16,14 @@ matrix:
       python: 3.8
     - arch: amd64
       python: pypy
+      before_install:
+        # Workaround for dependency errors, probably due to https://github.com/z4r/python-coveralls/issues/66
```
skirpichev, line 20: Hmm, why don't require a specific version in the setup.cfg or, perhaps better just blacklist wrong versions?

setup.cfg
```diff
@@ -6,15 +6,14 @@ author = Fredrik Johansson
 author_email = fredrik.johansson@gmail.com
 url = http://mpmath.org/
 license = BSD
+# When changing the supported python versions, please keep the README, .travis-ci.yml and setup.cfg in sync!
```
skirpichev, line 9: Maybe it does make sense to add a special section/page (e.g. Development, where can be documented such things as well as release process and so on) for docs, which will have such details?  I don't think it does make sense to do this in the current PR, but you could open a bug and delete all these scattered comments.

skirpichev (changes requested):

.travis.yml
MaxGaukler: I have no clue why the problem only exists for pypy but not for CPython. What do you suggest?

setup.cfg
MaxGaukler: Yes, as soon as such a document exists then it may be outsourced. Even then, a comment reminding to please read the doc before changing something could be helpful, as it is non-obvious for outsiders how to correctly change something.

In my experience, the CI config is something that is touched by "outsiders" more often than expected; often something suddenly breaks due to external dependencies and then someone randomly notices it and proposes some random fix without considering all the details. Just like me :-D

If whoever decides on merging this PR wants to remove the comment I'm happy to remove it.

.travis.yml
skirpichev: I don't know.  Perhaps, because PyPy != CPython in general, there are some incompatibilities (=PyPy bugs).

> What do you suggest?

You can blacklist "wrong" pytest-cov versions in the setup.cfg, using operators (!= or <).   See e.g. [this](https://setuptools.readthedocs.io/en/latest/userguide/dependency_management.html#declaring-required-dependency).
I think, you can even restrict this version restriction _only_ for PyPy (AFAIR this is something like: "pytest-cov != x.y.z; implementation_name='pypy'", you could take look on PEP 508 and setuptools docs).

setup.cfg
skirpichev: > comment reminding to please read the doc before changing something could be helpful

I strongly disagree.

> as it is non-obvious for outsiders how to correctly change something

Most non-obvious things/workflows - probably should be documented in the project docs.  Contributors must
read one.  It may worth to point on this in the [github-specific CONTRIBUTING file](https://docs.github.com/en/free-pro-team@latest/github/building-a-strong-community/setting-guidelines-for-repository-contributors).

> If whoever decides on merging this PR wants to remove the comment I'm happy to remove it.

It's difficult to keep similar comments in sync.  Better - have some project-wide guidelines in a single place.

.travis.yml
MaxGaukler: I now found the actual cause, which is a bug in pip that occurs if you an old version of an indirect dependency is already installed. This means the workaround is no longer needed.

setup.cfg
MaxGaukler: I removed these for now, I care most about getting it merged in so I can have my other pull requests merged in so I can then switch back my code to an upstream mpmath version.

.travis.yml
skirpichev: Hmm, maybe you could just set minimal pytest requirements in the setup.cfg.  This looks to me as a simplest solution.

"pytest >= 4.6" seems to be working, see [build logs](https://travis-ci.org/github/skirpichev/mpmath/builds/732148303).  Everything else then should be reverted (this works for all versions in the master) and, perhaps, you will want also rebase the PR to squash all commits to single one.

MaxGaukler:
I claim that your suggestion does not solve the root cause, but is merely a workaround that will fail as soon as the following conditions are met:

- pip < 20.3 (which is probably not yet released) 
- already installed pytest >= 4.6 (or whatever version number we depend on), but not the newest version 
- pytest-cov not yet installed
- newest version  of pytest-cov requires a newer pytest than what is installed. 

Then, the old pip dependency resolver will not detect that pytest must also be upgraded, but just warn that this causes an incompatible situation. 

The same issue can also reappear for other package combinations. 

I don't have the time for further experiments, but I can't imagine any other reasonable explanation why enabling the new resolver solves the problem.

skirpichev:
> I claim that your suggestion does not solve the root cause

Well, it fix CI tests.  And this approach seems to be more transparent for local setups.

Your pip workaround will not work in local setups, if old pytest version is present.  Either you should put additional
instructions to upgrade&run pip or specify the upcoming pip version as a requirement.  And I think even after the new pip
version will be released (with new resolver as default) - it's a bad idea to enforce such version in the setup.cfg.  We may set
up development requirements (e.g. pytest) to very up-to-date versions, but the pip package - is not such one, this may affect usual installations.

MaxGaukler:
The bug is in pip and can not be properly fixed by changing dependencies, except if you would 'pip freeze' every single dependency and its indirect dependencies. 

As explained, your solution will break again every time when pytest-cov or any other package increases its dependencies beyond what is installed locally. My suggestion would solve the problem forever.

skirpichev:
> The bug is in pip and can not be properly fixed by changing dependencies

That's fair.  Addition of minimal pytest requirements is just a workaround.

> your solution will break again every time when pytest-cov or any other package increases its dependencies

Minimal deps, yes.  I don't think this will happen too often.

> My suggestion would solve the problem forever.

I agree.  But I think, your solution is not a good idea right now and #554 may be merged independently, as a quick fix.

It's not enough to fix the problem just for Travis.  I believe, you should specify a minimal pip requirement in the "develop" extras.  Also, either we should add instructions somewhere in the docs on how to properly call pip, or we can wait for the new pip version, where the 2020-resolver will be a default (see [this](https://github.com/pypa/pip/milestone/38)).

MaxGaukler:
Closing because the workaround #554 was merged instead. Unlike this PR, it doesn't address the root cause, but at least temporarily suppresses the problem for a few years until pytest-cov or some other module again increases its dependencies.

Issue #553, opened 2020-10-02
Move testing to the Github Actions?


skirpichev:
For example, this wil allow to run all jobs in one shot (Travis has limit of 4 or 5 concurrent jobs vs 20 on the GA).

In other aspects they are more or less feature-compatible, but arm64 build will be harder to setup.  I believe, that #501 - should be reverted.  For a platform-independent library like mpmath is - such tests are redundant.

PR #501, opened 2019-12-30
Adding Travis-Ci Support For Arm64

ossdev07:
Travis-CI has added support for ARM64. Added ARM64 job in Travis-CI.

PR #554, opened 2020-10-02
Fix Travis-CI in the master & some cleanup of the Travis config

skirpichev:
* this reverts #501. I do think - these tests don't make sense for python-only library (see also #553)
* run tests on CPython 3.9

@MaxGaukler, you may want to look on this too, esp. the first patch.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,7 @@ python:
   - 3.5
   - 3.7
   - 3.8
-  - 3.9-dev
+  - 3.9
   - pypy
   - pypy3
 jobs:
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,7 @@ python:
   - 3.7
   - 3.8
   - 3.9
+  - 3.10-dev
   - pypy
   - pypy3
 jobs:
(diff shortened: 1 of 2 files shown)


skirpichev:
@fredrik-johansson, this is an alternative to #552

MaxGaukler:
As the gmpy backend outsources the integer computations to an external library, testing it on arm64 could catch architecture-dependent bugs in that library. You may want to look into whether we test both backends (gmpy and pure python).

skirpichev:
> As the gmpy backend

I think, this belongs to the gmpy testing (and it seems - there are arm64 tests).

> You may want to look into whether we test both backends (gmpy and pure python).

Unfortunately, no.  The current config tests only pure-python arithmetics.

.travis.yml
```diff
@@ -1,58 +1,28 @@
...
+python:
+  - 2.7
+  - 3.4
+  - 3.5
+  - 3.7
+  - 3.8
+  - 3.9-dev
+  - pypy
+  - pypy3
 matrix:
```
cclauss, line 11:
```suggestion
jobs:
```
Matrix is deprecated in favor of Jobs.  See Travis CI build config validation

PR #560, opened 2020-10-26
revert previous change that had unintended consequences.

klkuhlm:
I was too hasty with that previous change. this reverts it.  sorry!


fredrik-johansson:
Ah, my bad. It looked harmless :-)

PR #564, opened 2020-12-17
PrecisionManager preserves caller signature

DamSenViet:
This commit properly preserves the type annotations of the PrecisionManager caller.

Here's an example snippet:

```python
@workdps(3)
def test(arg1: int, arg2: str) -> None:
    """Preserve this docstring.

    :param arg1: Test
    :type arg1: int
    :param arg2: arg2
    :type arg2: str
    """
    return None
```

By properly wrapping the calling function, this  documentation could then be produced from probing the exports.

**BEFORE THE FIX**

image

**AFTER THE FIX**

image

The type annotations were properly kept after the fix.

### DIFF
--- a/mpmath/ctx_mp.py
+++ b/mpmath/ctx_mp.py
@@ -1322,8 +1322,6 @@ class PrecisionManager:
                     return f(*args, **kwargs)
             finally:
                 self.ctx.prec = orig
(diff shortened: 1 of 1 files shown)


fredrik-johansson:
Makes sense. Thanks.

PR #566, opened 2021-01-22
remove force_type arguments, they were broken

MaxGaukler:
Matrices should be type-stable, not allowing mixed types in one matrix. Always use the type of the context.
Therefore, specifying an element type by the "force_type" argument does not make sense.

It did not work properly before either, the data was not stored in the given type:
```
>>> M = mpmath.matrix(mpmath.matrix([[0,1,2]]), force_type=bool)
>>> M
matrix(
[['0.0', '1.0', '1.0']])
>>> type(M[0,0])
>>> M._matrix__data
{(0, 1): mpf('1.0'), (0, 2): mpf('1.0')}
```

Now the argument is removed and gives a warning instead:
```
>>> mpmath.matrix(mpmath.matrix([[0,1,2]]), force_type=bool)
/mpmath/mpmath/matrices/matrices.py:288: UserWarning: The force_type argument was removed, it did not work properly anyway. If you want to force floating-point or interval computations, use the respective methods from `fp` or `mp` instead, e.g., `fp.matrix()` or `iv.matrix()`. If you want to truncate values to integer, use .apply(int) instead.
  warnings.warn("The force_type argument was removed, it did not work"
matrix([['0.0', '1.0', '2.0']])
```

### DIFF
--- a/mpmath/calculus/optimization.py
+++ b/mpmath/calculus/optimization.py
@@ -598,7 +598,7 @@ def jacobian(ctx, f, x):
             J[i,j] = Jj[i]
     return J
 
-# TODO: test with user-specified jacobian matrix, support force_type
+# TODO: test with user-specified jacobian matrix
 class MDNewton:
     """
     Find the root of a vector function numerically using Newton's method.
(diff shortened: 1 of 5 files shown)


fredrik-johansson:
This PR contains some seemingly unrelated changes.

MaxGaukler:
All changes are related to the removal of force_type:
- Remove `force_type` for all functions that offer it
- add precise documentation of `iv.findroot()`, which is similar to the old `findroot(force_type=iv.mpf)` but may be misunderstood as "find an interval for the root".  (Maybe this was what you asked about?)
- adapt tests and documentation to new syntax (e.g., `fp.matrix(A)` instead of `matrix(A, force_type=float)`)
- Remove redundant uses of `convert()` that are no longer needed, now that `force_type` was removed

If this doesn't answer your question, could you please state more precisely what your concern is about?

fredrik-johansson:
Thanks, I understand with your clarifications.

PR #568, opened 2021-02-02
Automate PyPI uploads

skirpichev:


sandrotosi:
why not a separate action, just for PyPI publishing?

skirpichev:
I don't think it's useful here.  Unfortunately, there is no good way (well, when last time I checked it) to share common steps across several actions.  And the action is small so far.

sandrotosi:
why it's not useful? having an extra step in an action call "test" that's not related to testing, that's unexpected and likely source of confusion

skirpichev:
> why it's not useful?

Code duplication, as it was explained.  I don't think this will be a source of confusion: there is only one workflow, which fits on the screen.  Splitting it out right now seems like overkill for me.

fredrik-johansson:
I will try this.

fredrik-johansson:
How do I get the API token in the right place?

skirpichev:
@fredrik-johansson, as I said in the last commit message - you should add [API token](https://pypi.org/help/#apitoken) on PyPI and then add it's value as a secret in the mpmath repo.  Note, that the secret variable should be named as PYPI_API_TOKEN.

Added step will run next time, when you tag commit on the master branch.

fredrik-johansson:
Thanks, that should do it for the token.

You mean I have to make a new version tag (presumably 1.2.1) to get PyPI's attention, or can I force it to happen for the existing 1.2.0 tag in some other way?

oscarbenjamin:
You have to make a new tag. It isn't possible to unpush the tag and the action is triggered by pushing a tag.

skirpichev:
> You mean I have to make a new version tag (presumably 1.2.1) to get PyPI's attention

Yes.  No magic: existing releases must be uploaded by hand, sorry.

I would suggest using some alpha/beta versions to test things (I did this for the [Diofant](https://pypi.org/project/Diofant/#history)).  I think, it's ok to delete (some) alpha/beta releases from the PyPI afterwards.

fredrik-johansson:
Unfortunately, the builds are failing.

skirpichev:
Hmm, v1.2.1 seems to be on PyPI (both src and wheels), working fine.  Not sure that that means, perhaps a bug in the action (diofant uses v1.4.1 version).   I'll debug this.

skirpichev:
@fredrik-johansson, I can't reproduce this issue.  [Same code](https://github.com/diofant/diofant/commit/0d6b08b0c31b2c270ae6aeb4f6a5a1174b9c2fa9) is working fine for the Diofant.  I did [0.13.0a1](https://pypi.org/project/Diofant/0.13.0a1/) upload to test this.  Maybe you did some upload by hand for v1.2.1, right before?

fredrik-johansson:
No, I didn't try a manual upload.

Not a big deal that it errors as long as the uploads work though.

skirpichev:
> Not a big deal that it errors as long as the uploads work though.

An issue for me, exit  status should be ok. Unfortunately, I see no similar issues in the action repo and I can't reproduce this problem.

I'll try the last release and then create an issue.

sandrotosi:
it's pretty obvious reading the logs: the action already uploaded mpmath to PyPI when running 3.8, and when 3.9 runs and try to upload it again, PyPI says nonono (cause it already has a tarball for the same version).

once again, i want to highlight how pypi upload should be a separate action than CI, it would have avoided this situation

skirpichev:
@sandrotosi, I prepared a hotfix in #572, thank you.  I forgot to copy version selector while preparing a code snippet.

Feel free to provide a better solution, with a separate action.  As I said before, I don't see how we can share few common steps in actions.  And such a code duplication may lead to problems too.

sandrotosi:
you just dont share them. it's a handful of lines. not gonna spend more time debating this.

PR #573, opened 2021-02-15
Add optional gmpy deps & run tests with gmpy

skirpichev:


skirpichev:
On top of #575.

I see no significant speedup, but there are some coverage improvements.

PR #575, opened 2021-03-02
Drop support for CPython 3.4

skirpichev:


skirpichev:
Maybe 3.5 support should be also removed (as SymPy people did).

oscarbenjamin:
We have also discussed following nep 29:
https://numpy.org/neps/nep-0029-deprecation_policy.html

PR #581, opened 2021-03-18
Get version from __init__

### LINKED ISSUE #580: setuptools_scm dropped Python 2 breaking install
Thanks for the great project! 

It unfortunately looks like [setuptools_scm dropped support for Python 2](https://github.com/pypa/setuptools_scm/issues/541), which appears to have broken the install for mpmath on Python 2. Here's a [broken CI run](https://github.com/mikedh/trimesh/pull/1182/checks?check_run_id=2135678957) demonstrating the issue.

mikedh:
Attempt to fix #580. Gets version for setup from `__init__.py` to prevent duplication of version, and removes dependency on setuptools_scm. 

This is how many/most projects handle version; I'm not sure how mixing most data from `setup.cfg` with the version in `setup.py` works. Although in testing with `pip install -e` locally it worked fine.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -26,7 +26,6 @@ project_urls = Source = https://github.com/fredrik-johansson/mpmath
 [options]
 packages = find:
 setup_requires = setuptools>=36.7.0
-                 setuptools_scm>=1.7.0
 tests_require = mpmath[tests]
 [options.extras_require]
 tests = pytest >= 4.6
--- a/setup.py
+++ b/setup.py
@@ -1,6 +1,17 @@
 #!/usr/bin/env python
 
+import os
 import setuptools
 
+# where version is stored relative to setup.py
+_version_path = os.path.join(
+    os.path.dirname(__file__),
+    'mpmath/__init__.py')
+# get version without importing anything
+# note that this is relying on __version__
+# being the *first* line of the __init__ file
+with open(_version_path, 'r') as f:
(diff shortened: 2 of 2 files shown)


skirpichev:
I think, much simpler solution is to pin required deps (that may be a conditional pin, only for py2), until we not drop support for py2 as others.  No need to reinvent wheel.

asmeurer:
+1 for removing setuptools_scm.

mikedh:
I guess I don't know the full context of *why* one would want `setuptools_scm` From my perspective, it is a narrowly used package with a very casual attitude towards breaking things upstream ("just upgrade your pip"). I don't use Python 2, but I respect that somewhere between 5-10% of people do for various reasons, and my plan was to keep libraries I maintain installable on Python 2 until the wheels really fall off :smile:.  

That being said, I dropped `sympy` from Python 2 installs for this issue, so I no longer have a duck in this race and am I fine with whatever! There was some indications in that thread that pinning `setuptools_scm` version did not help until pip was upgraded (which happens in current `mpmath` builds so it's hard to say).

skirpichev:
> From my perspective, it is a narrowly used package with a very casual attitude towards breaking things upstream ("just upgrade your pip")

Looking on the thread, this seems to be an issue with [_very_ old pip versions](https://github.com/pypa/setuptools_scm/issues/541#issuecomment-802021317).  The Debian Stretch has something like this.

PR #583, opened 2021-03-25
Pin setuptools_scm deps for python_version<"3.6"

skirpichev:
Closes #580

An alternative to #581.

PR #586, opened 2021-03-26
Don't use setuptools_scm: revert to manual version management

### LINKED ISSUE #580: setuptools_scm dropped Python 2 breaking install

skirpichev:
The setuptools_scm usage was broken by 867d8784 (likely,
this is a source of problem in #585)

Closes #580 (an alternative to #583 and #581)
Solves #585 as well.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -2,6 +2,7 @@
 name = mpmath
 description = Python library for arbitrary-precision floating-point arithmetic
 long_description = file: README.rst
+version = attr: mpmath.__version__
 author = Fredrik Johansson
 author_email = fredrik.johansson@gmail.com
 url = http://mpmath.org/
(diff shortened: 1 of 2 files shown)


mikedh:
This looks like a great solution! 👍

skirpichev:
I'm not so sure.  `setuptools_scm` is a good solution for automation of trivial tasks.  Now maintainer will have to keep in sync `__version__` and git tag again.  For no reason, actually (#443 has no real problems, but it's "solution" #446 introduce a new one: #585).  py2 issue may be solved for setuptools_scm-based version management (#583).

Anyway, probably #583, #581 and #585 could be closed now.

mikedh:
Agreed, I think this fixes this issue. FWIW I tag-on-release using a [Github action](https://github.com/mikedh/trimesh/blob/master/.github/workflows/release.yml#L156-L177 ) which works pretty well.

skirpichev:
Should work too, why not.  Not sure if it's possible to sign such a tag, however.

PR #590, opened 2021-04-22
Trivial fix in betainc() for x1==x2

skirpichev:


skirpichev:
This also workaround test failure on the master, due to broken pytest on the recent CPython 3.10.

WarrenWeckesser (approved):
Looks like a straightforward fix.

PR #596, opened 2021-05-01
Restore testing doctests/docs on CI

skirpichev:

PR #597, opened 2021-05-02
 Remove support for Python versions < 3.5

skirpichev:
On top of #596


WarrenWeckesser:
I think it is definitely time to drop Python 2 support.  Many libraries are already dropping support for 3.7!

skirpichev:
Other libraries have maintainers...   See e.g. [this](https://groups.google.com/g/mpmath/c/v8VjGgR1HH4) and [this](https://github.com/fredrik-johansson/mpmath/issues/550).

PR #600, opened 2021-05-18
Allow failures for 3.10-dev

skirpichev:


skirpichev:
Unfortunately, GA has no Travis-like allow-failures option: the whole status will be "red".

skirpichev:
Probably, not an option, until https://github.com/actions/toolkit/issues/399 is open.

PR #620, opened 2022-02-21
Faster numerical inverse Laplace transform

guillermo-navas-palencia:
Faster numerical inverse Laplace transform by using a linear convergence acceleration method. The method is described in 
http://gnpalencia.org/blog/2022/invertlaplace/, including performance comparison against the three methods currently implemented.


WarrenWeckesser:
Several lines in the file inverselaplace.py had trailing white space, which makes pycodestyle fail in the github CI tests.  I submitted a pull request that does a bit of PEP-8 clean up for this file: https://github.com/fredrik-johansson/mpmath/pull/642

PR #621, opened 2022-03-17
Allow returning information from findroot solvers

orbisvicis:
This information can be used to create basin diagrams for roots. I have some large exponentials and luckily mpmath is able to solve roughly 2/3 of the points for which SciPy overflows. I don't know if the iteration counts are comparable with those from SciPy's "hybr" solver.

The general approach is provided in `findroot` with an example in `MDNewton`. I assume more information can be returned from other solvers as the need arises.

### DIFF
--- a/mpmath/calculus/optimization.py
+++ b/mpmath/calculus/optimization.py
@@ -671,7 +671,7 @@ class MDNewton:
             while True:
                 if x1 == x0:
                     if self.verbose:
-                        print("canceled, won't get more excact")
+                        print("canceled, won't get more exact")
                     cancel = True
                     break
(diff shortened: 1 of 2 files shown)


skirpichev (changes requested):
In principle, I'm -0 for this feature: you could easily change your function/jacobian functions to update some evaluation counters...  Can you suggest a use case where this will not work?

On another hand, scipy.optimize.OptimizeResult has similar counters...

But @orbisvicis please document new option and new returned value (for the findroot function and solver's).

skirpichev:
No activity here.  I'm closing this, feel free to address review and reopen.

PR #629, opened 2022-05-22
DEPS: drop py2 from test and doc, remove compat #594

### LINKED ISSUE #594: Drop Python 2 support
SymPy hasn't supported Python 2 for over a year. I would support dropping it in mpmath, especially if active development is being done on it.

_Originally posted by @asmeurer in https://github.com/fredrik-johansson/mpmath/issues/580#issuecomment-829737792_

An easy to do issue, if we decide.  There are also dozens of workarounds for py2.x versions and for py3 versions < 3.5.  We should clear this as well.

fangchenli:
close #594


cclauss:
Let's not just drop tests without fixing the code.  `basestring`, `long`, `raw_input`, `xrange` were all removed in Python 3

% `flake8 . --count --select=E9,F63,F7,F82,Y --show-source --statistics`
```
./mpmath/ctx_mp_python.py:144:29: F821 undefined name 'long'
    def __long__(s): return long(to_int(s._mpf_))
                            ^
./mpmath/libmp/backend.py:33:11: F821 undefined name 'long'
    MPZ = long
          ^
./mpmath/libmp/backend.py:34:14: F821 undefined name 'xrange'
    xrange = xrange
             ^
./mpmath/libmp/backend.py:35:18: F821 undefined name 'basestring'
    basestring = basestring
                 ^
./demo/mandelbrot.py:27:14: F821 undefined name 'xrange'
    for i in xrange(ITERATIONS):
             ^
./demo/pidigits.py:73:14: F821 undefined name 'raw_input'
    tofile = raw_input("Output to file? (enter a filename, or just press " \
        "enter\nto print directly to the screen) \n> ")
             ^
./demo/pidigits.py:79:5: F821 undefined name 'raw_input'
    raw_input("\nPress enter to close this script.")
    ^
```

.github/workflows/test.yml
```diff
@@ -5,7 +5,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [2.7, 3.5, 3.6, 3.7, 3.8, 3.9, '3.10', 3.11-dev, pypy2.7, pypy3.9]
+        python-version: ["3.5", "3.6", "3.7", "3.8", "3.9", "3.10", "3.11-dev", "pypy-3.7", "pypy-3.8", "pypy3.9"]
```
cclauss, line 8: This is a lot of CPU power.  It might be free in the sense of no cost but...  It will chew up electricity and heat the planet so let's be reasonable and not test all possible versions but only odd or even numbered versions.

skirpichev:
@fangchenli, unfortunately, there are test failures (mostly/all? related to range function vs kwarg conflict), please fix this.
See also previous attempt #597.  I think, you left some obsolete stuff (e.g. in docs/basics.rst or docs/functions/trigonometric.rst).

.github/workflows/test.yml
skirpichev: Tests are fast, so no reason to skip them.  @fangchenli but redundant quotes could be removed.  IIUIC, this is required only for "pypy-*" stuff.

skirpichev (changes requested):
Looks fine, except for comments above.

.github/workflows/test.yml
```diff
@@ -27,13 +27,10 @@ jobs:
...
     - name: Tests
       run: pytest
-    - name: Remove gmpy on 2.7
-      if: matrix.python-version == 2.7
-      run: pip uninstall -y gmpy2
     - name: Run coverage tests & upload results
       env:
         PYTEST_ADDOPTS: --cov mpmath
-      if: matrix.python-version == 2.7 || matrix.python-version == 3.6
+      if: matrix.python-version == 3.5 || matrix.python-version == 3.10
```
skirpichev, line 33: I think it's fine to test here just one py3 version, unless there are code path (I don't think so) which are covered by different python versions.

pyproject.toml
```diff
@@ -0,0 +1,9 @@
+[tool.pytest.ini_options]
+minversion = "6.0"
+testpaths = [
+    "mpmath",
+    "docs",
```
cclauss, line 5:
```suggestion
    "docs",
    "mpmath",
```

fredrik-johansson:
Hold on; this patch contains does a bunch of stuff beyond compatibility fixes, changing code like

    if ttype is x:

into

    if isinstance(t, x):

in performance-critical places (arithmetic operations). The ``isinstance`` calls are avoided on purpose here as they are much more expensive than ``is`` checks, or at least they used to be much more expensive in older versions of Python. Maybe newer Pythons made ``isinstance`` really cheap, but the code should not be changed without profiling to verify this, and such a change should go in a separate PR.

cclauss:
`isinstance()` is faster on modern Python.

% `python3.11 -m timeit "type(1) == int"`
```
20000000 loops, best of 5: 17.4 nsec per loop
```
% `python3.11 -m timeit "isinstance(1, int)"`
```
20000000 loops, best of 5: 16.6 nsec per loop
```

fredrik-johansson:
OK, but in the original code ``type()`` is called only once while the new code may call ``isinstance()`` several times, and the operator is ``is``, not ``==``.

Anyway, what you'd need to profile is actual mpmath code, not the atomic Python operations.

cclauss:
5000000 loops, best of 5: 91.5 nsec per loop  # Python 3.7
5000000 loops, best of 5: 63.9 nsec per loop
5000000 loops, best of 5: 92.8 nsec per loop  # Python 3.8
5000000 loops, best of 5: 49.6 nsec per loop
5000000 loops, best of 5: 79.8 nsec per loop  # Python 3.9
5000000 loops, best of 5: 54.5 nsec per loop
5000000 loops, best of 5: 55.5 nsec per loop  # Python 3.10
5000000 loops, best of 5: 55 nsec per loop
10000000 loops, best of 5: 30.9 nsec per loop  # Python 3.11
10000000 loops, best of 5: 26.9 nsec per loop
---
GitHub Action: `.github/workflows/type_benchmark.yml`
```
name: type_benchmark
on: [push]
jobs:
  type_benchmark:
    runs-on: ubuntu-latest
    steps:
      - uses: actions/setup-python@v4
        with:
          python-version: |
            3.7
            3.8
            3.9
            3.10
            3.11
      - env:
          OLD: "type(1) == int"
          NEW: "isinstance(1, int)"
        run: |
          python3.7  -m timeit "$OLD"
          python3.7  -m timeit "$NEW"
          python3.8  -m timeit "$OLD"
          python3.8  -m timeit "$NEW"
          python3.9  -m timeit "$OLD"
          python3.9  -m timeit "$NEW"
          python3.10 -m timeit "$OLD"
          python3.10 -m timeit "$NEW"
          python3.11 -m timeit "$OLD"
          python3.11 -m timeit "$NEW"
```

fredrik-johansson:
Timing some mixed mpf arithmetic operations, there doesn't seem to be any measurable impact on performance anymore.

So I'll withdraw my objection, considering that the new version of the code is simpler / more idiomatic.

skirpichev:
@fredrik-johansson, from simple benchmarks it looks like isinstance() calls slightly slower:
```
$ python -m timeit -r11 'isinstance(1, int)'
5000000 loops, best of 11: 90.5 nsec per loop
$ python -m timeit -r11 'type(1) == int'
5000000 loops, best of 11: 88.9 nsec per loop
$ python -m timeit -r11 'type(1) is int'
5000000 loops, best of 11: 74.1 nsec per loop
$ python -m timeit -r11 -s 'it=(int,)' 'type(1) in it'
5000000 loops, best of 11: 84.5 nsec per loop
$ python --version
Python 3.11.1+
```
The other point is that this change is totally unrelated to the subject of this pr (which is big enough).  There is nothing wrong with the old code from the py3 point of view, so I would suggest to revert that part.

docs/basics.rst
```diff
@@ -125,9 +125,8 @@ It is often useful to change the precision during only part of a calculation. A
     >>> # do_something()
     >>> mp.prec -= 2
 
-Since Python 2.5, the ``with`` statement along with the mpmath functions ``workprec``, ``workdps``, ``extraprec`` and ``extradps`` can be used to temporarily change precision in a more safe manner:
+The ``with`` statement along with the mpmath functions ``workprec``, ``workdps``, ``extraprec`` and ``extradps`` can be used to temporarily change precision in a more safe manner:
```
skirpichev, line 128: You left below yet another remark about the with statement ("The effect of the ``with`` statement can be emulated in Python 2.4...")

pyproject.toml
skirpichev, line 1: I don't think we should add pyproject.toml right here.  I've cherry-picked addition of `testpaths` option in #664

skirpichev (changes requested):
This pr still lacks some changes from #597.  So far:
- docs/functions/trigonometric.rst
- mpmath/ctx_fp.py (real/imag properties)
- mpmath/ctx_iv.py (numbers abc)
- mpmath/functions/functions.py (real/imag properties)
- mpmath/functions/zeta.py (py2.4 try block)
- mpmath/libmp/libmpf.py (from_float)
- mpmath/math2.py (WA for log/sqrt and real/imag)
- mpmath/rational.py (numbers abc)
- mpmath/tests/torture.py (multiprocessing)

Lets also revert isinstance() changes.  This doesn't belongs to the subject of the pr.

docs/basics.rst
```diff
@@ -125,9 +125,8 @@ It is often useful to change the precision during only part of a calculation. A
...
 
-    >>> from __future__ import with_statement  # only need this in Python 2.5
     >>> with workdps(20):  # doctest: +SKIP
```
skirpichev, line 130: I think you can also UnSKIP this test.

setup.cfg
```diff
@@ -29,7 +30,9 @@ packages = find:
 setup_requires = setuptools>=36.7.0
 tests_require = mpmath[tests]
 [options.extras_require]
-tests = pytest >= 4.6
+tests = 
+    pytest >= 6.0
+    pytest == 6.*; python_version < '3.7'
```
skirpichev, line 35:
```suggestion
tests = pytest >= 6
```
?

mpmath/rational.py
```diff
@@ -81,22 +75,20 @@ def __eq__(s, t):
         return NotImplemented
 
     def __ne__(s, t):
-        ttype = type(t)
-        if ttype is mpq:
+        if isinstance(t, mpq):
             return s._mpq_ != t._mpq_
-        if ttype in int_types:
+        if isinstance(t, int_types):
```
skirpichev, line 80: Not all type->isinstance changes were reverted.

skirpichev:
mpmath/libmp/libmpf.py, mpmath/ctx_mp.py, mpmath/matrices/matrices.py and mpmath/tests/test_matrices.py still have workarounds for py2 (or py3<3.5).  Try grep:)

skirpichev (approved):

cclauss (approved):
LGTM  Let's drop the legacy!!!

mpmath/functions/zeta.py
```diff
@@ -553,17 +550,13 @@ def zeta(ctx, s, a=1, derivative=0, method=None, **kwargs):
...
-                    pass
-            finally:
-                ctx.prec = prec
+            try:
+                if verbose:
+                    print("zeta: Attempting to use the Riemann-Siegel algorithm")
+                return ctx.rs_zeta(s, derivative, **kwargs)
+            except NotImplementedError:
+                if verbose:
+                    print("zeta: Could not use the Riemann-Siegel algorithm")
```
cbm755, line 559: maybe its explained in the commit messages, but why is the `finally: ctx.prec = prec` dropped here?

I don't immediately see it being changed, but I also haven't studied this code yet...

cbm755 (commented):
This looks great to me!  I made a few comments above, only one of any consequence (the dropping of a `finally: ctx.prec = prec`) block.

mpmath/functions/zeta.py
cbm755: I dug a little deeper: `rs_zeta` calls `rs_offline` which certainly changes the `.prec`.

mpmath/ctx_iv.py
```diff
@@ -538,14 +536,9 @@ def hypsum(ctx, p, q, types, coeffs, z, maxterms=6000, **kwargs):
...
-#     http://docs.python.org/2/library/numbers.html for list of abstract
-#     methods.
-try:
-    import numbers
-    numbers.Complex.register(ivmpc)
-    numbers.Real.register(ivmpf)
-except ImportError:
-    pass
+#   We do not subclass, hence we do not use the @abstractmethod checks. While
+#   his is less invasive it may turn out that we do not actually support
```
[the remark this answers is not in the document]
skirpichev:
```suggestion
#   this is less invasive it may turn out that we do not actually support
```

mpmath/functions/zeta.py
cbm755: It seems to be changed in 6437c031ab43d3d6c15a682198ab641fcbc0a643, but the commit log doesn't offer an explanation

skirpichev: Indeed, probably that was blindly copied from #597 and is my fault.  The workaround was introduced in b3bc191.

cbm755 (approved):

skirpichev (approved):

skirpichev:
@cbm755, thanks!  Lets merge this tomorrow.

skirpichev:
@fangchenli, thanks

PR #639, opened 2022-08-15
Change `bernoulli(1)` and `bernfrac(1)` to +½

### LINKED ISSUE #604: ellipe fails with pi as argument
When using `ellipe` with `pi`, there is an UnboundLocalError, while it works fine when using `+pi` or `1*pi`.

Here's an ipython session showing the problem:
```python
Python 3.9.2 (default, Feb 28 2021, 17:03:44) 
Type 'copyright', 'credits' or 'license' for more information
IPython 7.20.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: import mpmath

In [2]: mpmath.__version__
Out[2]: '1.2.1'

In [3]: mpmath.ellipe(mpmath.pi, 1)
UnboundLocalError                         Traceback (most recent call last)
<ipython-input-3-cbbd0507302e> in <module>
----> 1 mpmath.ellipe(mpmath.pi, 1)

/usr/lib/python3/dist-packages/mpmath/ctx_mp_python.py in f_wrapped(ctx, *args, **kwargs)
   1019                 try:
   1020                     ctx.prec += 10
-> 1021                     retval = f(ctx, *args, **kwargs)
   1022                 finally:
   1023                     ctx.prec = prec

/usr/lib/python3/dist-packages/mpmath/functions/elliptic.py in ellipe(ctx, *args)
   1274         RD = ctx.elliprd(x, y, 1)
   1275         return s*RF, -m*s**3*RD/3
-> 1276     return ctx.sum_accurately(terms) + P
   1277 
   1278 @defun_wrapped

/usr/lib/python3/dist-packages/mpmath/ctx_base.py in sum_accurately(ctx, terms, check_step)
    369                             break
    370                     k += 1
--> 371                 cancellation = max_mag - sum_mag
    372                 if cancellation != cancellation:
    373                     break

UnboundLocalError: local variable 'sum_mag' referenced before assignment

In [4]: mpmath.ellipe(+mpmath.pi, 1)
Out[4]: mpf('1.9999999999999998')

In [5]: mpmath.ellipe(1*mpmath.pi, 1)
Out[5]: mpf('1.9999999999999998')

In [5]: mpmath.ellipe(mpmath.e, 1)
Out[5]: mpf('1.5892187094970913')

```

Surprisingly it works with other `mpmath.ctx_mp_python.constant`s. I have not investigated what is going in detail.

Parcly-Taxel:
This pull request is part of the project at https://github.com/sympy/sympy/pull/23926 and also fixes #604.


mpmath/functions/zeta.py
```diff
@@ -288,20 +288,18 @@ def terms():
 
 @defun_wrapped
 def bernpoly(ctx, n, z):
-    # Slow implementation:
-    #return sum(ctx.binomial(n,k)*ctx.bernoulli(k)*z**(n-k) for k in xrange(0,n+1))
```
skirpichev, line 292: I think, this could be fixed, instead of silent removal.  `(-1)**k` factor, I guess.

mpmath/ctx_base.py
```diff
@@ -357,6 +357,7 @@ def sum_accurately(ctx, terms, check_step=1):
             while 1:
                 ctx.prec = prec + extraprec + 5
                 max_mag = ctx.ninf
+                sum_mag = ctx.zero
```
skirpichev, line 360: This is unrelated change, it lacks a test.  Lets fix https://github.com/mpmath/mpmath/issues/604 in https://github.com/mpmath/mpmath/pull/664.

skirpichev (changes requested):
I also would prefer a better commit message (describe reasons to change, etc).

By accepting this patch, we will be in sync with the SymPy.  But most of mathematical software still uses the current mpmath's convention (e.g. Mathematica's BernoulliB or scipy.special.bernoulli).  @fredrik-johansson, what do you think?

mpmath/ctx_base.py
Parcly-Taxel: OK, what do you want me to do with this PR **exactly**?

skirpichev: Lets just revert this:)

Parcly-Taxel: Done. Now what?

mpmath/functions/zeta.py
```diff
@@ -288,6 +288,8 @@ def terms():
 
 @defun_wrapped
 def bernpoly(ctx, n, z):
+    # Slow implementation:
+    #return sum(ctx.binomial(n,k)*ctx.bernoulli(k)*z**(n-k) for k in xrange(0,n+1))
```
skirpichev: That should be corrected, not just reverted.

Parcly-Taxel: I don't know how to do that. Can you suggest?

skirpichev: As I'd suggested above, probably you should add `(-1)**k` factor.

Parcly-Taxel: It can't be corrected by simply adding `(-1)**k` as an extra factor inside the summand. **The long code is required** to avoid accuracy issues with (say) `bernpoly(1000,10.5)`.

Parcly-Taxel: @skirpichev To illustrate what I mean, suppose we have
```python
@defun_wrapped
def bernpoly(ctx, n, z):
    n = int(n)
    return sum([ctx.binomial(n,k) * ctx.bernoulli(k) * (-1)**k * z**(n-k) for k in xrange(0,n+1)])
```
Then `bernpoly` is grossly wrong at large inputs:
```python
>>> from mpmath import *
>>> bernpoly(200, 17.5)
mpf('3.6365715774458614e+244')
>>> bernpoly(1000, 10.5)
mpf('-1.0815008290992428e+1778')
```

skirpichev: Indeed, I was misguided by your change in L327.

mpmath/functions/zeta.py
```diff
@@ -324,20 +324,20 @@ def terms():
             t = ctx.one
             k = 1
             while k <= n:
-                t = t*(n+1-k)/k * z
+                t = t*(k-1-n)/k*z
```
skirpichev, line 327: It seems, this sign change canceled by L332, isn't?  Then why is this?

mpmath/functions/zeta.py
skirpichev: @Parcly-Taxel, in fact, old version for the comment was incorrect as well for same reasons.  But without `(-1)**k` term now it's incorrect also for small arguments, e.g.:
```
>>> n=1;z=mpmath.mpf('1.1'); sum(mpmath.binomial(n,k)*(-1)**k*mpmath.bernoulli(k)*z**(n-k) for k in range(0,n+1))
mpf('0.60000000000000009')
>>> mpmath.bernpoly(n, z)
mpf('0.60000000000000009')
>>> n=1;z=mpmath.mpf('1.1'); sum(mpmath.binomial(n,k)**mpmath.bernoulli(k)*z**(n-k) for k in range(0,n+1))
mpf('2.1000000000000001')
>>> n=4;z=mpmath.mpf('1.3'); sum(mpmath.binomial(n,k)*(-1)**k*mpmath.bernoulli(k)*z**(n-k) for k in range(0,n+1))
mpf('0.11876666666666724')
>>> mpmath.bernpoly(n, z)
mpf('0.11876666666666673')
>>> n=4;z=mpmath.mpf('1.3'); sum(mpmath.binomial(n,k)*mpmath.bernoulli(k)*z**(n-k) for k in range(0,n+1))
mpf('8.9067666666666678')
```

Parcly-Taxel: Nah, that's just floating-point error. Nothing to change here.

skirpichev: The "floating-point error" is about 8.9 in the second example:)

Parcly-Taxel: If there is any error it's commented out. Do you want me to change the comment?

mpmath/function_docs.py
```diff
@@ -2291,6 +2291,10 @@
 
 For larger `n`, `B_n` is evaluated in terms of the Riemann zeta
 function.
+
+.. versionchanged :: 1.4
+    `bernoulli(1)` gives `+0.5` instead of `-0.5`. The previous
+    behavior can be obtained with `(-1)**n*bernoulli(n)`.
```
skirpichev, line 2297: I think the used convention should be also documented in the description of the function, before "examples" section.  And for the bernfrac() as well.  Also keep in mind that the default_role of sphinx is "math" for the mpmath.  So, for code you should use double backticks.

Parcly-Taxel: Done.

mpmath/function_docs.py
```diff
@@ -2237,6 +2237,12 @@
 returns a floating-point approximation. To obtain an exact
 fraction, use :func:`~mpmath.bernfrac` instead.
 
+.. versionchanged :: 1.4
+    ``bernoulli(1)`` gives ``+0.5`` instead of ``-0.5``.
+    This choice of value confers several theoretical advantages [1],
+    including an extension to complex parameters. The previous
+    behavior can be obtained with ``(-1)**n*bernoulli(n)``.
```
skirpichev, line 2244:
```suggestion
`B_1=0.5`, this choice of value confers several theoretical advantages [1].

.. versionchanged :: 1.4
    ``bernoulli(1)`` was changed from ``-0.5`` to ``+0.5``.  The previous
    behavior can be obtained with ``(-1)**n*bernoulli(n)``.
```
I meant something like this.  Take look on the CPython sphinx docs, for example, like https://docs.python.org/3/library/functions.html#float.  The versionchanged directive is just a log entry (and it could be removed).

Parcly-Taxel: Done.

skirpichev (approved):
Ok, this seems to be correct.  Of course, docstrings could be better, and there are still many unrelated changes.
Lets wait for Fredrik's opinion: if this does make sense at all (I'm +0).
UPD: https://groups.google.com/g/sage-devel/c/G4sbKoibXK4/m/LmO04RclCAAJ

mpmath/functions/zeta.py
skirpichev, line 1: I think you could revert your changes in this file.

skirpichev (changes requested):
After reading https://github.com/flintlib/flint2/pull/1179 and sage discussion, I think we should be more conservative.

Lets follow to the [Fredrik's suggestion](https://groups.google.com/g/sage-devel/c/G4sbKoibXK4/m/LmO04RclCAAJ) and introduce a keyword (`plus`?) for related functions (bernoulli and bernfrac).  The default value should correspond to the current convention.

Please adapt code and docstrings.

mpmath/functions/zeta.py
```diff
@@ -347,9 +347,9 @@ def eulerpoly(ctx, n, z):
         return z
     m = n+1
     if z == 0:
-        return -2*(ctx.ldexp(1,m)-1)*ctx.bernoulli(m)/m * z**0
+        return -2*(ctx.ldexp(1,m)-1)*ctx.bernoulli(m)/m
```
skirpichev, line 350:
```suggestion
        return -2*(ctx.ldexp(1,m)-1)*ctx.bernoulli(m)/m * z**0
```

mpmath/functions/zeta.py
```diff
@@ -347,9 +347,9 @@ def eulerpoly(ctx, n, z):
...
     if z == 1:
-        return 2*(ctx.ldexp(1,m)-1)*ctx.bernoulli(m)/m * z**0
+        return 2*(ctx.ldexp(1,m)-1)*ctx.bernoulli(m)/m
```
skirpichev, line 352:
```suggestion
        return 2*(ctx.ldexp(1,m)-1)*ctx.bernoulli(m)/m * z**0
```

skirpichev:
I fixed merge conflicts, yet I'm closing this stale pr.

Feel free to reopen it if you wish, but please adapt first code as it was suggested above (in short: control new behaviour with an option).

PR #576, opened 2021-03-04
Added the signal functions: square wave, triangle wave, sawtooth wave, and logistic sigmoid in signals.py and Sphinx documentation code in signals.txt

ndattani:

PR #640, opened 2022-08-18
Added the signal functions: square wave, triangle wave, sawtooth wave, and logistic sigmoid in signals.py and Sphinx documentation code in signals.txt

ndattani:
We squashed the 21 commits from Pull Request #576, into 3 commits: 1 by @ndattani, 1 by @ndattani and @t98yu, 1 by @deyanmihaylov.


ndattani:
Thanks @deyanmihaylov for fixing that problem I mentioned on Discord. Luckily that last few commits were all by you, so it should be easy to squash all your commits into 1.

Issue #638, opened 2022-08-09
`repr` of `mpmath.eps` always displays the working precision `2.22045e-16~`


WarrenWeckesser:
This is with mpmath 1.2.1:
```
In [1]: import mpmath

In [2]: mpmath.mp.dps = 40

In [3]: mpmath.eps   # By default, ipython displays the repr of the input
Out[3]: <epsilon of working precision: 2.22045e-16~>

In [4]: str(mpmath.eps)
Out[4]: '2.29588740394978028900143854926219858949e-41'
```
Why does `repr(mpmath.eps)` show `2.22045e-16`, while `str(mpmath.eps)` shows the expected value?

Presumably this is the result of the hard-coded argument `dps=15` in the `__repr__` method:
https://github.com/fredrik-johansson/mpmath/blob/fcc7e53362bc45fe372df4ac775e11678bd3f67d/mpmath/ctx_mp_python.py#L355-L356

Is that intentional?  I suspect line 356 should be changed to something like

```python
     return "<%s: %s~>" % (self.name, self.context.nstr(self, n=15)) 
```

PR #641, opened 2022-08-22
Fix repr(mp.eps)

WarrenWeckesser:
Before this change, repr(mp.eps) did not show the correct value
of eps.  The code was always converting the constant to dps=15, so
regardless of the setting of mp.dps, repr(mp.eps) would be

    '<epsilon of working precision: 2.22045e-16~>'

Closes gh-638.

### DIFF
--- a/mpmath/ctx_mp_python.py
+++ b/mpmath/ctx_mp_python.py
@@ -353,7 +353,7 @@ class _constant(_mpf):
         return self.func(prec, rounding)
 
     def __repr__(self):
-        return "<%s: %s~>" % (self.name, self.context.nstr(self(dps=15)))
+        return "<%s: %s~>" % (self.name, self.context.nstr(self, n=6))
 
 
 class _mpc(mpnumeric):
--- a/mpmath/tests/test_convert.py
+++ b/mpmath/tests/test_convert.py
@@ -35,6 +35,12 @@ def test_basic_string():
     assert str(mpf("-2163048125L/1088391168")) == '-1.98738118113799'
     assert str(mpf("2163048125/1088391168l")) == '1.98738118113799'
 
+
+def test_eps_repr():
+    mp.dps = 24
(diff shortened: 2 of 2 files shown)


fredrik-johansson:
Potentially an issue here is that if you set the precision very high, ``repr()`` will be slow for the constants that require a computation, ``khinchin`` for example. Not sure how to fix this easily without making a special case just for ``eps``.

WarrenWeckesser:
I see what you mean.  I just tried it with `dps=1500`, and the `repr()` for `khinchin` took about 10 seconds.  But the long computation apparently only happens with the first reference (presumably it is cached--I haven't dug into the code).  And a user working with `khinchin`  will have to pay that price at some point, whether it is in the `repr()` or an actual computation, so maybe it isn't that bad?

If the desire is to always show the same 6 digit approximation for the "true" constants (e.g. `e`, `pi`, `khinchin`) in the repr, perhaps the constants could have a string attribute added when the constant is created.  Then in `__repr__()`, only call `nstr(...)` if the attribute is `None`.  That way, no computation in `__repr__()`  is required for the true constants.  It would mean adding an optional argument in the appropriate constructor, e.g. `short_display=None` (naming is hard).  So `pi` would be created with `short_display='3.14159'`, `khinchin` with `short_display='2.68545'`, etc;  `eps` would not be given a `short_display`.  Just a thought--I haven't tried it, so I don't know if that would end up being messy to implement.

fredrik-johansson:
Yeah, you will pay the price eventually if you do some computation. Still it's surprising if you happen to have the precision set to 10000 digits and inspect

     >>> khinchin

just to remind yourself what this object is, only to find that viewing it practically hangs.

WarrenWeckesser:
There is another problem with this pull request: if the precision is really low, the value shown in the repr is the low precision value, e.g.

```
In [2]: mpmath.mp.dps = 2

In [3]: mpmath.pi
Out[3]: <pi: 3.14063~>
```

I don't think that is desirable, so I'm closing the PR.

Issue #495, opened 2019-11-20
Gegenbauer function within mp context throwing ValueError


jacksonvanover:
mpmath version: 1.1.0

For the same input, the gegenbauer function in the mp context throws a ValueError while the fp context version returns 0.0 which agrees with SciPy's output.

Example:
```
Python 3.6.8 (default, Aug 20 2019, 17:12:48) 
[GCC 8.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from mpmath import *
>>> from scipy import special
>>> fp.gegenbauer(1, 2.5, 0)
0.0
>>> special.eval_gegenbauer(1, 2.5, 0)
0.0
>>> mp.gegenbauer(1, 2.5, 0)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python3.6/dist-packages/mpmath/ctx_mp_python.py", line 1035, in f_wrapped
    retval = f(ctx, *args, **kwargs)
  File "/usr/local/lib/python3.6/dist-packages/mpmath/functions/orthogonal.py", line 326, in gegenbauer
    return ctx.hypercomb(h, [n], **kwargs)
  File "/usr/local/lib/python3.6/dist-packages/mpmath/functions/hypergeometric.py", line 127, in hypercomb
    [ctx.rgamma(b) for b in beta_s] + \
  File "/usr/local/lib/python3.6/dist-packages/mpmath/functions/hypergeometric.py", line 226, in hyper
    if   q == 1: return ctx._hyp2f1(a_s, b_s, z, **kwargs)
  File "/usr/local/lib/python3.6/dist-packages/mpmath/functions/hypergeometric.py", line 443, in _hyp2f1
    return ctx.hypsum(2, 1, (atype, btype, ctype), [a, b, c], z, **kwargs)
  File "/usr/local/lib/python3.6/dist-packages/mpmath/ctx_mp.py", line 708, in hypsum
    raise ValueError(ctx._hypsum_msg % (prec, prec+extraprec))
ValueError: hypsum() failed to converge to the requested 73 bits of accuracy
using a working precision of 3588 bits. Try with a higher maxprec,
maxterms, or set zeroprec.
```

skirpichev:
I don't think it's a real bug.  The exception message shows some hints you could use to compute the value, e.g.:
```pycon
>>> mp.gegenbauer(1, 2.5, 0, zeroprec=1000)
mpf('0.0')
```

[...]

Issue #296, opened 2014-09-24
improve accuracy of hypergeometric sums


rwst:
This is a placeholder for enhancements that try to improve the worst cases resulting in this error in Sage:

```
ValueError: hypsum() failed to converge to the requested 100 bits of accuracy
using a working precision of 7135 bits. Try with a higher maxprec,
maxterms, or set zeroprec.
```

A representative example would be http://ask.sagemath.org/question/24257/a-hypergeometric-series/

RylanSchaeffer:
I'm currently hitting this issue. Was it ever addressed?

RylanSchaeffer:
@skirpichev any advice on how to resolve this error?

skirpichev:
JFR, reproducer from the referenced thread:
```pycon
>>> A182122 = lambda n: 2*hyper([-n+1,2-n],[2],-1,zeroprec=1000)  # try with missing zeroprec!
```
I got a correct sequence, however.

I don't know a good solution in this case.  We can set "good enough" default `zeroprec`, but this may end in wrong answers.

RylanSchaeffer:
@skirpichev , I should have been more precise - my apologies. I meant that I'm hitting the above error, albeit in a different context.

In my context, I need to compute the hypergeometric function arising from a scaled Beta-Binomial compound distribution:

![image](https://github.com/user-attachments/assets/2a8ecabd-739b-49bf-bb57-386fa8d7e54f)

Here is a minimal reproducible example:

```
import mpmath
import numpy as np
from typing import Tuple


def compute_beta_binomial_three_parameters_distribution_neg_log_likelihood(
    params: Tuple[float, float, float],
    num_samples: np.ndarray,
    num_successes: np.ndarray,
) -> float:
    """
    3-parameter Beta-Binomial PMF using Gauss hypergeometric function:

    P(X=x) = binom(n, x) * [c^x / B(alpha, beta)] * B(x + alpha, beta) * _2F_1(arguments).
    """

    alpha, beta, scale = params
    nll_arr = np.zeros_like(num_samples, dtype=np.float64)
    for idx, (n, x) in enumerate(zip(num_samples, num_successes)):
        if not (0 <= x <= n):
            return 0.0

        # binomial coefficient binom(n, x)
        binom_factor = mpmath.binomial(int(n), int(x))

        # c^x
        c_to_x = mpmath.power(scale, x)

        # Beta(alpha, beta)
        B_a_b = mpmath.beta(alpha, beta)

[...]

Issue #14, opened 2014-02-06
mp.jacobi(1, 0, 0, 0) fails to evaluate


FreddieWitherden:
The with mpmath 0.18 the expression:

```
mp.jacobi(1, 0, 0, 0)
```

raises a ValueError exception on account of hypsum() failing to converge.  However, the equivalent expression:

```
mp.legendre(1, 0)
```

gives zero as expected.

skirpichev:
It looks, the problem is ``hyper([-1,2],[1], 0.5)``.  Like #296, it's a polynomial in z.  I'll close this issue as a duplicate.

PR #651, opened 2022-11-24
improved hypsum non-convergence behaviour

bfis:
This change includes two `hypsum` related fixes:
1. It turns the `ValueError` exception, raised in `mp.hypsum` when encountering convergence problems, into a more appropriate `NoConvergence` exception - just as `fp.hypsum` does already.
2. It properly handles this `NoConvergence` exception withing the "fast path" of [`hyp2f1`](https://github.com/mpmath/mpmath/blob/1258e33e1618a6c537947775a03a7746362c4cee/mpmath/functions/hypergeometric.py#L439-L443). The same behavior is already used in the "fast path"s of [`_hyp2f0`](https://github.com/mpmath/mpmath/blob/1258e33e1618a6c537947775a03a7746362c4cee/mpmath/functions/hypergeometric.py#L982-L988), [`_hypq1fq`](https://github.com/mpmath/mpmath/blob/1258e33e1618a6c537947775a03a7746362c4cee/mpmath/functions/hypergeometric.py#L496-L500), and [`_hyp_borel`](https://github.com/mpmath/mpmath/blob/1258e33e1618a6c537947775a03a7746362c4cee/mpmath/functions/hypergeometric.py#L694-L697) - and was broken since `hypsum` never raised the expected `NoConvergence` exception (in the `mp` context). Similar to `_hyp2f0`, this behavior can be suppressed with the keyword argument `force_series=True`.

I've also seen that similar issues may be affecting [`hypercomb`](https://github.com/mpmath/mpmath/blob/1258e33e1618a6c537947775a03a7746362c4cee/mpmath/functions/hypergeometric.py#L76), since it also raise `ValueError` instead of `NoConvergence` to signal convergence issues. If interested, I could also change that function appropriately.

I've verified in known pathological case (`mpmath.hyp2f1(3814.269794854973, -1985.4387491751415, 5369.1679108495828, 0.70000000000000007).ae('4.1656634124024037e-532')`) that both "fast path" (with raised `maxprec`) and the "slow path" (which is actually faster) bow yield the same result. If interested, I can also add this as a test case.

This change may positivity affect some of the cases alluded to in #296. For the issues #14 and #495 it shifts the convergence issue into the "slow path", which may be preferable.

### DIFF
--- a/mpmath/functions/hypergeometric.py
+++ b/mpmath/functions/hypergeometric.py
@@ -440,7 +440,11 @@ def _hyp2f1(ctx, a_s, b_s, z, **kwargs):
     # possibly in finitely many terms
     if absz <= 0.8 or (ctx.isint(a) and a <= 0 and a >= -1000) or \
                       (ctx.isint(b) and b <= 0 and b >= -1000):
-        return ctx.hypsum(2, 1, (atype, btype, ctype), [a, b, c], z, **kwargs)
+        try:
+            return ctx.hypsum(2, 1, (atype, btype, ctype), [a, b, c], z, **kwargs)
+        except ctx.NoConvergence:
+            if kwargs.get('force_series', False):
+                raise
 
     orig = ctx.prec
     try:
--- a/mpmath/ctx_mp.py
+++ b/mpmath/ctx_mp.py
@@ -705,7 +705,7 @@ maxterms, or set zeroprec."""
             max_total_jump += abs(d)
         while 1:
             if extraprec > maxprec:
-                raise ValueError(ctx._hypsum_msg % (prec, prec+extraprec))
+                raise ctx.NoConvergence(ctx._hypsum_msg % (prec, prec+extraprec))
             wp = prec + extraprec
             if magnitude_check:
                 mag_dict = dict((n,None) for n in magnitude_check)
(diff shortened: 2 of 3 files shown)


mpmath/functions/hypergeometric.py
```diff
@@ -440,7 +440,11 @@ def _hyp2f1(ctx, a_s, b_s, z, **kwargs):
     # possibly in finitely many terms
     if absz <= 0.8 or (ctx.isint(a) and a <= 0 and a >= -1000) or \
                       (ctx.isint(b) and b <= 0 and b >= -1000):
-        return ctx.hypsum(2, 1, (atype, btype, ctype), [a, b, c], z, **kwargs)
+        try:
+            return ctx.hypsum(2, 1, (atype, btype, ctype), [a, b, c], z, **kwargs)
+        except ctx.NoConvergence:
+            if kwargs.get('force_series', False):
+                raise
```
skirpichev, line 447: Please add a test for this case, see https://app.codecov.io/gh/mpmath/mpmath/pull/651

skirpichev (changes requested):

skirpichev:
Ok, I've cherry-picked this commit in #703.

PR #656, opened 2022-12-28
GitHub Actions: Test on Python 3.11 production release

cclauss:
https://docs.python.org/3.11/whatsnew/3.11.html
> Python 3.11 is between 10-60% faster than Python 3.10. On average, we measured a 1.25x speedup on the standard benchmark suite.

Downgrade from ubuntu-22.04 to ubuntu-20.04 for Python v3.5 and v3.6


cclauss:
@fredrik-johansson Your review, please.

PR #659, opened 2023-03-19
Fix running pytest

tornaria:
Running pytest doesn't work quite right. Changes here:
 - move all pytest configuration from setup.cfg to pytest.ini to avoid having different configurations
 - add . to pythonpath so the mpmath module here takes precedence in case mpmath is installed in system
 - fix norecursedirs: doc/source/plots -> docs/plots
 - add build to norecursedirs so pytest works even after building mpmath
 - add '*.rst' to globs so pytest picks a few tests in docs/*.rst

With this changes I can run `pytest` on a clean git clone. It will test this mpmath module (and not  the system-installed one, which is an older version). It will run all tests (not just the ones in`test_*.py`). Also, `pytest` still works after running `python setup.py build`. Adding this new `pytest.ini` to the extracted tarball from pypi also makes pytest work ok.

**Please ship `pytest.ini` with pypi tarballs so pytest works there.**

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -1,3 +1,10 @@
 [pytest]
+pythonpath = .
+doctest_optionflags = IGNORE_EXCEPTION_DETAIL
+addopts = --doctest-modules
+          --ignore=setup.py
+          --doctest-glob='*.txt'
+          --doctest-glob='*.rst'
+norecursedirs = docs/plots demo .eggs .git build
 markers =
     slow: marks tests as slow (deselect with '-m "not slow"')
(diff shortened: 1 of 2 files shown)


skirpichev:
> Running pytest doesn't work quite right.

It would be more helpful if you describe first what's exactly "doesn't work".

The documentation for installation/setup is not up-to date and it lacks instructions for a development version.  But it has instructions on how to run tests...
```
$ python3 -m venv ~/mpmath-venv
$ . ~/mpmath-venv/bin/activate
$ pip install -e .[gmpy,docs,tests]  # in the cloned tree
$ pytest mpmath docs  # ok
$ pytest --pyargs mpmath  # ok
$ cd
$ pytest --pyargs mpmath  # ok
$ pytest --pyargs mpmath -m slow  # ok
```
Same for non-editable installations.

So, I'm not sure if there is an issue with pytest.  Please fix conflicts an reopen, if you disagree.

PR #665, opened 2023-03-25
PEP8 fixes

cclauss:
Describe the tool `ruff`, how it finds three pep8 issues, and how it fixes those 3 issues.

[Ruff](https://beta.ruff.rs/) supports [over 500 lint rules](https://beta.ruff.rs/docs/rules) and can be used to replace [Flake8](https://pypi.org/project/flake8/) (plus dozens of plugins), [isort](https://pypi.org/project/isort/), [pydocstyle](https://pypi.org/project/pydocstyle/), [yesqa](https://github.com/asottile/yesqa), [eradicate](https://pypi.org/project/eradicate/), [pyupgrade](https://pypi.org/project/pyupgrade/), and [autoflake](https://pypi.org/project/autoflake/), all while executing (in Rust) tens or hundreds of times faster than any individual tool.

% `ruff --select=E703,E712,E713 --statistics .`
```
 12	E703	[*] Statement ends with an unnecessary semicolon
105	E712	[*] Comparison to `False` should be `cond is False`
 13	E713	[*] Test for membership should be `not in`
```
% ` ruff --select=E703,E712,E713 --fix .`
```
Found 130 errors (130 fixed, 0 remaining).
```

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -39,7 +39,7 @@ gmpy =
     gmpy2>=2.1.0a4; platform_python_implementation!="PyPy"
 docs = sphinx
 [pycodestyle]
-select = E10,E11,E13,E703,E71,E9,W1,W2,W3,W6
+select = E101,E111,E112,E113,E703,E712,E713,W191,W291,W292,W293,W391
 exclude = .eggs,.git
 [tool:pytest]
 testpaths = mpmath docs
--- a/setup.cfg
+++ b/setup.cfg
@@ -31,10 +31,9 @@ tests_require = mpmath[tests]
 [options.extras_require]
 tests = pytest >= 4.6
 develop = %(tests)s
-          codecov
           pycodestyle
           pytest-cov
-          ruff; python_version>"3.6"
+          codecov
           wheel
 gmpy =
     gmpy2>=2.1.0a4; platform_python_implementation!="PyPy"
--- a/setup.cfg
+++ b/setup.cfg
@@ -39,7 +39,7 @@ gmpy =
     gmpy2>=2.1.0a4; platform_python_implementation!="PyPy"
 docs = sphinx
 [pycodestyle]
-select = E101,W191,W291,W293,E111,E112,E113,W292,W391
+select = E101,W191,W291,W293,E111,E112,E113,W292,W391,E703,E712,E713
 exclude = .eggs,.git
 [tool:pytest]
 testpaths = mpmath docs
--- a/setup.cfg
+++ b/setup.cfg
@@ -31,9 +31,10 @@ tests_require = mpmath[tests]
 [options.extras_require]
 tests = pytest >= 4.6
 develop = %(tests)s
+          codecov
           pycodestyle
           pytest-cov
-          codecov
+          ruff; python_version>"3.6"
           wheel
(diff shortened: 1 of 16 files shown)


skirpichev (changes requested):
Seems ok, but please ensure, that these errors are tested in CI.  Could you also make commit message more descriptive?

pyproject.toml
skirpichev, line 1: No.  Lets just enable fixed errors/warnings for flake8.  Nothing fancy.

skirpichev (changes requested):

pyproject.toml
cclauss: Which changes do you want to be reverted?

skirpichev: pyproject.toml and .github/workflows/test.yml
Just enable fixed flake8 errors in the setup.cfg.

cclauss: Given [PEP 621](https://peps.python.org/pep-0621), Ruff and other modern tools no longer read `setup.cfg` files.

skirpichev: **Just add fixed flake8 codes to the setup.cfg.**  No ruff, no pyproject.toml, no changing test.yml

skirpichev: **Please enable fixed errors/warnings in the setup.cfg.**

skirpichev (approved):
Few unrelated changes, but LGTM.
Let's wait few days if someone will object against changes.  Code style changes always debatable.

pyproject.toml
skirpichev: Ok. I took the liberty to fix this issue by myself.

setup.cfg
```diff
@@ -39,7 +39,7 @@ gmpy =
     gmpy2>=2.1.0a4; platform_python_implementation!="PyPy"
 docs = sphinx
 [pycodestyle]
-select = E101,W191,W291,W293,E111,E112,E113,W292,W391
+select = E10,E11,E13,E703,E71,E9,W1,W2,W3,W6
```
skirpichev, line 42: No, please don't do this.  Feel free to enable more checks (already fixed;)), but lets do this explicitly.  I don't think we want to break our CI if flake8 people just add some new E9XY error with the new release.

skirpichev (changes requested):

setup.cfg
cclauss: The idea behind linters is that we set up guardrails so that if contributors break rules that the codebase already passes those violations can be discussed in code review.

cclauss, line 42:
```suggestion
select = E101,E111,E112,E113,E703,E712,E713,W191,W291,W292,W293,W391
```

skirpichev: > if contributors break rules that the codebase already passes

Exactly.  Not something that just coming with a new release of the linter.

oscarbenjamin: Another possibility is to pin the version of the linter.

skirpichev: New release might fix bugs in existing checks.  Probably, we want this and it shouldn't be too annoying, i.e. break too often and too much.  (Such fixes usually cover some corner cases and so on...)  But if it will - we can consider pinning.

skirpichev (approved):

PR #674, opened 2023-03-29
Use codecov/codecov-action

skirpichev:
FYI: https://github.com/codecov/codecov-python


.github/workflows/test.yml
```diff
@@ -61,8 +64,8 @@ jobs:
           coverage.xml
           build/coverage/html/
     - name: Publish package on PyPI
-      if: matrix.python-version == 3.9 && github.event_name == 'push' && startsWith(github.event.ref, 'refs/tags')
-      uses: pypa/gh-action-pypi-publish@v1.4.2
```
cbm755, line 59: I don't know this particular GitHub Action enough to know if this is a good change or not, but it does seem unrelated to the commit message.

cbm755: It looks like the new form here (`@release/v1`) is the current recommendation from upstream so I think it is a good change.  But probably it should be a deliberate change.

cbm755 (commented):
I think this looks good, although I have no experience with coverage on GitHub (on GitLab its integrated into the web UI, here it looks like an external service is being used).

Perhaps before it was computing coverage but not uploading the artifacts elsewhere and this change makes it so the artifacts are uploaded to the third-party for viewing.

I cannot speak to whether that is wanted or not but this looks like it does accomplish that.

skirpichev:
FYI: coverage uploaded to https://app.codecov.io/gh/mpmath/mpmath, only status checks aren't enabled (I can't do this without Fredrik).  But sometimes it's useful to have html/xml coverage data (e.g. codecov.io has bugs).

Now pypi things go to a separate commit and no more exclusions for doctests.

cbm755:
Thanks for explanation.  I have now learned what app.codecov.io is.  I don't understand how it was uploading there before this change, but it appears it somehow was.

Thus this is not a change from the project to adopt a new service but rather one it was already using. 

Is that correct?  If so, I have no objections.

(OTOH, if the previous CI was *not* uploading artifacts to codecov.io, then I think this is something that Fredrik would have to approve).

cbm755:
>  that Fredrik would have to approve).

I should say "someone other than me would have to approve".

skirpichev:
> Thus this is not a change from the project to adopt a new service

Nop.  Just little info for debugging code coverage.  (And BTW the coverage html output is more informative than the codecov.io)

cbm755 (approved):
My concerns addressed, +1 to merge.

PR #675, opened 2023-03-30
Drop support for Python versions < 3.8

### LINKED ISSUE #99: multithreaded runtests.py
bc.. It should be trivial to run several tests at once using Parallel Python or
something else. py.test offers such a feature already, so maybe we should
use it instead of a own solution.

p. Original issue for "#99":https://github.com/fredrik-johansson/mpmath/issues/99: "http://code.google.com/p/mpmath/issues/detail?id=59":http://code.google.com/p/mpmath/issues/detail?id=59

p. Original author: "https://code.google.com/u/Vinzent.Steinberg@gmail.com/":https://code.google.com/u/Vinzent.Steinberg@gmail.com/

p. Original owner: "https://code.google.com/u/Vinzent.Steinberg@gmail.com/":https://code.google.com/u/Vinzent.Steinberg@gmail.com/

### LINKED ISSUE #663: Drop support for old py3 versions (<3.8)
I think, drop everything <3.8 is fine (CPython 3.7 is provided by the Debian Buster, as an example).  @asmeurer I guess it's ok for SymPy?

skirpichev:
Closes #663

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -6,7 +6,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: [3.5, 3.6, 3.7, 3.8, 3.9, '3.10', 3.11, pypy3.7, pypy3.8, pypy3.9]
+        python-version: [3.8, 3.9, '3.10', 3.11, pypy3.8, pypy3.9]
     steps:
     - uses: actions/checkout@v3
     - name: Set up Python ${{ matrix.python-version }}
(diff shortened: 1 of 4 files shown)


cbm755 (approved):
Looks reasonable to me, certainly don't see much point supporting before 3.7.   Ubuntu 20.04 is on 3.8, and is used in lots of folks containers and this keeps support for that.

There is a burden in supporting old things, even if this still work.  For example contributors might use f-strings and not notice until CI fails, thus creating more work.

skirpichev:
Ok, this a tiny change so far (no code conversion to an incompatible syntax) and could be easily reverted (or partially reverted).  Let's merge this.

Ping @fredrik-johansson, FYI.

PR #678, opened 2023-04-01
prevent erroneous setting of dps/prec on mpmath module

### LINKED ISSUE #657: ENH: guard against incorrect use of `mp.dps`?
I am a SciPy maintainer, and we frequently use mpmath to compute reference values. (Thank you for your work! It's very helpful.) 

A common mistake is for contributors to do, e.g.:

```python3
import numpy as np
import mpmath as mp  # should be from mpmath import mp
mp.dps = 50
a, b = mp.mpf(1e-11), mp.mpf(1.001e-11)
print(np.float64(mp.ncdf(b) - mp.ncdf(a)))  # 3.885780586188048e-15
```

This suffers from catastrophic cancellation just as:
```python3
from scipy.special import ndtr
a, b = 1e-11, 1.001e-11
print(ndtr(b) - ndtr(a))  # 3.885780586188048e-15
```
does, but the fact that `mpmath` is _not_ working with 50 digits is obscured by the conversion to float. (I've run into other, subtler reasons for not noticing the problem, too.)

Another example in the wild: https://github.com/scipy/scipy/issues/18088#issuecomment-1712538038

I know that this is user error and not a bug in `mpmath`, but I wonder if it is possible to guard against this. For instance, `mpmath` doesn't currently have a `dps` attribute. Perhaps it could be added as a property, and a warning (or error) could be raised if modification is attempted, directing the user to `mpmath.mp.dps`?

cbm755:
~~What we really want for Issue #657 is an exception on write access. Instead here is a partial implementation for read access.  Not even close to a solution sadly.~~

**update**: exceptions on write access to `.dps` and `.prec` in `mpmath` module.

- - - -

### Todo list

- [x] ~~depends on #663~~
- [x] release notes?
- [x] test
- [x] ~~any performance concerns?~~

### potential changelog entry

My feeling is this maybe doesn't need a changelog, but just in case here's one:

```
Bug fixes:

  * Prevent accidentally setting `dps`/`prec` on the mpmath module (Colin Macdonald, Sergey Kirpichev, Warren Weckesser)
```

Closes #657


mpmath/__init__.py
```diff
@@ -19,6 +19,17 @@
...
 
+
+def __getattr__(name):
+    if name == "dps":
+        raise AttributeError(
+            f"Module {__name__!r} does not have attribute {name!r}; "
+            "you may want 'from mpmath import mp', see Issue #657"
+        )
+    # TODO: how to pass through to access any of the stuff in here?
+    # Maintaining `__all__` would be a pain...
```
skirpichev, line 30: IIUIC, You should raise here just a usual AttributeError:)

mpmath/__init__.py
```diff
@@ -19,6 +19,17 @@
 fp._iv = iv
 iv._iv = iv
 
+
+def __getattr__(name):
+    if name == "dps":
+        raise AttributeError(
+            f"Module {__name__!r} does not have attribute {name!r}; "
```
skirpichev, line 26: This won't work for 3.5, but don't worry.

skirpichev (changes requested):
I doubt this will fix the issue.  Misunderstanding starts with setting dps/prec attributes...

mpmath/__init__.py
```diff
@@ -19,6 +19,17 @@
 fp._iv = iv
 iv._iv = iv
 
+
+def __getattr__(name):
+    if name == "dps":
```
skirpichev, line 24: And prec.

mpmath/__init__.py
cbm755: Yeah forgot to say the PEP 562 needs Python 3.7

mpmath/__init__.py
skirpichev, line 30:
```suggestion
    raise AttributeError(f"Module {__name__!r} has no attribute {name!r})
```

cbm755:
This really doesn't work very well, not sure I understand why:
```
>> import mpmath as mp
>> mp.dps
AttributeException(...)
>> mp.dps = 64
>> mp.dps
64
```
eh?  I was expecting an exception.  So unless I've done this wrong PEP 562 really doesn't offer much control here.

mpmath/__init__.py
skirpichev, line 24:
```suggestion
    if name in ("dps", "prec"):
```

mpmath/__init__.py
cbm755: Ah, then I have misunderstood what this does.  I take it `__getattr__` is only called when the attribute does not exist :(

skirpichev:
> This really doesn't work very well, not sure I understand why:
> 
> ```
> >> import mpmath as mp
> >> mp.dps
> AttributeException(...)
> >> mp.dps = 64
> >> mp.dps
> 64
> ```
> eh? I was expecting an exception.

Unfortunately, this is not like the property attribute on classes.  If 'dps' exists in the namespace of the module - the `__getattr__()` helper will not be invoked.

> So unless I've done this wrong

Certainly not.

mpmath/__init__.py
```diff
@@ -1,5 +1,8 @@
 __version__ = '1.3.0'
 
+import sys
+import types
```
skirpichev, line 4: This should be deleted afterwards, or:
```python
>>> import mpmath
>>> 'types' in dir(mpmath)
```

mpmath/__init__.py
```diff
@@ -432,6 +436,34 @@
...
+
+    def _set_dps(self, value):
+        raise SetMPMathPrecisionError('dps')
+
+    dps = property(fset=_set_dps)
+
+    def _set_prec(self, value):
+        raise SetMPMathPrecisionError('prec')
+
+    prec = property(fset=_set_prec)
```
skirpichev, line 461: I've not seen this in issues, but there are examples of setting also `mp.pretty attribute in docs.  Also, `mp.trap_complex` is mentioned.  I think we should add at least these properties as well.

mpmath/__init__.py
```diff
@@ -432,6 +436,34 @@
 unit_triangle = mp.unit_triangle
 sigmoid = mp.sigmoid
 
+
+class SetMPMathPrecisionError(RuntimeError):
```
skirpichev, line 440: AttributeError?

skirpichev: Here is a version without this class (nothing goes to the global mpmath namespace):
```diff
diff --git a/mpmath/__init__.py b/mpmath/__init__.py
index 46a7c6f..6fcaaae 100644
--- a/mpmath/__init__.py
+++ b/mpmath/__init__.py
@@ -1,5 +1,9 @@
 __version__ = '1.3.0'
 
+import functools
+import sys
+import types
+
 from .usertools import monitor, timing
 
 from .ctx_fp import FPContext
@@ -432,6 +436,20 @@
 unit_triangle = mp.unit_triangle
 sigmoid = mp.sigmoid
 
+
+class _MPMathModuleType(types.ModuleType):
+
+    def _helper(self, *args, prop=''):
+        raise AttributeError("cannot set '{name}' on 'mpmath'. Did you mean to "
+                             "set '{name}' on 'mpmath.mp'?".format(name=prop))
+
+    dps = property(fset=functools.partial(_helper, prop='dps'))
+    prec = property(fset=functools.partial(_helper, prop='prec'))
+
+
+sys.modules[__name__].__class__ = _MPMathModuleType
+del functools, types, sys
+
 # be careful when changing this name, don't use test*!
 def runtests():
     """
```

skirpichev (changes requested):

skirpichev:
It seems, release notes (CHANGES) update happens just before release.  Probably, it's a good idea to automate this and keep things more up to date.  But don't worry here.

Yes, lets add tests.  Shouldn't be hard.

I don't think there is any performance penalty.  @oscarbenjamin?

mpmath/__init__.py
cbm755: yes, I like this better too

skirpichev (approved):

cbm755:
I don't know if this is significant *here* or not...  I figure it potentially matters when calling frunctions from `mpmath` in a loop:

`e.py`:
```python
import mpmath

def test1():
    f = mpmath.sin
    for i in range(0, 1000):
        f(i / 10)

def test2():
    for i in range(0, 1000):
        mpmath.sin(i / 10)
```

```
$ python -m timeit -r11 -s 'from e import test1' 'test1()'
100 loops, best of 11: 3.21 msec per loop
$ python -m timeit -r11 -s 'from e import test2' 'test2()'
100 loops, best of 11: 3.25 msec per loop
```

So its insignficant there, compared to the computation cost. 

Of course, if I remove the computation, then I see the same results as @skirpichev:

```python
import mpmath

def test1():
    f = mpmath.sin
    for i in range(0, 1000):
        f

def test2():
    for i in range(0, 1000):
        mpmath.sin
```

```
[cbm@x1c10 mpmath.git]$ python -m timeit -r11 -s 'from e2 import test1' 'test1()'
50000 loops, best of 11: 8.86 usec per loop
[cbm@x1c10 mpmath.git]$ python -m timeit -r11 -s 'from e2 import test2' 'test2()'
10000 loops, best of 11: 22.6 usec per loop
```

So is there some situation where one would tight-loop access to an attribute of `mpmath` without doing any real work in the loop?

Even if we cannot think of a "real life" example here, it definitely sounds like a point to make if you submit a PEP!

cbm755:
Here's an example, only one letter different `sign` instead of `sin`, which is *much* cheaper:

```python
import mpmath

def test1():
    f = mpmath.sign
    a = 0
    for i in range(0, 1000):
        a += f(i)

def test2():
    a = 0
    for i in range(0, 1000):
        a += mpmath.sign(i)
```

```
$ python -m timeit -r11 -s 'from e import test1' 'test1()'
50000 loops, best of 11: 9.11 usec per loop
$ python -m timeit -r11 -s 'from e import test2' 'test2()'
10000 loops, best of 11: 22.2 usec per loop
```
And we see the slow down in a vaguely real-life example...

skirpichev:
> So is there some situation where one would tight-loop access to an attribute of mpmath without doing any real work in the loop?

Something like `type(foo) is mpmath.mpf` or `isinstance(foo, mpmath.mpf)`.

> an example, only one letter different sign instead of sin, which is much cheaper ... And we see the slow down in a vaguely real-life example...

Of course, all this can be workarounded by using test1() layout (or `from mpmath import sign as f`) instead of test2()...  So, I'm still in favor of merging this.  Let's wait few days in case  someone else will review this.

> it definitely sounds like a point to make if you submit a PEP!

Unfortunately, I don't see too much usage examples for this "feature".  And mpmath's case seems to be *very* special.  What else?  To prevent mutations of the module `__dict__`, e.g. `mpmath.sin="spam"`?:)  On another hand, PEP 562 has not too much as well... (BTW, I don't see where stuff from that PEP was documented.)

oscarbenjamin:
I think that replacing the module in `sys.modules` like this is a bad idea. It is messing around with things that should not really be messed with and will cause something to break. I am certain that this change would end up being regretted (this looks like precisely the kind of thing that I would want to remove from a codebase if I saw it).

Being able to define `__setattr__` on a module would be a useful change to the Python language. In the absence of a properly supported way to prevent attributes being set on a module though it is better just to accept that this can happen even if it is unfortunate.

It would be better to focus on designing an interface for precision control that is less prone to mistakes and misunderstanding.

skirpichev:
> will cause something to break.

Could you provide an example?  We can't reject pr only on the ground of unspecified potential problems.

> Being able to define `__setattr__` on a module would be a useful change to the Python language.

I doubt it will be accepted.  (As I said above, there are no good practical examples of why it could be helpful.)  But if it will - we can switch the present workaround to that solution.

> In the absence of a properly supported way to prevent attributes being set on a module 

Actually, Python does support customizing module attribute access, it is only slow.  The recommended way in [the CPython docs](https://docs.python.org/3/reference/datamodel.html#customizing-module-attribute-access) is exactly how things are implemented in this pr.

skirpichev:
FYI: https://discuss.python.org/t/extend-pep-562-with-setattr-for-modules/25506

skirpichev:
@fredrik-johansson, how this solution looks for you?

skirpichev:
I think this could be merged as-is, I will do this on monday.  Unfortunately, from the [discussion thread](https://discuss.python.org/t/extend-pep-562-with-setattr-for-modules/25506) it seems for me, that proposing a PEP for `__setattr__` helper doesn't make sense.

1. There is some scenarios where measurable code slowdown exits, but I doubt this will pose real problems for someone.  And there are simple workarounds.  
2. @oscarbenjamin, if you have any clue on what this may "break" - please show us.  BTW, similar replacing of the module in `sys.modules` was used for a long time in the scipy (FFT submodule).
3. @asmeurer, I would appreciate your opinion.

oscarbenjamin:
> 2\. if you have any clue on what this may "break" - please show us

I don't have any specific examples. My experience is that when you do something like this it seems like you can reason about whether it should work but then something somewhere else turns out to have made some assumptions that will be broken as a result of the change.

I think it would be better to reflect on whether setting a global dps attribute is actually a good interface in the first place (I don't think it is) rather than making the change here to give an error for someone who sets the attribute on the wrong object. After that I would consider whether the docs can be improved so that the mistake is less likely. After that I would just do nothing. It isn't ultimately necessary to do anything in response to gh-657: sometimes an issue is just an issue but does not need to be "fixed".

cbm755:
Perhaps we should just leave this open, then see if such an alternative interface as @oscarbenjamin describes) can be developed.  If no such alternative materializes then we can revisit this, say in 6 months or so.

To be concrete: if the docs in the dev tree are are still recommending `mp.dps = 64` in 2024, then I suggest merging this.

skirpichev:
> I don't have any specific examples.

Good news.  Then I don't see technical objections: this pr does very tiny and fine grained customization of the module behavior, exactly on the way the official docs recommends.

> I think it would be better to reflect on whether setting a global dps attribute is actually a good interface in the first place (I don't think it is)

I think we all agreed that the context management could be better.  But this is much more complex change (and probably with API breakage).

> It isn't ultimately necessary to do anything in response to https://github.com/mpmath/mpmath/issues/657

This is a common gotcha for mpmath users (2-3 duplicates of this issue were mentioned in the issue thread), that's why I would like to include some fix in the next (minor?) release.

Alternatively, we can add something to the docs (as a note/warning or a dedicated section), but this solution seems much better for me: it just forbids wrong scenarios of altering context attributes.

asmeurer:
I think it's worth at least trying out https://github.com/mpmath/mpmath/issues/657#issuecomment-1498195928. It's much less invasive. I think if we put the check in the right place it won't be a major performance issue.

skirpichev:
> To be concrete: if the docs in the dev tree are are still recommending mp.dps = 64 in 2024, then I suggest merging this.

Ok.

oscarbenjamin:
> if the docs in the dev tree are are still recommending `mp.dps = 64` in 2024, then I suggest merging this.

I don't see any reason to wait until 2024: if you're going to do it then just do it now. Personally I would not do this now or later because it would be better to fix the interface design rather than add hacky workarounds to make the current interface slightly less confusing. Especially there should be no noticeable performance penalty for users who are using mpmath correctly just to provide better error messages to someone who uses it incorrectly.

It would be better to have an interface that does not depend on either attribute access or setting global variables at all and the docs could easily recommend that:
```python
from mpmath import make_context
mp = make_context(real=True, dps=100)
```

Note this from the README:

> # 4. Known problems
> Mpmath is a work in progress. Major issues include:
> ...
> - The interface for switching precision and rounding is not finalized. The current method is not threadsafe.

I suggest that we think about what is (or could be) a good recommended way for users to use `mpmath` in the first instance and then consider what to recommend in the docs. Once that is done we might find that this issue is just not an issue any more.

skirpichev:
> it would be better to fix the interface design

I've opened an issue #683

> Especially there should be no noticeable performance penalty

I don't think there is a practical performance penalty.

> at least trying out https://github.com/mpmath/mpmath/issues/657#issuecomment-1498195928. It's much less invasive. I think if we put the check in the right place

@asmeurer, I doubt there is such "right place".  You will have to do it differently for every context (mp, fp, etc).  E.g. for the mp context you, probably, should adapt defun* decorators, for the fp - `_mathfun()` helper and so on.  I would welcome a pr, but it will definitely will be more complex than the current solution.

skirpichev:
https://peps.python.org/pep-0726/

skirpichev:
> To be concrete: if the docs in the dev tree are are still recommending mp.dps = 64 in 2024, then I suggest merging this.

Ok, I'm going to merge this in few days.  Regardless on the PEP fate (https://github.com/python/steering-council/issues/216) - we don't have yet a better workaround for the issue.

skirpichev:
Thanks

mdhaber:
Thanks, all!

Issue #683, opened 2023-04-17
Create a better interface for switching precision and rounding


skirpichev:
The mpmath provides global [context objects](https://mpmath.readthedocs.io/en/latest/contexts.html) (mp, fp, iv).  Most functions in the global mpmath namespace are actually methods of the mp context.

It would be better, if we could introduce an explicit notion of the current context, that will prevent issues like #657, something more close to [gmpy2 contexts](https://gmpy2.readthedocs.io/en/latest/contexts.html) handling.  (Another example that worth considering is the [figfloat](https://bigfloat.readthedocs.io/en/latest/index.html) package.)

The current interface also is not thread-safe.

oscarbenjamin:
The thread-safe part can be addressed in a minimal way by ensuring that either the contexts or the attributes of the context use thread-local variables:
https://docs.python.org/3/library/threading.html#thread-local-data

That would entail some slowdown on attribute access but it is not clear how significant it is in the context of everything else. The slowdown of thread-local data access in Python is not as much as it would be when working in something like C which has less general overhead.

A thread-local is really just a global variable that is per-thread rather than being global to all threads. It is thread-safe but still otherwise has all of the pitfalls of global variables so it would be good to make it clear in the docs how to avoid using this altogether at least for downstream library code that really should not depend on using global state like this.

It is broadly possible to just use local contexts:
```python
In [30]: from mpmath import MPContext

In [31]: mp1 = MPContext()

In [32]: mp2 = MPContext()

In [33]: mp1.dps = 4

In [34]: mp2.dps = 8

In [35]: mp1.cos(1)
Out[35]: mpf('0.5403061')

In [36]: mp1.cos(1) + mp1.cos(1)
Out[36]: mpf('1.080612')

In [37]: mp2.cos(1) + mp2.cos(1)
Out[37]: mpf('1.080604611')
```
This works because the mpf instances store a live reference to the precision of the context that created them:
```python
In [43]: f1 = mp1.cos(1)

In [44]: f2 = mp2.cos(1)

In [45]: f1._ctxdata
Out[45]: 
[mpmath.ctx_mp_python.mpf,
 <function object.__new__(*args, **kwargs)>,
 [17, 'n']]

In [46]: f2._ctxdata
Out[46]: 
[mpmath.ctx_mp_python.mpf,
 <function object.__new__(*args, **kwargs)>,
 [30, 'n']]

In [47]: f1.context
[...]

PR #685, opened 2023-04-17
Run CI tests with pytest-xdist

### LINKED ISSUE #99: multithreaded runtests.py

skirpichev:
Closes #99

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -7,6 +7,8 @@ jobs:
       fail-fast: false
       matrix:
         python-version: [3.8, 3.9, '3.10', 3.11, pypy3.8, pypy3.9]
+    env:
+      PYTEST_ADDOPTS: -n auto
     steps:
     - uses: actions/checkout@v3
     - name: Set up Python ${{ matrix.python-version }}
@@ -21,6 +23,7 @@ jobs:
       run: |
         pip install --upgrade --force-reinstall setuptools pip
         pip install --upgrade .[develop,gmpy,docs]
+        pip install --upgrade pytest-xdist
     - name: Linting with pycodestyle, etc
       run: |
         python -We:invalid -m compileall -f mpmath -q
@@ -33,7 +36,7 @@ jobs:
       run: pip uninstall -y gmpy2
     - name: Run coverage tests
       env:
-        PYTEST_ADDOPTS: --cov mpmath
+        PYTEST_ADDOPTS: --cov mpmath -n auto
       if: matrix.python-version == 3.10 || matrix.python-version == 3.11
       run: |
         pytest
--- a/conftest.py
+++ b/conftest.py
@@ -1,0 +1,14 @@
+import pytest
+
+
+def pytest_configure(config):
+    config.addinivalue_line('markers', 'slow: marks tests as slow')
+
+
+@pytest.fixture(autouse=True)
+def reset_mp_globals():
+    from mpmath import mp, iv
+    mp.dps = 15
+    mp.pretty = False
+    iv.dps = 15
+    iv.pretty = False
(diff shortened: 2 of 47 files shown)


oscarbenjamin:
Obviously it is useful to be able to run the tests in parallel locally so this is good.

When I've tried to use `pytest-xdist` before in CI though it hasn't seemed to bring the expected speed up in CI itself. I think GA runners only have two cores and I guess there's just too much overhead in the main thread to get the benefit of distributing the work over two threads. I guess it doesn't hurt though and it is good to ensure that being able to run the tests in parallel is itself tested in CI.

One potential downside is that if the tests fail because of something hanging then it can be difficult to know from the xdist output which test has failed which is particularly awkward if the failure is seen in CI but not locally. I haven't found anything better for this than using pytest's verbose mode to print out each test name as it starts and finishes.

skirpichev:
> it hasn't seemed to bring the expected speed up in CI itself

Well, we have 5-8 min in this pr for py3.10/11 and 8-11 in the master.

> I guess it doesn't hurt though and it is good to ensure that being able to run the tests in parallel is itself tested in CI.

And we can drop workarounds for resetting globals (dps, prec, etc).

> One potential downside is that if the tests fail because of something hanging then it can be difficult to know from the xdist output which test has failed

I don't think it's a special downside of the xdist plugin.  Same will be for plain pytest.

> I haven't found anything better for this than using pytest's verbose mode to print out each test name as it starts and finishes.

Probably we could use the pytest-timeout plugin with some "big enough" default.

oscarbenjamin:
> Well, we have 5-8 min's in this pr for py3.10/11 and 8-11 in the master.

Okay, that's good. That didn't happen when I tried before.

> I don't think it's a special downside of the xdist plugin. Same will be for plain pytest.

No because plain pytest shows which test module it is currently working on and how many tests have already run like:
```
mpmath/usertools.py .                                                    [ 16%]
mpmath/visualization.py .                                                [ 17%]
mpmath/calculus/approximation.py ..                                      [ 17%]
mpmath/calculus/differentiation.py .....
```
With xdist the default output is just
```
gw0 I / gw1 I
gw0 [632] / gw1 [632]

........................................................................ [ 11%]
........................................................................ [ 22%]
........................................................................ [ 34%]
........................................................................ [ 45%]
........................................................................ [ 56%]
........................................................................ [ 68%]
........................................................................ [ 79%]
.......................................x.......s...................x.... [ 91%]
........................................................                 [100%]
```
If one of the two workers hangs then the other will complete the test suite and it will just show that it's stuck at 99%.

> Probably we could use the pytest-timeout plugin with some "big enough" default.

That's probably a better idea. None of the tests here take that long.

oscarbenjamin:
> And we can drop workarounds for resetting globals (dps, prec, etc).

Maybe it would be better not to have tests/doctests that modify the globals in the first place (i.e. gh-683).

skirpichev:
> mpmath/calculus/differentiation.py .....
> vs
> ........................................................                 [100%]

Neither variant shows a test that hangs.

> That's probably a better idea.

But I doubt we should add such dependency right now.

> Maybe it would be better not to have tests/doctests that modify the globals in the first place (i.e. https://github.com/mpmath/mpmath/issues/683).

Sure, but that's a different issue.

Issue #507, opened 2020-01-15
Dropped NaN in hyp1f1 function


jacksonvanover:
mpmath version: 1.1.0

When the third argument is a NaN, it is possible to choose first and second arguments such that the NaN is dropped and a 1.0 is returned. From some very limited trials, it seems like a zero as a first argument consistently drops the NaN as well as a first arg of 1 and a second arg greater than 1.

```
Python 3.6.8 (default, Oct  7 2019, 12:59:55) 
[GCC 8.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from mpmath import *     
>>> import numpy as np
>>> mp.hyp1f1(0, 1, np.nan)
mpf('1.0')
>>> mp.hyp1f1(1, 1, np.nan)
mpf('nan')
>>> mp.hyp1f1(1, 1.1, np.nan)
mpf('1.0')
>>> mp.hyp1f1(1, 2, np.nan)
mpf('1.0')
>>> mp.hyp1f1(1, 3, np.nan)
mpf('1.0')
>>> mp.hyp1f1(1, 4, np.nan)
mpf('1.0')
>>> mp.hyp1f1(2, 1, np.nan)
mpf('1.0')
>>> mp.hyp1f1(2, 2, np.nan)
mpf('nan')
>>> mp.hyp1f1(0, 2, np.nan)
mpf('1.0')
>>> mp.hyp1f1(0, 4, np.nan)
mpf('1.0')
```

Issue #489, opened 2019-11-12
Unexpected Result from hyp0f1 with NaN input


jacksonvanover:
mpmath version: 1.1.0

When the second argument for hyp0f1 in the mp context is a NaN, a 1 is returned
regardless of the other input. For what it's worth, the fp context
version throws a NoConvergence Exception. For the same input, GSL and
SciPy both agree on NaN as an output.

Example:
```
Python 3.6.8 (default, Aug 20 2019, 17:12:48) 
[GCC 8.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from mpmath import *
>>> import numpy as np
>>> mp.hyp0f1(2.5, np.nan)
mpf('1.0')
>>> mp.hyp0f1(25, np.nan)
mpf('1.0')
>>> mp.hyp0f1(2513, np.nan)
mpf('1.0')
>>> mp.hyp0f1(.25, np.nan)
mpf('1.0')
```

Issue #488, opened 2019-11-12
Unexpected Result from hyp1f1 with NaN input


jacksonvanover:
mpmath version: 1.1.0

When the third argument for hyp1f1 in the mp context is a NaN, a 1 is returned
regardless of the other input. For what it's worth, the fp context
version throws a NoConvergence Exception. For the same input, SciPy
and GSL agree on an output of NaN.

Example:
```
Python 3.6.8 (default, Aug 20 2019, 17:12:48) 
[GCC 8.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from mpmath import *
>>> import numpy as np
>>> mp.hyp1f1(2.5,2.2,np.nan)
mpf('1.0')
>>> mp.hyp1f1(1,2.2,np.nan)
mpf('1.0')
>>> mp.hyp1f1(1,2002.2,np.nan)
mpf('1.0')
```

Issue #485, opened 2019-11-06
Hang in legendre function with NaN argument


jacksonvanover:
mpmath version: 1.1.0

When the second argument to the legendre function is a NaN, a hang is
caused. This is true for both the fp and the mp contexts.

Example:
```
Python 3.6.8 (default, Oct  7 2019, 12:59:55) 
[GCC 8.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from mpmath import *
>>> import numpy as np
>>> mp.legendre(1.2, np.nan)
^CTraceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/jvanover/.local/lib/python3.6/site-packages/mpmath/ctx_mp_python.py", line 1035, in f_wrapped
    retval = f(ctx, *args, **kwargs)
  File "/home/jvanover/.local/lib/python3.6/site-packages/mpmath/functions/orthogonal.py", line 363, in legendre
    return ctx.hyp2f1(-n,n+1,1,(1-x)/2, **kwargs)
  File "/home/jvanover/.local/lib/python3.6/site-packages/mpmath/functions/hypergeometric.py", line 251, in hyp2f1
    return ctx.hyper([a,b],[c],z,**kwargs)
  File "/home/jvanover/.local/lib/python3.6/site-packages/mpmath/functions/hypergeometric.py", line 226, in hyper
    if   q == 1: return ctx._hyp2f1(a_s, b_s, z, **kwargs)
  File "/home/jvanover/.local/lib/python3.6/site-packages/mpmath/functions/hypergeometric.py", line 473, in _hyp2f1
    v = _hyp2f1_gosper(ctx,a,b,c,z,**kwargs)
  File "/home/jvanover/.local/lib/python3.6/site-packages/mpmath/functions/hypergeometric.py", line 380, in _hyp2f1_gosper
    kakbz = (k+a)*(k+b)*z / (4*(k+1)*kch*(k+c1h))
  File "<string>", line 7, in __mul__
  File "/home/jvanover/.local/lib/python3.6/site-packages/mpmath/libmp/libmpf.py", line 894, in python_mpf_mul
    return normalize1(sign, man, sexp+texp, bc, prec, rnd)
  File "/home/jvanover/.local/lib/python3.6/site-packages/mpmath/libmp/libmpf.py", line 230, in _normalize1
    if not man & 1:
KeyboardInterrupt
```

PR #688, opened 2023-04-19
Fix nan handling in fp.mag() and hyper()

skirpichev:
Closes https://github.com/mpmath/mpmath/issues/485
Closes https://github.com/mpmath/mpmath/issues/479
Closes https://github.com/mpmath/mpmath/issues/507
Closes https://github.com/mpmath/mpmath/issues/489
Closes https://github.com/mpmath/mpmath/issues/488
Closes https://github.com/mpmath/mpmath/issues/478

### DIFF
--- a/mpmath/ctx_fp.py
+++ b/mpmath/ctx_fp.py
@@ -156,7 +156,10 @@ class FPContext(StandardBaseContext):
 
     def mag(ctx, z):
         if z:
-            return ctx.frexp(abs(z))[1]
+            n, e = ctx.frexp(abs(z))
+            if e:
+                return e
+            return ctx.convert(n)
         return ctx.ninf
 
     def isint(ctx, z):
--- a/mpmath/functions/hypergeometric.py
+++ b/mpmath/functions/hypergeometric.py
@@ -196,6 +196,8 @@ def hyper(ctx, a_s, b_s, z, **kwargs):
     Hypergeometric function, general case.
     """
     z = ctx.convert(z)
+    if ctx.isnan(z):
+        return ctx.nan
     p = len(a_s)
     q = len(b_s)
     a_s = [ctx._convert_param(a) for a in a_s]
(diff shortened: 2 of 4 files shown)


mpmath/ctx_fp.py
```diff
@@ -200,7 +203,7 @@ def hypsum(ctx, p, q, types, coeffs, z, maxterms=6000, **kwargs):
             for i in num: t *= (coeffs[i]+k)
             for i in den: t /= (coeffs[i]+k)
             k += 1; t /= k; t *= z; s += t
-            if abs(t) < tol:
+            if not abs(t) >= tol:
```
dimpase, line 206: Why is this needed?

skirpichev: Yes.

skirpichev: BTW, maybe we should explicitly test for nan's (as b0b5446 does), not as here or in 0d97f93.  Or redo the while statement like 064e940.

dimpase: Please check for NaN explicitly, for code readability.

skirpichev: Ok, a quick check for nan was added to the hyper(), everything else was reverted.

skirpichev: @dimpase, does it make sense for you now?

dimpase: it's still puzzling how it's possible to get to the NaN check without getting an overflow (and thus an uncaught error). I.e., why is this check  needed, at all?

Also, you can get an `inf` rather than NaN, and `inf` is perfectly possible to meaningfully compare with non-infs.

skirpichev: > it's still puzzling how it's possible to get to the NaN check without getting an overflow (and thus an uncaught error). I.e., why is this check needed, at all?

The idea was to use double negation (`<` -> `not >=`) to make a quick exit for nan's.

I agree, this is much more clear in form like `while abs(t) >= tol: ...`

> Also, you can get an inf rather than NaN, and inf is perfectly possible to meaningfully compare with non-infs.

I think this piece of code can't handle inf's correctly, so nothing was broken.

Anyway, I've reverted this stuff: an explicit check for nan's in hyper() does the job.  The second commit is optional now, but I think it's a good idea to have same output for fp.mag() and mp.mag().   Right now we have 0 and nan for nan's, respectively.

skirpichev:
Ok, given the changes are trivial now, I'm going to merge this tomorrow.

Issue #532, opened 2020-03-24
Metaclassing causes different classes for custom contexts


sylee957:
```python3
import mpmath
from mpmath import MPContext

mp = mpmath.mp
mp0 = MPContext()

print(isinstance(mp.matrix([[1, 2], [3, 4]]), mp.matrix))
print(isinstance(mp.matrix([[1, 2], [3, 4]]), mp0.matrix))
print(isinstance(mp0.matrix([[1, 2], [3, 4]]), mp.matrix))
print(isinstance(mp0.matrix([[1, 2], [3, 4]]), mp0.matrix))

print(isinstance(mp.mpf(1), mp.mpf))
print(isinstance(mp.mpf(1), mp0.mpf))
print(isinstance(mp0.mpf(1), mp.mpf))
print(isinstance(mp0.mpf(1), mp0.mpf))

print(isinstance(mp.mpq(1), mp.mpq))
print(isinstance(mp.mpq(1), mp0.mpq))
print(isinstance(mp0.mpq(1), mp.mpq))
print(isinstance(mp0.mpq(1), mp0.mpq))
```

I see mp contexts are creating different classes on fly, but only `mpq` is hardcoded to a single class.
How should I check if a class is generally an instance of mpf or matrix across every contexts? Were those metaclsses necessary for mpf and matrix?

oscarbenjamin:
> you could check as `isinstance(obj, ctx.class)` (where `class` e.g. `mpf`).

This only works if you know which context the mpf should be from. The question is how to test if an object is an mpf presumably in a situation like an `__add__` method where you need to check what kind of object you have:
```python
def __add__(self, other):
    if isinstance(other, ...):
        return ...
```
All mpfs are subclasses of the internal `_mpf` class but that appears not be public.

I think it would be better not to use dynamically created `mpf` classes for each context (gh-696).

skirpichev:
> This only works if you know which context the mpf should be from.

Or you could *find* those context from some candidates.

> All mpfs are subclasses of the internal _mpf class but that appears not be public.

All numeric types (but not mpq) of the mpmath have the context attribute...  So, in your setup the only option is, probably, use duck typing, e.g.: `if hasattr(obj, 'context') and isinstance(obj, obj.context.mpf): ...`

> I think it would be better not to use dynamically created mpf classes for each context

Yes, I think so.

Let me know if you do think I should reopen this issue.

oscarbenjamin:
[...]

PR #691, opened 2023-04-25
Drop private mpq class, use Rational's, provided by backend

skirpichev:
* Missing feature (for pure-python) is an additional caching mechanism.  Not sure it worth.
* Unfortunately, gmpy2 has no released version with support for mpq.as_numer_denom(), hence I use foo.numerator/denominator to unpack components of rationals...  Edit: the numbers.Rational doesn't require this method, so we could not rely on it.


skirpichev:
Missing feature (for pure-python) is additional caching mechanism.

oscarbenjamin:
This looks good to me.

I don't know how significant the caching mechanism is but it does not look as if there are any places where `MPQ` is used in a tight loop.

skirpichev:
@oscarbenjamin, I've some doubts about the ctx.mpq attribute.  It's not documented, but apparently people discover it and use (e.g. #532).  Maybe we should keep this and issue a deprecation warning before removing?

oscarbenjamin:
> Maybe we should keep this and issue a deprecation warning before removing?

Oh, I didn't notice that this removes the `ctx.mpq` attribute as well as `ctx._mpq`. I would say that as long as `MPQ` exists then it is harmless to just leave `ctx.mpq` as an alias to it. If there is no significant gain in removing it then I don't see any need to deprecated it.

skirpichev:
> to just leave ctx.mpq as an alias to it.

mpq is a special attribute for the MPContext() only.

> If there is no significant gain in removing

Slightly better API?

PR #694, opened 2023-05-08
Use setuptools_scm to update __version__

skirpichev:


skirpichev:
@asmeurer, will you object this change, given the previous history (i.e. #443, #580)?
Keep in mind, that now we don't support py2 and setuptools_scm don't add additional dependencies (it's only in the build-system requirements, just like the versioner).

asmeurer:
This doesn't appear to do the right thing if I import the local copy of mpmath. It grabs the version from the installed version rather than the version that is actually imported. 

Basically, this seems to be assuming that some metadata is in sync with the source, whereas with the version being listed directly like this it's impossible for this issue to occur.

PR #695, opened 2023-05-08
Use `math.isqrt` in `isqrt_python` calculations

haru-44:
The `isqrt_python` calculation is faster using `math.isqrt` when x is small.

For speed comparisons, see the following PR

sympy/sympy#25111

### DIFF
--- a/mpmath/libmp/libintmath.py
+++ b/mpmath/libmp/libintmath.py
@@ -7,6 +7,7 @@ here from settings.py
 """
 
 import math
+import sys
 from bisect import bisect
 
 from .backend import BACKEND, gmpy, sage, sage_utils, MPZ, MPZ_ONE, MPZ_ZERO
@@ -297,12 +298,6 @@ def sqrtrem_python(x):
 
 def isqrt_python(x):
     """Integer square root with correct (floor) rounding."""
-    # Due to an implementation problem with `math.isqrt`,
-    # the speed is reversed around about 2**55000.
-    # Python 3.12 will resolve this, so we assume that in
-    # the future we will just call `math.isqrt`.
-    if x.bit_length() < 55000:
-        return math.isqrt(x)
     return sqrtrem_python(x)[0]
 
 def sqrt_fixed(x, prec):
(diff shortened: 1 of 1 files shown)


mpmath/libmp/libintmath.py
```diff
@@ -297,6 +297,12 @@ def sqrtrem_python(x):
 
 def isqrt_python(x):
     """Integer square root with correct (floor) rounding."""
+    # Due to an implementation problem with `math.isqrt`,
+    # the speed is reversed around about 2**55000.
+    # Python 3.12 will resolve this, so we assume that in
+    # the future we will just call `math.isqrt`.
+    if x.bit_length() < 55000:
+        return math.isqrt(x)
     return sqrtrem_python(x)[0]
```
oscarbenjamin, line 306: Maybe this whole thing could be wrapped in a Python version check:
```python
if sys.version_info >= (3, 12):
   def isqrt_python(...)
else:
   def isqrt_python(...)
```
I wonder what pypy does to provide isqrt.

skirpichev: > wrapped in a Python version check

Yep.  I would also prefer this.

> I wonder what pypy does to provide isqrt.

https://foss.heptapod.net/pypy/pypy/-/blob/branch/py3.10/pypy/module/math/app_math.py#L133

skirpichev (changes requested):
So, please just use math.isqrt for isqrt, isqrt_fast, etc for python versions >= 3.12, if the BACKEND is not 'gmpy' or 'sage'.

mpmath/libmp/libintmath.py
skirpichev:
```python
else:
    if sys.version_info >= (3, 12):
        isqrt_small = isqrt_fast = isqrt = math.isqrt
    else:
        isqrt_small = isqrt_small_python
        isqrt_fast = isqrt_fast_python
        isqrt = isqrt_python
```
Yes.

haru-44: Fixed.

skirpichev (approved):

PR #697, opened 2023-05-08
Use gcd() from the bigint' backend

skirpichev:
Also drop some workarounds for ancient gmpy's.

PS: Probably same as for the gcd() we could do for the ifac().  But this function has a different caching mechanism than e.g. the math.factorial, so it may worth to do some benchmarks fisrt.

### DIFF
--- a/mpmath/libmp/libintmath.py
+++ b/mpmath/libmp/libintmath.py
@@ -9,6 +9,7 @@ here from settings.py
 import math
 import sys
 from bisect import bisect
+from functools import lru_cache
 
 from .backend import BACKEND, gmpy, sage, sage_utils, MPZ, MPZ_ONE, MPZ_ZERO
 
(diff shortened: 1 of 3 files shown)


mpmath/libmp/libintmath.py
```diff
@@ -305,21 +299,20 @@ def sqrt_fixed(x, prec):
...
     isqrt_small = isqrt_fast = isqrt = \
         getattr(sage_utils, "isqrt", lambda n: MPZ(n).isqrt())
     sqrtrem = lambda n: MPZ(n).sqrtrem()
+    gcd = sage.gcd
 else:
     isqrt_small = isqrt_small_python
     isqrt_fast = isqrt_fast_python
     isqrt = isqrt_python
     sqrtrem = sqrtrem_python
+    gcd = math.gcd
```
oscarbenjamin, line 315: In Python 3.8 this function can only accept two arguments unlike the `gcd` function that this replaces. Is it only needed to have a two argument version?

skirpichev: > Is it only needed to have a two argument version?

Tests pass:)  BTW, the libintmath is an internal module.

skirpichev: > Looks like it is only used once in the entire codebase

Indeed.  And some gcd imports are unused.

> It is exported publicly though from libmp ... It is also attached to contexts as _gcd

1. Probably we could safely consider underscore-prefixed context method as private (PEP 8).
2. The gcd() is undocumented (no docstring and sphinx docs), it will not be imported with ``from mpmath import *``.  On those conditions, I think we could count this function as private.

(BTW, I'm not sure that attaching the private method to the context is better than just a direct import from the helper library...)

The global mpmath namespace already too big:
```
>>> len(dir(mpmath))
398
```
So, probably it's a good idea to open an issue about the public API of the mpmath.  In a first approximation I would consider that we could break any "public" (per PEP 8) interface, that wasn't documented.

skirpichev: I've added vararg support for gcd on python versions < 3.9

oscarbenjamin:
Looks good to me.

Issue #696, opened 2023-05-08
Simplify the codebase by avoiding metaprogramming


oscarbenjamin:
Following on from https://github.com/mpmath/mpmath/issues/683#issuecomment-1537412261 concerning the speed of creating a context object and see the past discussions in gh-178.

The mpmath codebase uses a number of metaprogramming features that increase complexity for what I suspect is relatively little benefit (mostly DRY and micro-optimisation). These make the codebase hard to follow and also defeat static code analysis which is increasingly integrated in modern Python editors. I propose to remove these and move towards ordinary Python code with plain functions, methods and attributes.

Specifically I suggest:

- Eliminate dynamic class generation. Make e.g. mpf a simple type that just has `_mpf_` and `ctx` attributes.
- Write methods in ordinary Python code rather than generating them dynamically either with exec or by attaching them from decorators.
- Use more traditional methods for reducing boilerplate like reusable functions.

One example of this is dynamically created classes like:
https://github.com/mpmath/mpmath/blob/29b47ce9af262287c7dd00f537eded49a74e3885/mpmath/ctx_mp_python.py#L584-L596
There are also functions and methods created from code strings at runtime:
https://github.com/mpmath/mpmath/blob/29b47ce9af262287c7dd00f537eded49a74e3885/mpmath/ctx_mp_python.py#L280-L293
We also have methods that are dynamically added to classes:
https://github.com/mpmath/mpmath/blob/29b47ce9af262287c7dd00f537eded49a74e3885/mpmath/functions/functions.py#L16-L20

Some of these happen at import time and others happen when a context object is created and add measurable overhead to the cost of creating a context:
```python
In [9]: from mpmath import MPContext, mp

In [10]: %timeit MPContext()
538 µs ± 16.6 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)

In [11]: %timeit mp.clone()
561 µs ± 27.3 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)

In [12]: %timeit mp.cos(1)
10.9 µs ± 253 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)

In [18]: %prun -s cumulative [mp.clone() for _ in range(1000)]
         863004 function calls in 1.736 seconds

   Ordered by: cumulative time

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
        1    0.000    0.000    1.736    1.736 {built-in method builtins.exec}
[...]

PR #702, opened 2023-05-19
Avoid dynamic method creation in _mpf

skirpichev:
a first step to solve #696


oscarbenjamin:
I think this looks good.

Perhaps it could be for a separate PR but it looks like these methods could be simplified more along the lines of the `_mpc` methods:
https://github.com/mpmath/mpmath/blob/8c4256c0d2202863d8e5ab070601b73f4e797023/mpmath/ctx_mp_python.py#L615-L622

Probably the `_mpf` methods should delegate the complex case to `_mpc` rather than having inline logic to handle it. Then you could have something like:
```python
# class _mpf
def __eq__(self, other):
    # handle the common case up front
    if isinstance(other, _mpf):
        return mpf_eq(self._mpf_, other._mpf_)
    self_new, other_new = ctx.convert_pair(self, other)
    if self_new is NotImplemented:
        return NotImplemented
    # might both be _mpc so we call __eq__ rather than mpf_eq
    return self_new.__eq__(other_new)
```

skirpichev:
> Probably the `_mpf` methods should delegate the complex case to `_mpc`

I think it's a good idea, but I doubt it should be implemented there.

We also could, for example, just return NotImplemented for anything except `mpf`/int/floats in `mpf` methods.  Such changes, however, must be carefully benchmarked.

PR #414, opened 2018-10-18
Binary/octal/hexadecimal string support

warnerjon12:
Response to issue #345.

Closes #345

Issue #337, opened 2016-11-25
Add mpf.__format__() dunder methods


ThomasA:
I am struggling with the conversion of a script written for Python 2.7 to Python 3 (PS: I know Python 3 still understands the old formatting syntax, but I would like to convert to the new formatting syntax as well). Consider the following example (IPython with output):

    In [1]: import mpmath
    
    In [2]: value = mpmath.mp.mpf('1.0')
    
    In [3]: print('%10.0f' % value)
             1
    
    In [4]: print('{:10.0f}'.format(value))
    ---------------------------------------------------------------------------
    TypeError                                 Traceback (most recent call last)
    <ipython-input-4-f8a6817daeeb> in <module>()
    ----> 1 print('{:10.0f}'.format(value))

Is there a sensible fix for this?

I have tried:

     print('{:10s}'.format(mpfloat.nstr(value)))

instead, but I cannot seem to figure out how to configure `nstr` to behave like line 3 above. The latter seems to perform rounding of some numbers that the former does not...

skirpichev:
> I know Python 3 still understands the old formatting syntax

Unfortunately, mpmath types can't be formatted with the old formatting syntax.  Your above examples works because implicit conversion to float.  But CPython floats have only 53 bits of precision, so:
```pycon
>>> float(mpf(2)**2000)
inf
```

> Is there a sensible fix for this?

Yes, provide support for the new formatting syntax.  Meanwhile you can convert mpf objects (using as_integer_ratio() method) either to the CPython's Fraction type or to the gmpy2's mpfr type.  Both types support new formatting.  The mpfr type additionally supports different rounding modes:
```pycon
>>> "{0:15.3Uf}".format(mpfr("1.23456"))
'          1.235'
>>> "{0:15.3Df}".format(mpfr("1.23456"))
'          1.234'
```

Or use ``nstr()``.

> I cannot seem to figure out how to configure nstr to behave like line 3 above

You can't drop ``.0`` in the ``nstr()`` output.  Use ``nint()`` to get nearest integer.

skirpichev:
Ok, now almost all is finished.

Work in progress:
* https://github.com/mpmath/mpmath/pull/841
* https://github.com/mpmath/mpmath/issues/837

"new style" formatting support also lacks documentation.

skirpichev:
@javierelpianista, I think this can be closed by the documentation pr.

PR #711, opened 2023-06-28
Support binary/hexadecimal string output

### LINKED ISSUE #345: hex representation
It would be nice to be able to convert mpmath floats to and from a string hex representation (like `float.hex` and `float.fromhex`).

skirpichev:
Adapted from #414.

Closes #345

Co-authored-by: Sergey B Kirpichev <skirpichev@gmail.com>

* this version doesn't touch argument names (dps->ps)
* ~~no support for float.hex()-style binary exponents~~

PS: `from_str()` part was implemented in #703.  @asmeurer, I would appreciate your review, as you've opened #345.

### DIFF
--- a/mpmath/ctx_mp_python.py
+++ b/mpmath/ctx_mp_python.py
@@ -37,11 +37,13 @@ class _mpf(mpnumeric):
         or a decimal string representing a number in floating-point
         format."""
         prec, rounding = cls.context._prec_rounding
+        base = 0
         if kwargs:
             prec = kwargs.get('prec', prec)
             if 'dps' in kwargs:
                 prec = dps_to_prec(kwargs['dps'])
             rounding = kwargs.get('rounding', rounding)
+            base = kwargs.get('base', base)
         v = new(cls)
         if type(val) is cls:
             val = val._mpf_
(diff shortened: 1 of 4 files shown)


skirpichev:
This is ready for review.  I would like to merge this in few days to use that for ``mpf.__format__()``.

oscarbenjamin:
How would you use this functionality with mpf rather than `_mpf_`?

```
>>> import mpmath
>>> from mpmath.libmp import to_str, from_str
>>> f = mpmath.sqrt(2)
>>> to_str(f, 10, base=2)
  File "/home/oscar/current/active/mpmath/mpmath/libmp/libmpf.py", line 1128, in to_str
    if not s[1]:
           ~^^^
TypeError: 'mpf' object is not subscriptable
>>> to_str(f._mpf_, 10, base=2)
'0b1.0110101'
```
It looks like the `mpf` constructor accepts it:
```
>>> mpmath.mpf('0b1.0110101')
mpf('1.4140625')
```

skirpichev:
> How would you use this functionality with mpf rather than _mpf_?

First, we could provide ``__format__()`` dunders to support "new-style" formatting of mpf/mpc's (issue mpmath/mpmath#337).  See e.g. gmpy2: https://gmpy2.readthedocs.io/en/latest/mpfr.html#gmpy2.mpfr.__format__  I'm planning to do this like https://github.com/python/cpython/pull/113805 (i.e. "x" will be used to format numbers in the hexadecimal notation).  Another option (as we don't support old-style string formatting) may be using "a" letter for this (in printf-style, like the MPFR).  Support for optional rounding mode (gmpy2-like) also does make sense for me (see #757).  Maybe we should also support formatting binary floats ("b"?).

> It looks like the mpf constructor accepts it

Yes, strings with prefixes should be accepted automatically.  Most important case is hexadecimal floats, of course - this will allow us to pass exact floating point values to the mpf constructor (see also CPython pr: https://github.com/python/cpython/pull/114668).

But to handle arbitrary bases we should add ``base`` keyword argument (default=0) to mpf/mpc constructors (as gmpy2 does).

PR #727, opened 2023-10-10
CI: Add integration testing with Sage

### LINKED ISSUE #705: Test sage integration or remove sage backend
We have a special value 'sage' for the BACKEND global variable.  If this is a part of the mpmath codebase - this should be tested with our regular CI tests.

mkoeppe:
Resolves #705

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,5 +1,5 @@
 name: test
-on: [push, pull_request]
+on: [push, pull_request, workflow_dispatch]
 jobs:
   build:
     runs-on: ubuntu-20.04
@@ -48,7 +48,7 @@ jobs:
         coverage html
         coverage xml
     - name: Upload coverage data
-      if: matrix.python-version == 3.10 || matrix.python-version == 3.11
+      if: "(matrix.python-version == 3.10 || matrix.python-version == 3.11) && github.repository == 'mpmath/mpmath'"
       uses: codecov/codecov-action@v3
       with:
         files: ./coverage.xml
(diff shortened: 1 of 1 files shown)


oscarbenjamin:
> Hmm, ~1h 57m for an incomplete test... I doubt it's a good idea to run this on every pr. Perhaps, for tag push?

The SymPy CI runs this after a merge to master rather than on each push to a PR.

> Let me know if you think that we should do proper deprecation of this private module.

I would say that since the module is being used in some downstream codebases let's not remove it yet in the next mpmath release. I don't know what exactly was the history of all this or why it was being used but at least SymPy and Sage were using it so I would expect that something else is using it as well. It would be better to remove all known downstream usage first before removing or deprecating the module in mpmath.

mkoeppe:
> IIUIC, this integration will not run the mpmath tests suite (only own sage tests),

This can be added from the Sage side by creating a test script build/pkgs/mpmath/spkg-check.in; PR welcome

skirpichev:
> The SymPy CI runs this after a merge to master rather than on each push to a PR.

Yeah I saw this.  But I was hoping there is a faster way...

> I would say that since the module is being used in some downstream codebases let's not remove it yet in the next mpmath release.

I think we can restore some minimal version of it (#728) and then - do proper deprecation.

> This can be added from the Sage side by creating a test script build/pkgs/mpmath/spkg-check.in

Can this be patched right in the workflow, just as with ``update-pkgs.sh``?

mkoeppe:
> > The SymPy CI runs this after a merge to master rather than on each push to a PR.
> 
> Yeah I saw this. But I was hoping there is a faster way...

Of course only a subset of the ~ 3000 modules of Sage may be enough to test. But we don't easily know which ones are relevant for proper coverage of mpmath. 

> > This can be added from the Sage side by creating a test script build/pkgs/mpmath/spkg-check.in
> 
> Can this be patched right in the workflow, just as with `update-pkgs.sh`?

Sure.

skirpichev:
> But we don't easily know which ones are relevant for proper coverage of mpmath.

None.  Mpmath has no non-optional dependencies.

skirpichev:
see discussion in #705

Issue #753, opened 2024-02-14
Better integration with numpy


sunqm:
The functionalities of mpmath matrix class has overlap with the numpy ndarray class. But matrix class and numpy array class are not compatible. Currently, linalg functions such as `qr`, `lu`, `norm` can only work with mpmath matrix. It seems not difficult to change the code base to make these functions support numpy arrays.

It would be convenient if mpmath matrix and the relevant functions are compatible with numpy arrays. This allows users to use numpy array to manage mpmath.mpf objects, as numpy array has so many functionalities to manipulate array and matrix objects that mpmath matrix does not have.

Will the integration with numpy be considered in the future?

skirpichev:
see also https://github.com/mpmath/mpmath/issues/217, https://github.com/mpmath/mpmath/issues/618

sunqm:
Technically, it's not that difficult to make mpmath aligned better with numpy, IMO. The question is the goal and the perspective of the package: whether numpy compatibility will be considered in the future releases. If so, I'm happy to make contributions. At beginning, it is not necessary to integrate every feature of numpy. Upgrading the existing functionalities does not look like too much work. It just needs to make the matrix class a subclass of numpy array, and refactor the relevant linalg and calculas functions following the numpy array APIs.

oscarbenjamin:
mpmath does not depend on numpy so it is not possible to subclass numpy's ndarray.

Adding an array API compatible module as an interface within mpmath could be reasonable though.

sunqm:
It may not be necessary to sublcass numpy array or install numpy. At some point, this NEP https://numpy.org/neps/nep-0041-improved-dtype-support.html#fixed-high-precision-math will be available in numpy. Then the numpy array can provide the functionalities of mp.matrix .

In mpmath, functions in the linalg module just need to follow the numpy API conventions. This will allow mpmath linalg functions supporting numpy arrays automatically. For example, one of the simplest change would be to replace the `rows` and `cols` in mp.matrix with an attribute `shape`. functions like eig, schur will automatically support numpy arrays.

skirpichev:
> In mpmath, functions in the linalg module just need to follow the numpy API conventions.

That does make sense.  Just keep in mind, that linalg functions and the matrix class are part of the mpmath's public API.  We should support old attributes for some time, add deprecation warnings, etc.

Issue #217, opened 2010-01-18
use numpy for fp matrix arithmetic


vks:
```
Were in the code should this be implemented? ctx_fp.py? The probably some code has to be 
moved.
```

Original issue for #217: http://code.google.com/p/mpmath/issues/detail?id=177
Original author: https://code.google.com/u/Vinzent.Steinberg@gmail.com/
Original owner: https://code.google.com/u/Vinzent.Steinberg@gmail.com/

skirpichev:
Lets close this in favor of #753

PR #754, opened 2024-02-16
Update mp.matrix and linalg functions following np.ndarray APIs

sunqm:
* Code refactoring:
Following the discussion in #753, the mp.matrix class and relevant linalg and arithmetic functions are updated, following the array API conventions in numpy. mp.matrix can be converted to numpy array using the `.to_numpy()` method, e.g.
```
mp.eye(3).to_numpy()
```
The mp-dtype numpy arrays can be passed as arguments to the following functions
```
expm
cosm
sinm
sqrtm
logm
powm
hessenberg
schur
eig
eigh
svd
LU_decomp
L_solve
U_solve
lu_solve
lu
inverse
residual
qr_solve
cholesky
cholesky_solve
det
cond
qr
```
When these functions are called with numpy arrays, their return objects are also encoded in numpy array.

* Bugfix:
This PR also fixes a bug in `cholesky_solve` for complex matrix.


skirpichev (dismissed):
1. We can't just drop public attributes for public classes.  These should be deprecated first.
2. Could you please factor out the cholesky_solve() bugfix to a separate pr?
3. There are test failures (as well as problems with scipy installation).  Can't you use [numpy.linalg](https://numpy.org/doc/stable/reference/routines.linalg.html#module-numpy.linalg) for tests?

sunqm:
> 1. We can't just drop public attributes for public classes.  These should be deprecated first.
> 2. Could you please factor out the cholesky_solve() bugfix to a separate pr?
> 3. There are test failures (as well as problems with scipy installation).  Can't you use [numpy.linalg](https://numpy.org/doc/stable/reference/routines.linalg.html#module-numpy.linalg) for tests?

cholesky_solve bugfix is issued in #755

skirpichev (commented):
Sorry, I don't think I can approve this and I don't see how this could be improved.  Count me -1.

This pr mix code refactoring (using the shape attribute where rows/cols were used) with actual changes.  It makes the returned type either matrix or ndarray.  And it seems that we can't, in general, use single code path for matrix and ndarray data internally.

Lets keep this for a while, maybe someone else could review.  @oscarbenjamin?

The ``to_numpy()`` helper could be added in a separate pr.  Maybe the shape attribute too.

skirpichev:
> Is the intention that it should be possible to use mpmath functions like det with a numpy array as input?

That was my guess as well.

> It looks as if the code here does not convert the numpy arrays to mpmath

Actually, no code need to convert numpy arrays to mpmath.

Using numpy arrays internally is another thing.  That does make sense for me in the way it was outlined by Fredrik: https://github.com/mpmath/mpmath/issues/217#issuecomment-41250152

oscarbenjamin:
If someone is using mpmath's det rather than numpy's det then they presumably want the benefit of mpmath's multiprecision support. That can only work if the arithmetic uses mpmath types though.

Otherwise if mpmath's det function uses arithmetic with the numpy scalar types then I don't see why anyone would want to use it instead of just using numpy's det function.

Or is the intention that someone would have a numpy array with dtype=object and mpmath mpfs as the entries in the array?

skirpichev:
> that someone would have a numpy array with dtype=object and mpmath mpfs as the entries in the array?

Only this case does make sense for me.  But elements of the input numpy array should be converted to appropriate mpf's (i.e., array should be recreated).

sunqm:
> > Is the intention that it should be possible to use mpmath functions like det with a numpy array as input?
> 
> That was my guess as well.
> 
> > It looks as if the code here does not convert the numpy arrays to mpmath
> 
> Actually, no code need to convert numpy arrays to mpmath.
> 
> Using numpy arrays internally is another thing. That does make sense for me in the way it was outlined by Fredrik: [#217 (comment)](https://github.com/mpmath/mpmath/issues/217#issuecomment-41250152)

The intention is to use numpy array with the mpf and mpc object . As the reason I stated in issue #753 , it is a useful feature to allow numpy array to manage a group of mpf objects. This allows us to use numpy ufuncs and other features for high precision math. However, functions in mpmath linear algebra modules are not compatible with numpy array. This make the use of numpy-mpf  objects inconvenient. This PR updates the structure of mpmath.matrix class and the implementation of relevant linalg functions. This solves the compatibility issue.

It does not use numpy internally. Just to ensure the relevant functions do not fail when processing numpy arrays.

sunqm:
> If someone is using mpmath's det rather than numpy's det then they presumably want the benefit of mpmath's multiprecision support. That can only work if the arithmetic uses mpmath types though.
> 
> Otherwise if mpmath's det function uses arithmetic with the numpy scalar types then I don't see why anyone would want to use it instead of just using numpy's det function.
> 
> Or is the intention that someone would have a numpy array with dtype=object and mpmath mpfs as the entries in the array?

Exactly. The intention is to use numpy array with `dtype=object` .

skirpichev:
> it is a useful feature to allow numpy array to manage a group of mpf objects

But this is - a numpy feature, not mpmath.

> However, functions in mpmath linear algebra modules are not compatible with numpy array.

Could you provide some concrete example?  Numpy arrays should be accepted by the matrix constructor.  Maybe some mpmath functions just don't check argument types.

sunqm:
Numpy does support to use the mpf object as its dtype. The issue is that the mpmath.matrix class uses a different API convention than numpu array. This causes the linalg functions failed to process numpy arrays. Here is a simple example
```
a = np.array([[mp.mpf(1), mp.mpf(0)], [mp.mpf(0), mp.mpf(1)]])
mp.eigh(a)
```

skirpichev:
> The issue is that the mpmath.matrix class uses a different API convention than numpu array.

Really?  I don't think so.

> ``mp.eigh(a)``

Great!  This concrete example shows where is the problem:
https://github.com/mpmath/mpmath/blob/ea3ec6b6b4c36d684b44ae765bf3f36263e0d8ed/mpmath/matrices/eigen_symmetric.py#L556-L560
Apparently, the code does assume that the input is a matrix instance.  We could just do automatic conversion if the input type is wrong.  I don't understand why we should instead introduce the shape attribute, etc

sunqm:
I don't understand why this library is so resistant to numpy. Numpy is almost the de facto standard for matrix and array objects.  Leveraging the structure of numpy arrays, or a similar array structure, such as the attribute `shape`, `size` etc can offer lots of convenience to users. And it does not compromise any functionality of this library.

oscarbenjamin:
The question is how exactly numpy and mpmath should interoperate. There are many different ways that that might be expected to work. This PR supposes one possibility but it is not clear why that is better than others.

sunqm:
> I don't see a big issue if you just add these attributes to the matrix type. The question is why do we need all these changes like rows -> shape[0] across the whole codebase.

There is no attribute `rows` in numpy array. If rows,cols are replaced by `shape`, the linalg functions can handle matrix and numpy array the same way. There is no need to convert array to matrix and backward.

> 
> As you could see, hardly this does make sense in general (maybe for the fp context). To get accurate result, you should manually increase precision for the whole computation (temporary increase of ctx.prec in the expm() will not affect np.matrix'es).
> 

Creating np.array with `mpf` type, although displayed as `dtype=object`, is different to initializing numpy array with `dtype=object`. If you create a `np.matrix([[1, 2], [3, 4]], dtype=object)` , the elements are treated as `object`, than numeric numbers. This is the logic of numpy dtype. You can convert to the `mpf` type implicitly. But this is not a good design to practice type. You can check how numpy treats the `object` type for numeric operations.

> Your solution is something in between. It makes an illusion, that functions handle np.array's transparently, but it's not the case. Sometimes you do conversion to ctx.matrix and vice versa, sometimes not. Why do you think this code complication does make sense for the mpmath?

This is a compromise since using numpy array in mpmath is rejected. Numpy array is designed better than mpmath.matrix in many aspects. The ideal solution is to deprecate mp.matrix and use numpy array as the matrix container. If the library can accept this, code can be much simpler.

skirpichev:
> Creating np.array with mpf type, although displayed as dtype=object, is different to initializing numpy array with dtype=object.

You are right, my example should be corrected here.  But the conclusion holds, computation will lose precision:
```pycon
>>> mp.expm(mp.matrix([[1, 2], [3, 4]]))
matrix(
[['51.968956198705', '74.7365645670032'],
 ['112.104846850505', '164.07380304921']])
>>> mp.expm(np.matrix(mp.matrix([[1, 2], [3, 4]])).reshape(2,2))
matrix([[mpf('51.677495573354904'), mpf('74.311779826707408')],
        [mpf('111.46766974006113'), mpf('163.14516531341602')]],
       dtype=object)
>>> with mp.extraprec(1000):
...     m = mp.expm(np.matrix(mp.matrix([[1, 2], [3, 4]])).reshape(2,2))
... 
>>> m
matrix([[mpf('51.968956193868668'), mpf('74.736564559954574')],
        [mpf('112.10484683993186'), mpf('164.07380303380053')]],
       dtype=object)
```

skirpichev (changes requested):
It seems, this patch not covered by tests.  See failed CI jobs.

sunqm:
The missing lines in coverage are all related to numpy. Unless allowing installing numpy in tests, these code cannot be tested.

skirpichev:
> Unless allowing installing numpy in tests

It's allowed.  Optional dependencies for CI include numpy.

skirpichev:
Feel free to reopen this, if you are ready to continue this work.

PR #756, opened 2024-02-19
Fix null matrix determinant and string representation

### LINKED ISSUE #745: Null matrices don't properly convert to repr string and attempting to compute the determinant of 0×0 matrix raised an exception.
```python
>>> from mpmath import mp, matrix
>>> mp.pretty = False
>>> null_matrix = matrix(0)
>>> null_matrix
matrix(
])
>>> mp.pretty = True
>>> null_matrix

>>> null_matrix = matrix(0, 2)
>>> null_matrix

>>> mp.pretty = False
>>> null_matrix
matrix(
])
>>> null_matrix = matrix(2, 0)
>>> null_matrix
matrix(
],
]])
>>> mp.pretty = True
>>> null_matrix
[]
[]
>>> mp.det(matrix(0, 0))
  File "C:\Python\Python310\lib\site-packages\mpmath\matrices\linalg.py", line 545, in det
    R, p = ctx.LU_decomp(A)
  File "C:\Python\Python310\lib\site-packages\mpmath\matrices\linalg.py", line 125, in LU_decomp
    tol = ctx.absmin(ctx.mnorm(A,1) * ctx.eps) # each pivot element has to be bigger
  File "C:\Python\Python310\lib\site-packages\mpmath\matrices\matrices.py", line 997, in mnorm
    return max(ctx.fsum((A[i,j] for i in xrange(m)), absolute=1) for j in xrange(n))
ValueError: max() arg is an empty sequence
```
The 0 × _n_ matrix should be converted to repr string as ```matrix(0, n)```.
The _n_ × 0 matrix should be converted to repr string as ```matrix(n, 0)```.
The determinant of the 0×0 matrix [should be 1](https://en.wikipedia.org/wiki/Matrix_(mathematics)#Empty_matrix).

mart-mihkel:
Fixes #745 

* Fix null matrix repr string
* Make determinant of null matrix return 1
* Update docstring of `det`
* Add null matrix det test case to `test_linalg.py`


mpmath/matrices/linalg.py
```diff
@@ -592,6 +598,10 @@ def det(ctx, A):
         try:
             # do not overwrite A
             A = ctx.matrix(A).copy()
+
+            if A.rows == 0 or A.cols == 0:
+                return 1
```
skirpichev, line 603: Can't we fix instead the ``LU_decomp()``, without a special case here?  E.g. for ``p = []`` the code below will return ``1``.

To fix exception in the ``min()`` builtin you could provide appropriate default.

mpmath/matrices/matrices.py
```diff
@@ -394,9 +394,11 @@ def tolist(self):
...
         if self.ctx.pretty:
             return self.__str__()
-        s = 'matrix(\n'
-        s += self._toliststr(avoid_type=True) + ')'
-        return s
+
+        if self.__rows == 0 or self.__cols == 0:
+            return f'matrix({self.__rows}, {self.__cols})'
+
+        return f'matrix(\n{self._toliststr(avoid_type=True)})'
```
skirpichev, line 401: Why not fix instead the ``_toliststr()`` helper and the matrix constructor?

```pycon
>>> mp.pretty = True
>>> np.array([[]])
array([], shape=(1, 0), dtype=float64)
>>> matrix([[]])  # this probably should be [[]]?
[]
>>> np.array([])
array([], dtype=float64)
>>> _.shape
(0,)
>>> matrix([])  # this doesn't, but it should 
[...]
IndexError: list index out of range
```

And you should add tests.

mpmath/matrices/linalg.py
mart-mihkel: Yes, that would probably make for a better solution.
Adding a default value to the builtin in `mnorm()` seems to have fixed it, but still had to add an extra condition to an if caluse in `LU_decom()`.

mpmath/matrices/linalg.py
```diff
@@ -146,7 +146,7 @@ def LU_decomp(ctx, A, overwrite=False, use_cache=True):
                 A[i,j] /= A[j,j]
                 for k in range(j + 1, n):
                     A[i,k] -= A[i,j]*A[j,k]
-        if ctx.absmin(A[n - 1,n - 1]) <= tol:
+        if len(p) > 0 and ctx.absmin(A[n - 1,n - 1]) <= tol:
```
skirpichev, line 149:
```suggestion
        if p and ctx.absmin(A[n - 1,n - 1]) <= tol:
```

mpmath/matrices/matrices.py
```diff
@@ -988,8 +990,8 @@ def mnorm(ctx, A, p=1):
             p = ctx.convert(p)
         m, n = A.rows, A.cols
         if p == 1:
-            return max(ctx.fsum((A[i,j] for i in range(m)), absolute=1) for j in range(n))
+            return max((ctx.fsum((A[i,j] for i in range(m)), absolute=1) for j in range(n)), default=0)
```
skirpichev, line 993: Why 1?  I would expect 0 here.

mpmath/matrices/matrices.py
```diff
@@ -988,8 +990,8 @@ def mnorm(ctx, A, p=1):
...
         elif p == ctx.inf:
-            return max(ctx.fsum((A[i,j] for j in range(n)), absolute=1) for i in range(m))
+            return max((ctx.fsum((A[i,j] for j in range(n)), absolute=1) for i in range(m)), default=0)
```
skirpichev, line 995: Ditto here.

mpmath/matrices/matrices.py
mart-mihkel: Was just playing with values and forgot to change it back 👍

mpmath/matrices/matrices.py
mart-mihkel: Adjusted the `_toliststr()` helper, to work for the given cases.
But still left  `__nstr__()` which does the pretty printing untouched.
Should it be adjusted aswell?

mpmath/matrices/matrices.py
```diff
@@ -362,28 +363,21 @@ def __nstr__(self, n=None, **kwargs):
     def __str__(self):
         return self.__nstr__()
 
-    def _toliststr(self, avoid_type=False):
```
mart-mihkel, line 365: Removed the extra parameter here because this private method is invoked only from `__repr__()`  with the value set to `True

mpmath/matrices/matrices.py
```diff
@@ -362,28 +363,21 @@ def __nstr__(self, n=None, **kwargs):
...
+    def _toliststr(self):
         """
         Create a list string from a matrix.
-
-        If avoid_type: avoid multiple 'mpf's.
         """
-        # XXX: should be something like self.ctx._types
         typ = self.ctx.mpf
-        s = '['
+        rows = []
```
skirpichev, line 371:
```suggestion
        s = '['
```

mpmath/matrices/matrices.py
```diff
@@ -362,28 +363,21 @@ def __nstr__(self, n=None, **kwargs):
...
-        s += ']'
-        return s
+            row_str = ', '.join(map(
+                lambda e: str(e) if isinstance(e, typ) else repr(e),
+                (self[i, j] for j in range(self.__cols))
+            ))
+            rows.append('[' + row_str + ']')
+
+        res = ',\n        '.join(rows)
+        return f'matrix([{res}])'
```
skirpichev, line 380:
```suggestion
            s += '['
            for j in range(self.__cols):
                if not isinstance(self[i,j], typ):
                    a = repr(self[i,j])
                else:
                    a = "'" + str(self[i,j]) + "'"
                s += a + ', '
            if s[-1] != '[':
                s = s[:-2]
            s += '],\n '
        if s[-1] != '[':
            s = s[:-3]
        s += ']'
        return s    
```

mpmath/matrices/matrices.py
```diff
@@ -392,11 +386,7 @@ def tolist(self):
         return [[self[i,j] for j in range(self.__cols)] for i in range(self.__rows)]
 
     def __repr__(self):
-        if self.ctx.pretty:
-            return self.__str__()
-        s = 'matrix(\n'
-        s += self._toliststr(avoid_type=True) + ')'
-        return s
+        return str(self) if self.ctx.pretty else self._toliststr()
```
skirpichev, line 389:
```suggestion
        if self.ctx.pretty:
            return self.__str__()
        s = 'matrix(\n'
        s += self._toliststr() + ')'
        return s
```

mpmath/matrices/matrices.py
```diff
@@ -362,28 +363,27 @@ def __nstr__(self, n=None, **kwargs):
...
-            s = s[:-2]
+            if s[-1] != '[':
+                s = s[:-2]
             s += '],\n '
-        s = s[:-3]
+        if s[-1] != '[':
+            s = s[:-3]
         s += ']'
-        return s
+        return s
```
skirpichev, line 386:
```suggestion
        return s
```

mpmath/matrices/matrices.py
```diff
@@ -280,38 +280,39 @@ def __init__(self, *args, **kwargs):
...
         self._LU = None
+
         if "force_type" in kwargs:
             warnings.warn("The force_type argument was removed, it did not work"
                 " properly anyway. If you want to force floating-point or"
                 " interval computations, use the respective methods from `fp`"
                 " or `mp` instead, e.g., `fp.matrix()` or `iv.matrix()`."
                 " If you want to truncate values to integer, use .apply(int) instead.")
-        if isinstance(args[0], (list, tuple)):
-            if isinstance(args[0][0], (list, tuple)):
```
skirpichev, line 290: Are there reasons we have such huge changes in the constructor just for one failing case (``matrix([])``)?  I would expect something like
```diff
diff --git a/mpmath/matrices/matrices.py b/mpmath/matrices/matrices.py
index 7447ce1..f1b49b3 100644
--- a/mpmath/matrices/matrices.py
+++ b/mpmath/matrices/matrices.py
@@ -287,7 +287,11 @@ def __init__(self, *args, **kwargs):
                 " or `mp` instead, e.g., `fp.matrix()` or `iv.matrix()`."
                 " If you want to truncate values to integer, use .apply(int) instead.")
         if isinstance(args[0], (list, tuple)):
-            if isinstance(args[0][0], (list, tuple)):
+            if not args[0]:
+                A = args[0]
+                self.__rows = 0
+                self.__cols = 0
+            elif isinstance(args[0][0], (list, tuple)):
                 # interpret nested list as matrix
                 A = args[0]
                 self.__rows = len(A)
```

skirpichev (changes requested):
1. I would expect minimal changes in the matrix constructor
2. pretty printing should be fixed too
3. pr should have printing tests

mpmath/function_docs.py
```diff
@@ -10082,11 +10082,11 @@
...
-    [1.0  0.0  0.0  0.0  0.0]
-    [0.0  1.0  0.0  0.0  0.0]
-    [0.0  0.0  1.0  0.0  0.0]
-    [0.0  0.0  0.0  1.0  0.0]
-    [0.0  0.0  0.0  0.0  1.0]
+    [[1.0  0.0  0.0  0.0  0.0]
+     [0.0  1.0  0.0  0.0  0.0]
+     [0.0  0.0  1.0  0.0  0.0]
+     [0.0  0.0  0.0  1.0  0.0]
+     [0.0  0.0  0.0  0.0  1.0]]
```
skirpichev, line 10089: You break a lot of tests!

skirpichev (changes requested):
You shouldn't break pretty-printing tests.

mart-mihkel:
Sorry for the mess

skirpichev:
Thank you for your work.

PR #758, opened 2024-02-23
Do not build depend on pip

tornaria:
It seems to me that pip is not necessary to build.

In fact, I succesfully built 1.4.0a0 without pip, passing `--skip-dependency-check` option to `python -m build` so it won't complain.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,5 +1,5 @@
 [build-system]
-requires = ['pip>=20.3', 'setuptools>=61', 'wheel',
+requires = ['setuptools>=61', 'wheel',
             'setuptools_scm[toml]>=6.0']
 build-backend = 'setuptools.build_meta'
 [project]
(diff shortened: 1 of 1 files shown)


skirpichev (changes requested):
Yeah, please remove also the wheel package.  See e.g. https://github.com/aleaxit/gmpy/pull/432  
Actually, I have similar patch in another set of small fixes.

Thanks for noticing this!

tornaria:
Not sure if there's a way to squash both commits from the gh interface. I can do it if you want.

Also, if you have this in a different branch, feel free to drop and close this commit if you prefer.

FWIW, I built and tested 1.4.0a0 for our void linux package, x86_64 (glibc), x86_64 (musl) and i686 (glibc).

skirpichev (approved):
LGTM, thanks.

PR #760, opened 2024-02-28
Added kron function 

cicilapetitesorciere:
I've added a new function, `kron`, which is has similar functionality to [`numpy.kron`](https://numpy.org/doc/stable/reference/generated/numpy.kron.html)


mpmath/matrices/matrices.py
```diff
@@ -993,3 +993,26 @@ def mnorm(ctx, A, p=1):
             return max(ctx.fsum((A[i,j] for j in range(n)), absolute=1) for i in range(m))
         else:
             raise NotImplementedError("matrix p-norm for arbitrary p")
+        
+    def kron(ctx, *args):
+        """
+        Calculates the tensor products of 2 or more matrices, similar to numpy.kron
+        """
+        new_rows: int = 1
+        new_cols: int = 1
```
skirpichev, line 1002:
```suggestion
        new_rows = 1
        new_cols = 1
```
We don't include type hints.

mpmath/matrices/matrices.py
```diff
@@ -993,3 +993,26 @@ def mnorm(ctx, A, p=1):
             return max(ctx.fsum((A[i,j] for j in range(n)), absolute=1) for i in range(m))
         else:
             raise NotImplementedError("matrix p-norm for arbitrary p")
+        
+    def kron(ctx, *args):
+        """
+        Calculates the tensor products of 2 or more matrices, similar to numpy.kron
```
skirpichev, line 999: We should have a proper docstring.
1. It's not the tensor product, but a special form of it.
2. We can't just reference numpy.kron here

See other docstrings in the module.

mpmath/matrices/matrices.py
```diff
@@ -993,3 +993,26 @@ def mnorm(ctx, A, p=1):
...
+            for j in range(new_cols):
+                partition_rows: int = 1
+                partition_cols: int = 1
+                for m in args:
+                    partition_rows *= m.rows
+                    partition_cols *= m.cols
+                    res[i, j] *= m[
+                            int(i * partition_rows / new_rows) % m.rows, 
+                            int(j * partition_cols / new_cols) % m.cols
+                        ]
```
skirpichev, line 1017: Why not use same two-arguments form as numpy.kron?  As TP is associative, we could use reduce to compute more general case.

Given ``def kron(ctx, a, b):``,
```suggestion
                res[i, j] = a[i // b.rows, j // b.cols]*b[i % b.rows, j % b.cols]
```

skirpichev (changes requested):
Also, this need tests.

mpmath/matrices/matrices.py
cicilapetitesorciere: I'm not sure what would be gained from doing this, and it would make things significantly less efficient for calculations involving more than two inputs since it would involve allocating and re-allocating memory for successively larger intermediate matrices

cicilapetitesorciere:
I'll get the docstring and tests committed soonish

mpmath/matrices/matrices.py
```diff
@@ -998,3 +998,64 @@ def mnorm(ctx, A, p=1):
...
+        new_rows = 1
+        new_cols = 1
+        for m in args:
+            new_rows *= m.rows
+            new_cols *= m.cols
+        res = ctx.ones(new_rows, new_cols)
+        for i in range(new_rows):
+            for j in range(new_cols):
+                partition_rows: int = 1
+                partition_cols: int = 1
```
skirpichev, line 1053:
```suggestion
                partition_rows = 1
                partition_cols = 1
```

mpmath/matrices/matrices.py
```diff
@@ -998,3 +998,64 @@ def mnorm(ctx, A, p=1):
...
+                for m in args:
+                    partition_rows *= m.rows
+                    partition_cols *= m.cols
+                    res[i, j] *= m[
+                            int(i * partition_rows / new_rows) % m.rows,
+                            int(j * partition_cols / new_cols) % m.cols
```
skirpichev, line 1059:
```suggestion
                            (i * partition_rows // new_rows) % m.rows,
                            (j * partition_cols // new_cols) % m.cols
```

mpmath/matrices/matrices.py
skirpichev: > I'm not sure what would be gained from doing this

Simple API (compatible wrt numpy), obvious code.  You also could initialize res as an empty matrix.

> significantly less efficient for calculations involving more than two inputs

How much gain we well have, could argue this?  And/Or provide benchmarks?  I would expect, that multiplication cost will be dominant, esp. for the mp context.

```
In [21]: %timeit -r11 r=mp.kron(a,a,a)
3.98 ms ± 3.72 µs per loop (mean ± std. dev. of 11 runs, 100 loops each)

In [22]: %timeit -r11 r=reduce(mp.kron, [a, a, a])
3.55 ms ± 12.2 µs per loop (mean ± std. dev. of 11 runs, 100 loops each)

In [23]: a = fp.matrix([[1, 2], [3, 4]])

In [24]: %timeit -r11 r=reduce(fp.kron, [a, a, a])
1.53 ms ± 16.5 µs per loop (mean ± std. dev. of 11 runs, 1,000 loops each)

In [25]: %timeit -r11 r=fp.kron(a,a,a)
1.71 ms ± 5.35 µs per loop (mean ± std. dev. of 11 runs, 1,000 loops each)

In [26]: def kron2(ctx, a, b):
    ...:     new_rows = a.rows*b.rows
    ...:     new_cols = a.cols*b.cols
    ...:     res = ctx.matrix(new_rows, new_cols)
    ...:     for i in range(new_rows):
    ...:         for j in range(new_cols):
    ...:             res[i, j] = a[i // b.rows, j // b.cols]*b[i % b.rows, j % b.cols]
    ...:     return res
    ...: 

In [27]: kron2fp = lambda a, b: kron2(fp, a, b)

In [28]: %timeit -r11 r=reduce(kron2fp, [a, a, a])
632 µs ± 1.66 µs per loop (mean ± std. dev. of 11 runs, 1,000 loops each)

In [29]: kron2mp = lambda a, b: kron2(mp, a, b)

In [30]: a = mp.matrix([[1, 2], [3, 4]])

In [31]: %timeit -r11 r=reduce(kron2mp, [a, a, a])
1.41 ms ± 12 µs per loop (mean ± std. dev. of 11 runs, 1,000 loops each)
```

skirpichev (changes requested):
I would like to see some benefits of multiple-arguments version.

cicilapetitesorciere:
SymPy supports an arbitrary number of arguments in [their tensor product](https://docs.sympy.org/latest/modules/physics/quantum/tensorproduct.html) so it would be more compatible with that. And since it still can take two arguments it doesn't really lose any compatibility with `numpy`. Also I don't think most people know about `functools.reduce`. I didn't until this this conversation, and neither did the original author of [the repo I originally wrote this function for](https://github.com/cicilapetitesorciere/magicstates). So in practice people won't just use `functools.reduce` for the general case, they'll write a whole new function.

oscarbenjamin:
How often does n-ary Kronecker product come up in practice?

Some people might not know about `reduce` but they can probably figure out `kron(a, kron(b, c))`.

There is always this kind of tension in API design between simple API that is easy to design, understand and implement vs the temptation to generalise everything as much as possible. It might seem like generalising things is always a good thing but there are always downsides like the interface is harder to define and explain. The extra generalisation makes it harder to improve, optimise or refactor the code in future etc. There is also value in matching the interface and signature of other libraries/environments.

The performance argument can be swapped around: if binary kron is the common case then would we end up having a faster binary kron if we were not trying to implement n-ary kron?

SymPy's `KroneckerProduct` is a symbolic representation of an associative operation which by convention means that it is represented as an n-ary operation:
```
In [3]: A, B, C = [MatrixSymbol(s, 2, 2) for s in 'ABC']

In [4]: KroneckerProduct(A, B)
Out[4]: A ⨂ B

In [5]: KroneckerProduct(A, B, C)
Out[5]: A ⨂ B ⨂ C
```
Both numpy and Matlab have `kron` functions that are binary functions taking a single argument and are more like the `kron` function here.

AthenaCaesura:
Hello! I'm Cici's collaborator on this project, thought I might have some insight here. We are implementing `kron` so that we can use mpmath in to perform high accuracy simulations of quantum dynamics. So keep in mind, I'm a little biased towards quantum computing use cases and my examples will probably reflect that. 😅 

For your first question: 
Typically, `kron` would be used to concatenate smaller (2x2) Hilbert spaces together in order to simulate the full quantum computer which might be a $2^{20} \times 2^{20}$ matrix. This often requires concatenating multiple matrices, and so the n-ary kronecker product would be helpful here.

As for possibly confusing new users, I think this can be prevented just by including some usage examples which show that you can use two or more arguments to the `kron` function. New users typically gravitate towards examples anyway.

For your second question:
I don't think we could see any performance improvement by making `kron` a binary operator. In fact, you might be able to improve performance in the future by taking advantage of the order in which one performs kronecker products. Although I wouldn't worry about this optimization for a while.

skirpichev:
> I don't think most people know about functools.reduce

You can do same, using simple for loop:
```
In [1]: from mpmath import *

In [2]: a = mp.matrix([[1, 2], [3, 4]])

In [3]: %timeit r = mp.kron(a, a, a)
3.99 ms ± 6.97 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

In [4]: def kronv2(a, b):
   ...:     new_rows = a.rows*b.rows
   ...:     new_cols = a.cols*b.cols
   ...:     res = mp.matrix(new_rows, new_cols)
   ...:     for i in range(new_rows):
   ...:         for j in range(new_cols):
   ...:             res[i, j] = a[i // b.rows, j // b.cols]*b[i % b.rows, j % b.cols]
   ...:     return res
   ...: 

In [5]: from functools import reduce

In [6]: %timeit r = reduce(kronv2, [a, a, a])
1.36 ms ± 4.21 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)

In [7]: def myreduce(func, it):
   ...:     it = iter(it)
   ...:     val = next(it)
   ...:     for elt in it:
   ...:         val = func(val, elt)
   ...:     return val
   ...: 

In [8]: %timeit r = myreduce(kronv2, [a, a, a])
1.36 ms ± 1.26 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
```

> So in practice people won't just use functools.reduce for the general case

I'm not sure.  This function is a part of the stdlib and in general it's a good idea to learn the language used.

(BTW, not sure why this is not a part of builtins.  map-reduce - very common programming pattern.)

> This often requires concatenating multiple matrices, and so the n-ary kronecker product would be helpful here.

I don't see any problems with using 2-ary functions from the numpy to use for n-ary case.  BTW you also could use ``np.kron`` to work with mpmath's types:
```
In [18]: b = np.matrix(a).reshape(2, 2)

In [19]: %timeit r = reduce(np.kron, [b, b, b])
717 µs ± 1.74 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
```

That brings another question to me.  Should we mirror every function from the numpy, even if there is nothing specific to the mpmath in it's implementation?  Another example is #764.

> I don't think we could see any performance improvement by making kron a binary operator.

That's fine, even if simple benchmarks above could disprove this claim.  But I'm worry more about advantages of n-ary implementation.  So far we have these arguments:

1. "it would make things significantly less efficient for calculations involving more than two inputs"
2. most people don't know about functools.reduce

I doubt we have seen any proofs for (1).  The next argument doesn't look too severe for me.

> you might be able to improve performance in the future by taking advantage of the order in which one performs kronecker products

What kind of?

oscarbenjamin:
> As for possibly confusing new users, I think this can be prevented just by including some usage examples which show that you can use two or more arguments to the `kron` function.

The point is that this confuses things for users who just want the 2 argument kron function because the definition of what the function does in general is more complicated. It is generally better just to define simple clean API and also to mirror what analogous APIs do.

Performance is a red herring here because mpmath's matrix type is not well optimised at all. It uses a sparse internal data structure but then all operations loop over all rows and columns meaning that they never benefit from sparseness. Inner loops use the slow `__setitem__` and `__getitem__` methods everywhere (like all `kron` implementations shown above). There is no way to efficiently construct the internal `__data` dict and then make a new `mp.matrix` from that etc.

If we want to optimise things then all the methods of `mp.matrix` can be better optimised easily. The way to make a kron function that could benefit from that is not to micro-optimise `kron` but rather to implement it in terms of higher-level methods and optimise those instead. This is how I would do it with primitives available in sympy for example:
```python
def kron(a, b):
    return Matrix([[aij*b for aij in ai] for ai in a.tolist()])
```
Now the task is to make all of the components that are used here efficient e.g.:

1. Matrix-scalar multiplication should be made faster.
2. The `tolist()` method should be made faster.
3. mpmath has no function for concatenating matrices.
4. It might be better to have something like `todict()` since the internal data structure is a dict and we can exploit potential sparseness.

It isn't worth even talking about performance/benchmarks for a `kron` function while everything else is so inefficient. It also seems odd to me to want to add this `kron` function rather than improving `matrix` more generally. Especially it seems odd because if mpmath's matrix had more basic utilities available then the `kron` function would be trivial like the sympy implementation that I showed above.

Adding those basic utilities and making basic operations more efficient would be a better use of everyone's time here in mpmath development rather than quibbling API details for something relatively obscure like `kron`.

oscarbenjamin:
Perhaps rather than `kron` it would be better first to add a function analogous to numpy's `block`:
https://numpy.org/doc/stable/reference/generated/numpy.block

skirpichev:
Thanks for your contribution and patience, but I'm closing this pr.

As it mentioned above, the current version has nothing specific to mpmath c.f. the ``np.kron()``, you could just use that function.  If you wish to contribute, take attention to @oscarbenjamin suggestions above.

But feel free to reopen pr if you disagree and have new arguments.

Issue #613, opened 2021-10-30
Issue converting strings with _ to mpf


oscargus:
From Python 3.7 it is possible to use _ in strings converted to floats (and maybe other types), so `float('1_2.3_4')` is valid and gives 12.34.

This causes a problem when converting this type of numbers using `from_str` or more exactly here:
https://github.com/fredrik-johansson/mpmath/blob/c11db84b3237bd8fc6721f5a0c5d7c0c98a24dc1/mpmath/libmp/libmpf.py#L1303-L1306

Should be solved by replacing `parts[1].rstrip('0')` with `parts[1].rstrip('0').replace('_', '')`.

skirpichev:
Actually, this will be a violation of [PEP 515](https://www.python.org/dev/peps/pep-0515/).  But see [this](https://bugs.python.org/issue44267) and [this](https://github.com/aleaxit/gmpy/commit/ec8dabff9730b753f38647fb5af0c4bad7b492cb).

PR #769, opened 2024-03-30
v1.4.0a1

skirpichev:
Lets include some compatibility wrappers for sympy and co, then do alpha release.


oscarbenjamin:
Thanks for doing this. These changes look good.

Actually right now the best thing would just be for mpmath to hold off on making any prereleases though. Right now things seem to be working and there is a risk that further changes might cause further problems. I don't know what strange things people are doing downstream with packaging tools...

I would like to get a sympy release out with an upper cap on mpmath: https://github.com/sympy/sympy/pull/26413 https://github.com/sympy/sympy/issues/26273

Then it should be possible for mpmath to put out prereleases without anything breaking.

skirpichev:
> right now the best thing would just be for mpmath to hold off on making any prereleases though

Ok.  Though, I think changes aren't too big:
* 22859b0 Merge pull request #711 from skirpichev/hex
* afb829b Merge pull request #759 from skirpichev/misc
* d294c04 Merge pull request #761 from skirpichev/pickling
* 3512b74 Merge pull request #763 from skirpichev/misc
* 4c1b435 Merge pull request #767 from skirpichev/misc
* aa1c0b9 Merge pull request #770 from ChefAharoni/patch-1


> Then it should be possible for mpmath to put out prereleases without anything breaking.

Unless someone will install sympy pre-release...

oscarbenjamin:
> > Then it should be possible for mpmath to put out prereleases without anything breaking.
> 
> Unless someone will install sympy pre-release...

I think we just need to ensure that each release of sympy whether it is a full release or a prerelease has an upper cap on mpmath and is fully tested with that capped version.

I want to cap `mpmath < 1.4` so that a mpmath 1.3.1 release would still be picked up in case there is a need to put out 1.3.1 e.g. for Python 3.13 support or something like that.

skirpichev:
> we just need to ensure that each release of sympy whether it is a full release or a prerelease has an upper cap on mpmath

Limiting upper versions does make sense for stable releases.  But for pre-releases - I don't think so, unless you have **many** deps.  People should install pre-releases only intentionally and with understanding of consequences.

oscarbenjamin:
> People should install pre-releases only intentionally and with understanding of consequences.

The purpose of installing the prerelease is to test in advance what can be expected to happen once the final release is released so there should not be any difference in behaviour. I think that a pre-release of e.g. sympy should have version constraints that match what the future final release would be expected to have. For example if I `pip install sympy==1.13.0rc1` then I would hope that this gets the same version of mpmath that `pip install sympy==1.13.0` would be expected to get if there are no changes between rc1 and final.

skirpichev:
> I think that a pre-release of e.g. sympy should have version constraints that match what the future final release would be expected to have.

That depends, e.g. clearly make sense for late rc-releases.  But alpha- and beta-releases - also pre-releases c.f. pypi point of view.

oscarbenjamin:
I would say that the alpha/beta releases are where you bump the dependency constraints.

If someone wants to use `pip install --pre` then they can install prereleases of everything transitively and deal with the bleeding edge consequences. If they just track prereleases of sympy though then the bumped version constraints would be part of what they are tracking.

skirpichev:
Ok, I did tests for mpq in domains - that works too.  The only missing thing from https://github.com/sympy/sympy/pull/25290/commits/eef227fbac2c3542758379277727d9aafc7229fd is ``from_str()`` changes.

It's support for underscores, 9585bb9.  Obviously, it's a compatibility break.  I think it's a good idea to support this, but we could implement that differently.  The current code uses gmpy2 approach: just filter out underscores.  This accepts valid python strings, but also much more.  Instead, we could choose to be strictly python-compatible.  Let me know if it worth to reopen #613 for discussion.  But see also aleaxit/gmpy#210.

oscarbenjamin:
> What do you think?

I think that sympy reaches into mpmath's internals too much and it is inevitable that if mpmath is actively developed then future versions of mpmath are not going to be compatible with current versions of sympy. We can coordinate this to keep sympy up to date with latest mpmath changes but we need sympy releases to have a version cap like `mpmath < 1.4`.

The first thing to consider though before deprecating or removing things is what will ultimately be the public API that should be used instead. Currently the awkward thing about mpmath's documented API is that it is all based on contexts and the idea of a global precision. What is nice about the libmp functions is that they are pure functions that take an explicit precision argument.

skirpichev:
> We can coordinate this to keep sympy up to date with latest mpmath changes

As a first step for the sympy master, I suggest something like https://github.com/diofant/diofant/pull/1361/commits/1dd72fbf4bd9c2cf28fa3d7de850fd3f82359dc5: all required libmp's imports (including the ``giant_steps()``) could be imported from the ``mpmath.libmp`` namespace.  (``from mpmath.libmp import foo, bar`` or ``import mpmath.libmp as libmp``)

> that it is all based on contexts and the idea of a global precision.

And contexts serve both to hold some attributes (like precision or rounding mode) and as namespaces for functions (namespaces, where we actually have same set of mathematical functions for all contexts).

> What is nice about the libmp functions is that they are pure functions that take an explicit precision argument.

Yet the namespace of libmp is huge:
```pycon
>>> len(libmp_names)  # not prefixed by '_'
320
```
While e.g. SymPy (v1.12) is using:
```
>>> len(sympy_names)  # from above script, maybe I miss something
78
```

Not all functions have common naming pattern (e.g. ``mpf_foo``), common arguments convention (say, function arguments, precision, optional rounding mode), have different defaults for prec/rnd, etc...

I think we could start with something, that fits SymPy's needs, with few functions being deprecated (e.g. to/from_pickable() or bitcount()) and treat the rest of libmp as private stuff.

skirpichev:
Ok, let me know when I can release a new alpha.

oscarbenjamin:
The latest sympy prerelease (`sympy 1.12.1rc1`) pins mpmath < 1.4.0 so it should be fine. Go ahead and put out prereleases if you want to get them out.

Thanks.

skirpichev:
v1.4.0a1 released, I hope https://github.com/sympy/sympy/issues/26273 is fixed.

oscarbenjamin:
It should be fine as far as prereleases go. Thanks Sergey.

Issue #720, opened 2023-09-23
Remove bitcount field of the mpf tuple?


skirpichev:
[Our docs](https://mpmath.readthedocs.io/en/latest/technical.html#representation-of-numbers) says:
> More technically, mpmath represents a floating-point number as a 4-tuple (sign, man, exp, bc) where sign is 0 or 1 (indicating positive vs negative) and the mantissa is nonnegative; bc (bitcount) is the size of the absolute value of the mantissa as measured in bits. Though redundant, keeping a separate sign field and explicitly keeping track of the bitcount significantly speeds up arithmetic (the bitcount, especially, is frequently needed but slow to compute from scratch due to the lack of a Python built-in function for the purpose).

Since CPython 3.1 there is a such function.  It's the [int.bit_length()](https://docs.python.org/3/library/stdtypes.html#int.bit_length) method.

Some timings:
```
$ python -m timeit -r11 -s 'from mpmath.libmp.libintmath import python_bitcount as bc' 'bc(1000)'
500000 loops, best of 11: 965 nsec per loop
$ python -m timeit -r11 -s 'from mpmath.libmp.libintmath import python_bitcount as bc' 'bc(10)'
500000 loops, best of 11: 930 nsec per loop
$ python -m timeit -r11 -s 'from mpmath.libmp.libintmath import python_bitcount as bc' -s 'from math import factorial;n=factorial(100)' 'bc(n)'
100000 loops, best of 11: 3.18 usec per loop
```
vs
```
$ python -m timeit -r11 '(1000).bit_length()'
5000000 loops, best of 11: 75.6 nsec per loop
$ python -m timeit -r11 '(10).bit_length()'
5000000 loops, best of 11: 74.9 nsec per loop
$ python -m timeit -r11 -s 'from math import factorial;n=factorial(100)' 'n.bit_length()'
5000000 loops, best of 11: 97.2 nsec per loop
```

So, using the bit_length() is clearly better and it's available for all backends.  Does it make sense to keep now the bc field of the mpf tuple or it should be deprecated?

oscarbenjamin:
The tricky thing is that the mpf tuple is used as a conversion format in many places so there would still be a need to generate the 4-tuple with bit count often.

Otherwise though I agree that a 3 tuple is better and it can save a lot of wasted arithmetic (and bugs) trying to keep the bit count in sync.

skirpichev:
[...]

Issue #704, opened 2023-05-21
import sympy fails with mpmath master


oscarbenjamin:
Since 3414c6ec51c3393b2baca8f5062c73cfa8184372 from gh-701 import sympy fails:
```python
>>> import sympy
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/oscar/current/active/sympy/sympy/__init__.py", line 30, in <module>
    from sympy.core.cache import lazy_function
  File "/home/oscar/current/active/sympy/sympy/core/__init__.py", line 9, in <module>
    from .expr import Expr, AtomicExpr, UnevaluatedExpr
  File "/home/oscar/current/active/sympy/sympy/core/expr.py", line 4159, in <module>
    from .mul import Mul
  File "/home/oscar/current/active/sympy/sympy/core/mul.py", line 2193, in <module>
    from .numbers import Rational
  File "/home/oscar/current/active/sympy/sympy/core/numbers.py", line 30, in <module>
    from mpmath.ctx_mp import mpnumeric
ImportError: cannot import name 'mpnumeric' from 'mpmath.ctx_mp' (/home/oscar/current/active/mpmath/mpmath/ctx_mp.py)
```
The `mpnumeric` class is the base class for `_mpf` and `_mpc` and is used by sympy for detecting when an object is an instance of `mpf` or `mpc`. This is needed since `_mpf` and `_mpc` seem to be designated private and `mpf` and `mpc` are dynamically generated types (gh-696, gh-532).

The fix for that is straight-forward:
```diff
diff --git a/mpmath/ctx_mp.py b/mpmath/ctx_mp.py
index c17017e..6cf8e33 100644
--- a/mpmath/ctx_mp.py
+++ b/mpmath/ctx_mp.py
@@ -31,6 +31,9 @@
     from .ctx_mp_python import PythonMPContext as BaseMPContext
 
 
+from .ctx_mp_python import _mpf, _mpc, mpnumeric
+
+
 class MPContext(BaseMPContext, StandardBaseContext):
     """
     Context for multiprecision arithmetic with a global precision.
```
However then a new error arises:
```python
>>> import sympy
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/oscar/current/active/sympy/sympy/__init__.py", line 30, in <module>
    from sympy.core.cache import lazy_function
  File "/home/oscar/current/active/sympy/sympy/core/__init__.py", line 9, in <module>
[...]

PR #783, opened 2024-04-10
Misc fixes

skirpichev:
* don't unpack mpf tuples directly in ctx_mp*.py
* add signed option to to_man_exp()

### DIFF
--- a/mpmath/ctx_mp.py
+++ b/mpmath/ctx_mp.py
@@ -16,7 +16,8 @@ from .libmp import (MPQ, MPZ_ONE, ComplexResult, dps_to_prec, finf, fnan,
                     mpf_apery, mpf_catalan, mpf_degree, mpf_div, mpf_e,
                     mpf_euler, mpf_glaisher, mpf_khinchin, mpf_ln2, mpf_ln10,
                     mpf_mertens, mpf_mul, mpf_neg, mpf_phi, mpf_pi, mpf_rand,
-                    mpf_sub, mpf_twinprime, repr_dps, to_str)
+                    mpf_sign, mpf_sub, mpf_twinprime, repr_dps, to_man_exp,
+                    to_str)
 
 
 get_complex = re.compile(r"""
@@ -337,8 +338,10 @@ class MPContext(BaseMPContext, StandardBaseContext):
         if not x:
             return True
         if hasattr(x, '_mpf_'):
-            sign, man, exp, bc = x._mpf_
-            return sign and exp >= 0
+            if ctx.isfinite(x):
+                _, exp = to_man_exp(x._mpf_)
+                return mpf_sign(x._mpf_) < 0 and exp >= 0
+            return False
         if hasattr(x, '_mpc_'):
             return not x.imag and ctx.isnpint(x.real)
         if type(x) in int_types:
@@ -1113,21 +1116,19 @@ maxterms, or set zeroprec."""
             im_dist = ctx.ninf
         elif hasattr(x, "_mpc_"):
             re, im = x._mpc_
-            isign, iman, iexp, ibc = im
+            iman, iexp = to_man_exp(im)
             if iman:
-                im_dist = iexp + ibc
-            elif im == fzero:
-                im_dist = ctx.ninf
+                im_dist = iexp + iman.bit_length()
             else:
-                raise ValueError("requires a finite number")
+                im_dist = ctx.ninf
         else:
             x = ctx.convert(x)
             if hasattr(x, "_mpf_") or hasattr(x, "_mpc_"):
                 return ctx.nint_distance(x)
             else:
                 raise TypeError("requires an mpf/mpc")
-        sign, man, exp, bc = re
-        mag = exp+bc
+        man, exp = to_man_exp(re)
+        mag = exp+man.bit_length()
         # |x| < 0.5
         if mag < 0:
             n = 0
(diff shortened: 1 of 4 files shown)


mpmath/ctx_mp.py
```diff
@@ -1113,21 +1117,25 @@ def nint_distance(ctx, x):
...
             if hasattr(x, "_mpf_") or hasattr(x, "_mpc_"):
                 return ctx.nint_distance(x)
             else:
                 raise TypeError("requires an mpf/mpc")
-        sign, man, exp, bc = re
-        mag = exp+bc
+        try:
+            man, exp = to_man_exp(re)
+        except ValueError:
+            raise ValueError("requires a finite number")
```
pearu, line 1137: Considering the number of similar try-except blocks in this PR, would it make sense to introduce a `isfinite` function to mpmath/libmp/libmpf.py:
```python
def isfinite(s):
    sign, man, exp, bc = s
    return (not man) and exp
```
so that here and elsewhere one can use
```python
if not isfinite(re):
    raise ValueError(...)
man, exp = to_man_exp(re)
```

mpmath/ctx_mp_python.py
```diff
@@ -117,9 +117,9 @@ def mpf_convert_lhs(cls, x):
         return x
 
     man_exp = property(lambda self: to_man_exp(self._mpf_))
-    man = property(lambda self: self._mpf_[1])
-    exp = property(lambda self: self._mpf_[2])
-    bc = property(lambda self: self._mpf_[3])
+    man = property(lambda self: self.man_exp[0])
+    exp = property(lambda self: self.man_exp[1])
+    bc = property(lambda self: self.man.bit_length())
```
pearu, line 122: OT: it looks like the 4-th item in the `_mpf_` tuple could be eliminated. That is, can we assume that
```python
assert _mpf_[1].bit_length() == _mpf_[3]
```
holds and what will be the performance implications?

pearu (approved):
I have a couple of comments but otherwise LGTM. Thanks, @skirpichev!

mpmath/ctx_mp.py
skirpichev: In fact, there is such a function (on context instance).  I'll refactor 1st commit, thanks.

mpmath/ctx_mp_python.py
skirpichev: > OT: it looks like the 4-th item in the _mpf_ tuple could be eliminated.

This is true (and maybe the *first* item as well, if we decide to live without signed zero).  But we could do this only via series of backward compatibility breaks.  See https://github.com/mpmath/mpmath/issues/720

mpmath/ctx_mp.py
skirpichev: I hope, the 3rd commit fixes this issue.

pearu (approved):
I have a couple of BE nits but otherwise LGTM!

mpmath/libmp/libmpf.py
```diff
@@ -235,13 +236,24 @@ def from_int(n, prec=0, rnd=round_fast):
             return int_cache[n]
     return from_man_exp(n, 0, prec, rnd)
 
-def to_man_exp(s):
+def to_man_exp(s, signed=None):
     """Return (man, exp) of a raw mpf. Raise an error if inf/nan."""
+    if signed is None:
+        warnings.warn("Returning unsigned mantissa value per default "
+                      "is deprecated.  Please use signed=False to "
+                      "keep this behaviour.", DeprecationWarning)
```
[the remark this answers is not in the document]
skirpichev: > Also, what is the aim deprecation?

The aim is to deprecate direct operations with components of the mpf tuple.  That's one step on a way to solve issue https://github.com/mpmath/mpmath/issues/720.

To access mantissa and exponent - to_man_exp(s) should be used, not ``s[1]`` or ``s[2]``.

> Does it mean that signed=True will be default behavior in the future?

Yes, I think it does make sense:
1. for compatibility with other libraries like gmpy2 (mpfr has as_mantissa_exp() method)
2. to support from_man_exp/to_man_exp round-trip per default (for normal finite numbers)

> If so, when?

I'm planning to add above warning in the next release (v1.14).  Then, we could change default value (``None`` -> ``True``) and deprecate the ``signed`` option in the v1.15.

> Would it make sense to describe this in the doc-string or/and in the warning message?

Functions in the libmp/ - kinda private stuff, see #704.  I'll adjust the warning message.

mpmath/ctx_mp.py
```diff
@@ -1113,21 +1117,25 @@ def nint_distance(ctx, x):
...
             if hasattr(x, "_mpf_") or hasattr(x, "_mpc_"):
                 return ctx.nint_distance(x)
             else:
                 raise TypeError("requires an mpf/mpc")
-        sign, man, exp, bc = re
-        mag = exp+bc
+        if mpf_is_finite(re):
+            man, exp = to_man_exp(re, signed=False)
+        else:
+            raise ValueError("requires a finite number")
```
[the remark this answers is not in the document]
skirpichev: > to_man_exp already raises ValueError on non-finite inputs with slightly different message but basically with the same meaning.

The exception message seems too "low-level" for me.  But maybe we could change that to something like:
1. "finite number expected"
2. "mantissa and exponent are undefined"
3. ???

?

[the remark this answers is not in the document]
pearu: "mantissa and exponent are defined for finite numbers only"?

skirpichev:
Bugfix for tan/tanh moved to #785

mpmath/ctx_mp.py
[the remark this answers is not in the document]
skirpichev: > "mantissa and exponent are defined for finite numbers only"?

LGTM, commited.

PR #791, opened 2024-04-19
Skip np.float128 assertion in test_convert if type doesn't exist

d-torrance:
It isn't available on all systems (e.g., 32-bit x86 machines).


skirpichev (approved):

skirpichev:
Thanks.

PR #792, opened 2024-04-23
Fallback to fixed precision if prec change is infinite

### LINKED ISSUE #520: besselk function doesn't support limit approaching infinity
SciPy, GSL, and jmat all support evaluating the limit of the besselk function as it approaches infinity, returning zero. However, neither the fp nor the mp contexts for mpmath do so even though limits are in other functions.

```
>>> mp.besselk(0, inf)
  File "/usr/local/lib/python3.6/dist-packages/mpmath/ctx_mp_python.py", line 1035, in f_wrapped
    retval = f(ctx, *args, **kwargs)
  File "/usr/local/lib/python3.6/dist-packages/mpmath/functions/bessel.py", line 169, in besselk
    ctx.prec += M
  File "/usr/local/lib/python3.6/dist-packages/mpmath/ctx_mp_python.py", line 611, in _set_prec
    ctx._prec = ctx._prec_rounding[0] = max(1, int(n))
  File "/usr/local/lib/python3.6/dist-packages/mpmath/ctx_mp_python.py", line 141, in __int__
    def __int__(s): return int(to_int(s._mpf_))
  File "/usr/local/lib/python3.6/dist-packages/mpmath/libmp/libmpf.py", line 351, in to_int
    raise ValueError("cannot convert inf or nan to int")
ValueError: cannot convert inf or nan to int
>>> mp.besselk(1, inf)
  File "/usr/local/lib/python3.6/dist-packages/mpmath/ctx_mp_python.py", line 1035, in f_wrapped
    retval = f(ctx, *args, **kwargs)
  File "/usr/local/lib/python3.6/dist-packages/mpmath/functions/bessel.py", line 169, in besselk
    ctx.prec += M
  File "/usr/local/lib/python3.6/dist-packages/mpmath/ctx_mp_python.py", line 611, in _set_prec
    ctx._prec = ctx._prec_rounding[0] = max(1, int(n))
  File "/usr/local/lib/python3.6/dist-packages/mpmath/ctx_mp_python.py", line 141, in __int__
    def __int__(s): return int(to_int(s._mpf_))
  File "/usr/local/lib/python3.6/dist-packages/mpmath/libmp/libmpf.py", line 351, in to_int
    raise ValueError("cannot convert inf or nan to int")
ValueError: cannot convert inf or nan to int
>>> fp.besselk(1, inf)
  File "/usr/local/lib/python3.6/dist-packages/mpmath/ctx_fp.py", line 57, in f_wrapped
    return f(ctx, *args, **kwargs)
  File "/usr/local/lib/python3.6/dist-packages/mpmath/functions/bessel.py", line 173, in besselk
    return ctx.hypercomb(h, [n], **kwargs)
  File "/usr/local/lib/python3.6/dist-packages/mpmath/functions/hypergeometric.py", line 127, in hypercomb
    [ctx.rgamma(b) for b in beta_s] + \
  File "/usr/local/lib/python3.6/dist-packages/mpmath/functions/hypergeometric.py", line 219, in hyper
    if   q == 1: return ctx._hyp0f1(b_s, z, **kwargs)
  File "/usr/local/lib/python3.6/dist-packages/mpmath/functions/hypergeometric.py", line 307, in _hyp0f1
    return ctx.hypsum(0, 1, (btype,), [b], z, **kwargs)
  File "/usr/local/lib/python3.6/dist-packages/mpmath/ctx_fp.py", line 213, in hypsum
    raise ctx.NoConvergence
mpmath.libmp.libhyper.NoConvergence
```

### LINKED ISSUE #522: Unexpected return from hyp1f1 when evaluating limit at negative infinity
Though mpmath appears to support evaluating limits approaching positive infinity, matching the results returned by SciPy and GSL, and jmat, it diverges from those three libraries when evaluating the limit at negative infinity with the first argument as 1; `mp.hyp1f1` returns NaN when the other libraries agree on zero.

```
>>> special.hyp1f1(0, 1, np.inf)
1.0
>>> mp.hyp1f1(0, 1, np.inf)
mpf('1.0')
>>> special.hyp1f1(0, 1, np.NINF)
1.0
>>> mp.hyp1f1(0, 1, np.NINF)
mpf('1.0')
>>> special.hyp1f1(1, 2, np.NINF)
0.0
>>> mp.hyp1f1(1, 2, np.NINF)
mpf('nan')
>>> special.hyp1f1(2, 2, np.NINF)
0.0
>>> mp.hyp1f1(2, 2, np.NINF)
mpf('0.0')
>>> mp.hyp1f1(1, 5, np.NINF)
mpf('nan')
>>> 
```

skirpichev:
This will leave handling of special numbers to fewer functions, e.g. to hypercomb().

Closes #520
Closes #522

### DIFF
--- a/mpmath/ctx_mp_python.py
+++ b/mpmath/ctx_mp_python.py
@@ -775,6 +775,8 @@ class PythonMPContext:
         ctx.trap_complex = False
 
     def _set_prec(ctx, n):
+        if not ctx.isfinite(n):
+            return
         ctx._prec = ctx._prec_rounding[0] = max(1, int(n))
         ctx._dps = prec_to_dps(n)
 
--- a/mpmath/functions/hypergeometric.py
+++ b/mpmath/functions/hypergeometric.py
@@ -316,10 +316,8 @@ def _hyp1f1(ctx, a_s, b_s, z, **kwargs):
         return ctx.one+z
     magz = ctx.mag(z)
     if magz >= 7 and not (ctx.isint(a) and ctx.re(a) <= 0):
-        if ctx.isinf(z):
-            if ctx.sign(a) == ctx.sign(b) == ctx.sign(z) == 1:
-                return ctx.inf
-            return ctx.nan * z
+        if ctx.isinf(z) and ctx.sign(a) == ctx.sign(b) == ctx.sign(z) == 1:
+            return ctx.inf
         try:
             try:
                 ctx.prec += magz
--- a/mpmath/tests/test_functions2.py
+++ b/mpmath/tests/test_functions2.py
@@ -85,6 +85,7 @@ def test_bessel():
     assert besselk(0,3+4j).ae(-0.007239051213570155013+0.026510418350267677215j)
     assert besselk(0,j).ae(-0.13863371520405399968-1.20196971531720649914j)
     assert (besselk(3, 10**10) * mpf(10)**4342944824).ae(1.1628981033356187851)
+    assert besselk(1,inf) == 0
     # test for issue 331, bug reported by Michael Hartmann
     for n in range(10,100,10):
         mp.dps = n
@@ -482,6 +483,12 @@ def test_hyper_1f1():
     assert hyp1f1(-2, 1, 10000).ae(49980001)
     # Bug
     assert hyp1f1(1j,fraction(1,3),0.415-69.739j).ae(25.857588206024346592 + 15.738060264515292063j)
+    # issue 522
+    assert hyp1f1(0, 1, +inf) == 1
+    assert hyp1f1(0, 1, -inf) == 1
+    assert hyp1f1(1, 2, -inf) == 0
+    assert hyp1f1(2, 2, -inf) == 0
+    assert hyp1f1(1, 5, -inf) == 0
 
 def test_hyper_2f1():
     v = 1.0652207633823291032


mpmath/ctx_mp_python.py
```diff
@@ -775,6 +775,8 @@ def default(ctx):
         ctx.trap_complex = False
 
     def _set_prec(ctx, n):
+        if not ctx.isfinite(n):
+            return
         ctx._prec = ctx._prec_rounding[0] = max(1, int(n))
```
oscarbenjamin, line 780: This seems like an odd fix: just ignore the precision if it is `mpmath.inf`?

Is it valid to set the `ctx.prec = mpmath.inf`?

If it is valid then surely it should be used for something rather than ignored. If it is not valid then surely it should raise an error rather than being ignored. Either way ignoring does not seem like the right thing to do.

I would have thought that the right way to fix this is here by checking if `M` is infinite before doing `ctx.prec += M`:
https://github.com/mpmath/mpmath/blob/dcc591cb678dcef0c0b37f45d430d105ee258766/mpmath/functions/bessel.py#L159-L172

skirpichev: > This seems like an odd fix: just ignore the precision if it is mpmath.inf?

Yes, but see: https://github.com/mpmath/mpmath/pull/792#issuecomment-2071327300

> Is it valid to set the ctx.prec = mpmath.inf?

Yes, that was legal before.

> I would have thought that the right way to fix this is here by checking if M is infinite before doing ctx.prec += M:

That was my original fix.  The problem is that this will be required for many functions, that rely on hypercomb().  While with this approach we just fall deeper and rely on this function to handle special values for both mp and fp context.

BTW, to workaround SymPy failure we could use ctx.isinf instead.

skirpichev:
> In mpmath 1.3.0 it gives an error:

Oops, indeed.  That was the reason why infinities were handled incorrectly.

> Either way ignoring does not seem like the right thing to do.

Probably you are right.  But I dislike approach "lets add workarounds to every hypercomb-based function".  Maybe there are other options?

What if we make this *legal*?  The prec=0 is a special case for "exact" computations (without rounding).  See the ``mp._parse_prec()`` helper, it supports also ``ctx.inf``.  But now this is available only for kwargs of some mp functions.

oscarbenjamin:
Maybe there could be come sort of helper function for calling hypercomb with precision increased by `mag(z)`:
```python
def hypercomb_prec_magz(ctx, ..., z):
    magz = ctx.mag(z)
    if ctx.isfinite(magz):
        # for finite z we need magz extra precision in hypercomb
        with ctx.extraprec(magz):
            return ctx.hypercomb(..., z)
    else:
        # z is infinite so let hypercomb handle special values.
        # increased precision is not needed in this case.
        return ctx.hypercomb(..., z)
```
I don't know how many cases there are like that.

Alternatively:
```python
@contextmanager
def extraprec_inf(ctx, extra):
    """Increase precision if extra is finite."""
    if ctx.isfinite(extra):
        with ctx.extraprec(extra):
             yield
    else:
        # This branch is for handling special values where something is infinite
        # No extra precision is needed for that case.
        yield
```

skirpichev:
> I don't know how many cases there are like that.

```
$ git grep ctx.hypercomb mpmath/functions/|wc -l
57
```
But I think same story is valid for other functions (hyper/hyp2f1/etc).

> What if we make this legal? The prec=0 is a special case for "exact" computations (without rounding). See the mp._parse_prec() helper, it supports also ctx.inf.

Tests pass with
```diff
diff --git a/mpmath/ctx_mp_python.py b/mpmath/ctx_mp_python.py
index 335cd58..9d4f120 100644
--- a/mpmath/ctx_mp_python.py
+++ b/mpmath/ctx_mp_python.py
@@ -774,8 +774,8 @@ def default(ctx):
         ctx.trap_complex = False
 
     def _set_prec(ctx, n):
-        if not ctx.isfinite(n):
-            return
+        if ctx.isinf(n):
+            n = 0
         ctx._prec = ctx._prec_rounding[0] = max(1, int(n))
         ctx._dps = prec_to_dps(n)
 
```

skirpichev:
See #801

PR #803, opened 2024-05-05
Improve accuracy of log1p()

### LINKED ISSUE #790: Poor result in log1p(complex)
A case where lop1p of a complex number gets a poor real part:

```python
>>> import mpmath
>>> mpmath.__version__
'1.3.0'
>>> from mpmath import mp, mpc, log, log1p, workprec
>>> c = 1.8370676479640493e-39 - 4.6885882517313053e-20j
>>> complex(log1p(c))
(1.0991429897136409e-39-4.6885882517313053e-20j)
>>> with workprec(500): complex(log1p(c))
(2.93621063767769e-39-4.6885882517313053e-20j)
```

With the higher precision, the real part of the result is entirely different. I don't understand the code:

```python
>>> mp.mag(c), mp.prec
(-63, 53)
```

Because -63 < -53, I expected the code to take the branch that uses a 2-term expansion, which works fine:

```python
>>> c - 0.5*c**2 # matches the high-precision result
(2.93621063767769e-39-4.6885882517313053e-20j)
```

But I must not have found the correct source code, since editing it myself has no visible effect.

The other branch in the code calls `log()` instead after adding 1 to c with double precision:

```python
>>> complex(log(mp.fadd(c, 1, prec=106)))
(1.0991429897136409e-39-4.6885882517313053e-20j) # reproduces the "bad" result
```

It's not enough extra precision to help. In this case,

```python
>>> mp.mag((2*c.real + c.real**2 + c.imag**2))
-127
>>> complex(log(mp.fadd(c, 1, prec=127+53))) # so this is enough
(2.93621063767769e-39-4.6885882517313053e-20j)
>>> complex(log(mp.fadd(c, 1, prec=127+53-2))) # but 2 bits less not quite enough
(2.9362106376776874e-39-4.6885882517313053e-20j)
```

To be clear, "the code" I'm talking about is this, from functions/functions.py:

```python
@defun_wrapped
def log1p(ctx, x):
    if not x:
        return ctx.zero
    if ctx.mag(x) < -ctx.prec:
        return x - 0.5*x**2
    return ctx.log(ctx.fadd(1, x, prec=2*ctx.prec))
```

LATER: found the right source code. `ctx.mag(x)` and `-ctx.prec` in the above _both_ return -63 in this case, so that's why the "wrong" branch is taken. I have no idea how `ctx.prec` ended up at 63. As shown above, I never changed it from the default 53.

skirpichev:
Patch, ported from Tim Peters version at:
https://github.com/mpmath/mpmath/issues/790#issuecomment-2084094631

And few tests from the issue thread

Closes #790


mpmath/functions/functions.py
```diff
@@ -164,11 +164,80 @@ def expm1(ctx, x):
...
 @defun_wrapped
 def log1p(ctx, x):
-    if not x:
-        return ctx.zero
-    if ctx.mag(x) < -ctx.prec:
-        return x - 0.5*x**2
-    return ctx.log(ctx.fadd(1, x, prec=2*ctx.prec))
+    LOG1P_EXTRAPREC = 10  # ctx._wrap_specfun()
+    # Note that all cases could by handled by log(1+c) provided the
+    # add is done exactly. Our aim here is be much faster than that,
```
tim-one, line 169:
```suggestion
    # add is done exactly. Our aim here is to be much faster than that,
```
Adding a "to" missing from my original.

mpmath/functions/functions.py
```diff
@@ -164,11 +164,80 @@ def expm1(ctx, x):
...
+        # the imag part, and _usually_ the "a*a/2" in the real
+        # part. What remains is cheap to compute. In the real
+        # part, though, if `a` is negative, the remaining
+        # a+b**2/2 can suffer massive cancellation - even total.
+        real = a + b*b*0.5 # usually the real part of the result
+        if (a < 0.0
+            and ctx.mag(real) <= ctx.mag(a) - LOG1P_EXTRAPREC):
+            # The guard bits were lost to cancellation. Rare. At
+            # the contrived
+            # -1.999999873062092e-40+1.999999936531045e-20j
```
tim-one, line 224: Should add this input to the new tests - it's, by far, "the hardest" input I found.

skirpichev:
LGTM, I'm going to merge this in few days.

Issue #815, opened 2024-07-02
`mpmath/tests/test_convert.py::test_compatibility` fails with NumPy 2.0.0


mgorny:
FWICS numpy was pinned in 80fed12e34930679e179402975cb6dfab03edeb1 but I couldn't find any bug to track the test failure. FWICS only one test fails on Gentoo amd64:

```pytb
$ python -m pytest mpmath/tests/test_convert.py::test_compatibility 
========================================================= test session starts =========================================================
platform linux -- Python 3.10.14, pytest-8.2.2, pluggy-1.5.0
mpmath backend: python
mpmath mp class: <mpmath.ctx_mp.MPContext object at 0x7f22888dcee0>
mpmath version: 1.4.0a2.dev36+g8a4ce81
Python version: 3.10.14 (main, May 22 2024, 06:23:13) [GCC 14.1.1 20240516]
rootdir: /tmp/mpmath
configfile: pyproject.toml
plugins: xdist-3.6.1
collected 1 item                                                                                                                      

mpmath/tests/test_convert.py F                                                                                                  [100%]

============================================================== FAILURES ===============================================================
_________________________________________________________ test_compatibility __________________________________________________________

    def test_compatibility():
        np = pytest.importorskip("numpy")
        # numpy types
>       for nptype in np.core.numerictypes.typeDict.values():

mpmath/tests/test_convert.py:247: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
.venv/lib/python3.10/site-packages/numpy/core/__init__.py:31: in __getattr__
    _raise_warning(attr_name)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

attr = 'numerictypes', submodule = None

    def _raise_warning(attr: str, submodule: str = None) -> None:
        new_module = "numpy._core"
        old_module = "numpy.core"
        if submodule is not None:
            new_module = f"{new_module}.{submodule}"
[...]

PR #816, opened 2024-07-02
Fix the comparison in test_compatibility for NumPy 2.0.0

mgorny:
The comparison in `mpmath/tests/test_convert.py::test_compatibility` failed for `np.float16` in NumPy 2.0.0 since `2.0**-53` cannot be represented in half-precision floating point type.  Convert the LHS to `np.float64` to ensure that the comparison is done in sufficiently precise type.

This fixes the actual test failure from #815 but the deprecation warning remains.

### DIFF
--- a/mpmath/tests/test_convert.py
+++ b/mpmath/tests/test_convert.py
@@ -254,7 +254,10 @@ def test_compatibility():
         # Handle the weird types
         try: diff = np.abs(type(np.sqrt(x))(sqrt(x)) - np.sqrt(x))
         except: continue
-        assert diff < 2.0**-53
+        # we need to explicitly specify a sufficiently precise type
+        # for 2.0**-53, as otherwise numpy may convert it to LHS type
+        # that is not precise enough (e.g. float16)
+        assert diff < np.float64(2.0**-53)
     assert mpf(np.float64('inf')) == inf
     assert isnan(mp.npconvert(np.float64('nan')))
     if hasattr(np, "float128"):


mpmath/tests/test_convert.py
```diff
@@ -254,7 +254,9 @@ def test_compatibility():
         # Handle the weird types
         try: diff = np.abs(type(np.sqrt(x))(sqrt(x)) - np.sqrt(x))
         except: continue
-        assert diff < 2.0**-53
+        # numpy-2 does the comparison on LHS type, so we need to convert
+        # it to a type that can actually represent 2.0**-53
+        assert np.float64(diff) < 2.0**-53
```
skirpichev, line 259: Shouldn't you cast the second operand instead?
```suggestion
        assert diff < np.float64(2.0**-53)
```

> numpy-2 does the comparison on LHS type

Could you point to docs?  This is unusual wrt the CPython behaviour.

mgorny: > Shouldn't you cast the second operand instead?

Oh yes, that makes more sense. Done that.

> > numpy-2 does the comparison on LHS type
> 
> Could you point to docs? This is unusual wrt the CPython behaviour.

It's just my guess. Couldn't find anything specific in the changelog or docs. Do you want me to report it to numpy, in case it wasn't an intentional change?

skirpichev: > It's just my guess.

On which ground?

I would expect, that rich comparison will behave like it's specified in the Python language docs:
https://docs.python.org/3/reference/datamodel.html#object.__lt__

Else, numpy should document differences.

PS: please avoid force pushes.  Just add new commits.

mgorny: > On which ground?

Err, the results? Though I guess it's really more like "given numpy dtype and Python float, convert to numpy dtype".

skirpichev:
Thanks, I've cherry-picked your commit in #813

mgorny:
Thanks!

PR #819, opened 2024-07-19
Implement mpf.__format__

### LINKED ISSUE #337: Add mpf.__format__() dunder methods

javierelpianista:
Implemented the `__format__` method for the `_mpf` type.

Currently, `f`, `F`, `g`, `G`, `e`, and `E` formats are implemented, following the same specifications as specified for regular floats [here](https://docs.python.org/3/library/string.html#formatspec).
Only nearest rounding is implemented.

Partially resolves #337.


skirpichev (commented):
Very good start!

Please take look on comments.  Also, it seems the code is not fully covered by tests (see CI failure).

mpmath/tests/test_format.py
```diff
@@ -0,0 +1,465 @@
+import mpmath as mp
+
+def test_mpf_fmt_cpython():
```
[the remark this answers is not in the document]
skirpichev: I think you should test here that ``.format(mp.mpf)`` (with default precision) match ``.format(fp.mpf)``.  Maybe it's a good idea to use hypothesis here to get random input.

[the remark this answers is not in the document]
javierelpianista: I forgot to call `mp.mpf` inside the `format()` calls. Fixed that!

Also, I added a set of 1000 random tests with random format strings and random numbers.

mpmath/libmp/libfmt.py
```diff
@@ -0,0 +1,424 @@
...
+
+    # Check for rounding type
+    if pos < len(format_spec):
+        if format_spec[pos] in ('U', 'D', 'Y', 'Z', 'N'):
+            format_dict['rounding'] = format_spec[pos]
+            pos += 1
+
+    # Check for format type
+    if pos < len(format_spec):
+        if format_spec[pos] in ('f', 'F', 'g', 'G', 'e', 'E'):
```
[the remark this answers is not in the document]
javierelpianista: I mainly followed the format specification for Python floats, but in the future it would be nice to have these, as well as directed rounding.

mpmath/libmp/libfmt.py
```diff
@@ -0,0 +1,424 @@
...
+    Returns the rounded digits, and the number of places the decimal point was
+    shifted.
+    '''
+
+    exponent = 0
+
+    if rounding == round_nearest:
+        rnd_digs = stddigits[(base//2 + base % 2):base]
+    else:
+        raise ValueError('Rounding: {} not implemented yet.')
```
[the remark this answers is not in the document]
javierelpianista: I removed this part of the code since the `rounding` parameter is always set as the default `round_nearest`. Just kept line 15.

mpmath/libmp/__init__.py
```diff
@@ -66,3 +66,5 @@
                      mpi_mul, mpi_ne, mpi_neg, mpi_pos, mpi_pow, mpi_pow_int,
                      mpi_rgamma, mpi_sin, mpi_sqrt, mpi_str, mpi_sub, mpi_tan,
                      mpi_to_str)
+
+from .libfmt import (format_mpf)
```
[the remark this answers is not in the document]
javierelpianista: Moved the functions to libmpf.py.

mpmath/libmp/libfmt.py
```diff
@@ -0,0 +1,424 @@
...
+        lpad = (ntotal - nchars)//2
+        rpad = ntotal - nchars - lpad
+    else:
+        lpad = 0
+        rpad = 0
+
+    return (lpad, rpad)
+
+
+def read_format_spec(format_spec):
```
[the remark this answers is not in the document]
javierelpianista: I had copied over the code from `formatter_unicode.c` in CPython, but forgot the last section that checked this.
Anyways, I went ahead with your suggestion, the regex is a more elegant solution.
Also added a test so that `format(fp.mpf(1.1), "e ")` fails.

mpmath/tests/test_format.py
```diff
@@ -0,0 +1,552 @@
...
+    assert '{:.0f}'.format(mp.mpf(10.1)) == '10'
+    assert '{:.0f}'.format(mp.mpf(10.01)) == '10'
+    assert '{:.0f}'.format(mp.mpf(123.456)) == '123'
+    assert '{:.0f}'.format(mp.mpf(1234.56)) == '1235'
+    assert '{:.0f}'.format(mp.mpf(1e49)) == '9999999999999999464902769475481793196872414789632'
+    assert '{:.0f}'.format(mp.mpf(9.9999999999999987e+49)) == '99999999999999986860582406952576489172979654066176'
+    assert '{:.0f}'.format(mp.mpf(1e50)) == '100000000000000007629769841091887003294964970946560'
+
+    # precision 1
+    assert '{:.1f}'.format(mp.mpf(0.0001)) == '0.0'
```
skirpichev, line 118: You also could use formatted string literals in tests.

mpmath/tests/test_format.py
```diff
@@ -0,0 +1,554 @@
...
+    assert '{:+G}'.format(ninf) == '-INF'
+
+    assert '{:g}'.format(nan) == 'nan'
+    assert '{:+g}'.format(nan) == '+nan'
+    assert '{:G}'.format(nan) == 'NAN'
+    assert '{:+G}'.format(nan) == '+NAN'
+
+
+def test_errors():
+    with pytest.raises(Exception):
```
skirpichev, line 538: Hmm, shouldn't you test rather a specific exception instead?

mpmath/libmp/libmpf.py
```diff
@@ -1310,6 +1348,333 @@ def from_str(x, prec=0, rnd=round_fast, base=0):
...
+    elif num == fninf:
+        sign = '-'
+        digits = 'INF' if capitalize else 'inf'
+    elif num == fzero:
+        sign = '' if format_dict['sign'] == '-' else format_dict['sign']
+        digits = format_zero(
+                fmt_type=fmt_type,
+                precision=precision, capitalize=capitalize,
+                alternate=format_dict['alternate'],
+                thousands_separators=format_dict['thousands_separators'])
```
skirpichev, line 1622:
```suggestion
    # Special cases:
    if num in (fnan, finf, fninf):
        return format(format_spec, float(num))
    elif num == fzero:
        sign = '' if format_dict['sign'] == '-' else format_dict['sign']
        digits = format_zero(
                fmt_type=fmt_type,
                precision=precision, capitalize=capitalize,
                alternate=format_dict['alternate'],
                thousands_separators=format_dict['thousands_separators'])
```
Probably you could just fallback to python code here.  Maybe for zeros too.

mpmath/tests/test_str.py
```diff
@@ -13,11 +13,11 @@ def test_nstr():
                 [-0.0299195971, 0.205663228, 0.64453125e-20]])
     assert nstr(m, 4, min_fixed=-inf) == \
     '''[    0.75  0.1909                    -0.02992]
-[  0.1909  0.6563                      0.2057]
+[  0.1909  0.6562                      0.2057]
```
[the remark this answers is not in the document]
skirpichev: No, it's not wrong.  Test was correct.

mpmath/libmp/libmpf.py
```diff
@@ -1180,20 +1230,8 @@ def to_str(s, dps, strip_zeros=True, min_fixed=None, max_fixed=None,
...
-            if i >= 0:
-                digits = digits[:i] + stddigits[int(digits[i], base) + 1] + \
-                    '0' * (dps - i - 1)
-            else:
-                digits = '1' + '0' * (dps - 1)
-                exponent += 1
-        else:
-            digits = digits[:dps]
+        digits, exp_add = round_digits(digits, dps, base)
+        exponent += exp_add
```
skirpichev, line 1234: So, I think you should revert that, unless you want to implement also the round_up mode in round_digits().

mpmath/libmp/libmpf.py
skirpichev:
```suggestion
    # Special cases:
    if num in (fnan, finf, fninf, fzero, fnzero):
        return format(to_float(num), format_spec)
```
Yeah, that should work for all special numbers.

mpmath/libmp/libmpf.py
```diff
@@ -1310,6 +1348,333 @@ def from_str(x, prec=0, rnd=round_fast, base=0):
...
+    else:  # fmt_type == 'g'
+        if alternate:
+            return format_zero('f', precision-1, alternate, capitalize,
+                               thousands_separators)
+        else:
+            return '0'
+
+    return digits
+
+
```
skirpichev, line 1592: So, you won't need this:
```suggestion
```

mpmath/libmp/libmpf.py
skirpichev:
```suggestion
        if len(digits) > dps and digits[dps] in rnd_digs:
            digits = digits[:dps]
            i = dps - 1
            dig = stddigits[base-1]
            while i >= 0 and digits[i] == dig:
                i -= 1
            if i >= 0:
                digits = digits[:i] + stddigits[int(digits[i], base) + 1] + \
                    '0' * (dps - i - 1)
            else:
                digits = '1' + '0' * (dps - 1)
                exponent += 1
        else:
            digits = digits[:dps]
```

mpmath/tests/test_str.py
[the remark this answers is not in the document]
skirpichev:
```suggestion
[  0.1909  0.6563                      0.2057]
```

mpmath/tests/test_str.py
```diff
@@ -13,11 +13,11 @@ def test_nstr():
...
 [-0.02992  0.2057  0.000000000000000000006445]'''
     assert nstr(m, 4) == \
     '''[    0.75  0.1909   -0.02992]
-[  0.1909  0.6563     0.2057]
+[  0.1909  0.6562     0.2057]
```
skirpichev, line 20:
```suggestion
[  0.1909  0.6563     0.2057]
```

mpmath/tests/test_format.py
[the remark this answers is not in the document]
skirpichev: Ok, maybe it's too much for one pr;)  Lets do this later.

skirpichev (approved):
LGTM, with one comment on round_digits().  Please address this.

I took liberty to do some stylistic adjustments.  Please take look, lets be sure your are OK with this.

mpmath/libmp/libmpf.py
```diff
@@ -1101,6 +1102,49 @@ def to_digits_exp(s, dps, base=10):
...
+                round_up = True
+                break
+        if digits[dps-1] in stddigits[1:base:2]:
+            round_up = True
+
+    if not round_up:
+        digits = digits[:dps] + stddigits[int(digits[dps], base) - 1]
+
+    # Rounding up kills some instances of "...99999"
+    if len(digits) > dps and digits[dps] in rnd_digs:
```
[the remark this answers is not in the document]
javierelpianista: I am OK with the changes. The only small caveat is that now `to_str` rounds to nearest, ties away from zero whereas `format_mpf` rounds to nearest, ties to even. I am OK with this as well.

I removed the check for `len(digits)` because `digits` always comes from `to_digits_exp`, called with `dps = dps+3`, meaning that `digits` has always more than `dps` numbers.

In the case of the new `round_digits` function, since this check was on a line by itself, the branch where `len(digits) <= dps` was never covered by the tests, so I had to remove the line to pass the coverage test. I could not think of a way to test for this branch without resorting to calls to functions that are not meant to be used by the end user.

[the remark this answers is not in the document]
skirpichev: The rounding behaviour of to_str() seems to be a feature.  We might consider changing that, but it will costs a backward compatibility.

> I removed the check for len(digits) because digits always comes from to_digits_exp, called with dps = dps+3, meaning that digits has always more than dps numbers.

Maybe you should add a comment or assert.

mpmath/libmp/libmpf.py
```diff
@@ -1101,6 +1102,49 @@ def to_digits_exp(s, dps, base=10):
     exponent += len(digits) - fixdps - 1
     return sign, digits, exponent
 
+def round_digits(digits, dps, base):
+    '''
+    Returns the rounded digits, and the number of places the decimal point was
+    shifted.
+    '''
+
```
skirpichev, line 1110:
```suggestion

    assert len(digits) > dps
```

mpmath/libmp/libmpf.py
skirpichev, line 1129:
```suggestion
    if digits[dps] in rnd_digs:
```

skirpichev (changes requested):
Ah, you should test also subnormal values for floats.  Either in test_mpf_float() or in test_mpf_fmt_cpython().

mpf has no restrictions on exponent range, yet I think that output should here match Python's as well.

javierelpianista:
> Ah, you should test also subnormal values for floats. Either in test_mpf_float() or in test_mpf_fmt_cpython().
> 
> mpf has no restrictions on exponent range, yet I think that output should here match Python's as well.

If I understand you correctly, you mean mpf initialized from floats? Such as 
```python
num = 2e-315
print('{:22.15f}'.format(num))
print('{:22.15f}'.format(mpf(num)))
print('{:22.15g}'.format(num))
print('{:22.15g}'.format(mpf(num)))
print('{:22.15e}'.format(num))
print('{:22.15e}'.format(mpf(num)))
```
resulting in
```
     0.000000000000000
     0.000000000000000
 2.00000000190402e-315
 2.00000000190402e-315
2.000000001904024e-315
2.000000001904024e-315
```
But I would expect
```python
num_str = '2e-315'
print('{:22.15f}'.format(mpf(num_str)))
print('{:22.15g}'.format(mpf(num_str)))
print('{:22.15e}'.format(mpf(num_str)))
```
to result into 
```
     0.000000000000000
                2e-315
2.000000000000000e-315
```
because in this case the mpf is correctly initialized.

skirpichev:
> If I understand you correctly, you mean mpf initialized from floats?

Yes, as in other your compatibility tests.  And with subnormal values, like in shown example.

> But I would expect [...] because in this case the mpf is correctly initialized.

Nothing incorrect in using mpf(float) for default settings.

mpmath/tests/test_format.py
```diff
@@ -0,0 +1,577 @@
...
+        # We skip the case when num == 0 since mpmath does not have negative
+        # zero.
+        if num != 0 and random.randint(0, 1):
+            num = -num
+        assert fmt_str.format(num) == fmt_str.format(mp.mpf(num))
+
+
+def test_mpf_fmt():
+    '''
+    These tests are either specific tests to mpf, or tests that cover
```
skirpichev, line 428:
```suggestion
    These tests are either specific tests to mpf, or tests that cover
```

skirpichev:
Formatting support must be documented, probably here: https://mpmath.readthedocs.io/en/latest/general.html#conversion-and-printing

But maybe we can do this later in a separate pr.

mpmath/libmp/libmpf.py
```diff
@@ -1311,6 +1356,299 @@ def from_str(x, prec=0, rnd=round_fast, base=0):
     return s
 
 
+#----------------------------------------------------------------------------#
+#                              String formatting                             #
+#----------------------------------------------------------------------------#
+
+blog2_10 = 3.3219280948873626
```
skirpichev, line 1363: This constant should be moved up (near math_float_inf, I think) and reused in a couple of places:
```
$ git grep 3.321928 mpmath/libmp/libmpf.py
mpmath/libmp/libmpf.py:    return max(1, int(round(int(n)/3.3219280948873626)-1))
mpmath/libmp/libmpf.py:    return max(1, int(round((int(n)+1)*3.3219280948873626)))
mpmath/libmp/libmpf.py:        prec = int(len(x.as_tuple()[1])*3.3219280948873626)
mpmath/libmp/libmpf.py:        blog2 = 3.3219280948873626
```

mpmath/libmp/libmpf.py
```diff
@@ -1311,6 +1356,299 @@ def from_str(x, prec=0, rnd=round_fast, base=0):
...
+    if exponent + precision < -1:
+        if precision > 0:
+            digits = '0.' + precision*'0'
+        else:
+            digits = '0'
+
+        if no_neg_0:
+            sign = '' if sign_spec == '-' else sign_spec
+    else:
+        # Rounding up kills some instances of "...99999"
```
skirpichev, line 1494:
```suggestion
```

skirpichev (approved):
LGTM with few comments

mpmath/tests/test_format.py
```diff
@@ -0,0 +1,577 @@
...
+                    )
+                )
+
+        assert fmt_str.format(fp.mpf(num)) == fmt_str.format(mp.mpf(num))
+
+    # These additional tests test for mpf initialized form subnormal floats
+    for _ in range(10000):
+        fmt_str = random_fmt()
+        num = fp.mpf('{:.15f}e{:d}'.format(
+                     random.uniform(1, 10), random.randint(-324, -308)))
```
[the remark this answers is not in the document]
javierelpianista: If I use something like `random.uniform(0, 1e-308)`, it will yield most numbers with exponent between -308 and -309. This way every subnormal exponent has the same probability to appear.

skirpichev:
@javierelpianista, thank you!

skirpichev:
@javierelpianista, are you planing to work on mpc.__format__?

javierelpianista:
> @javierelpianista, are you planing to work on `mpc.__format__`?

I could give it a try in the future, but not right now.

PR #826, opened 2024-08-01
mpf.__round__() returns mpf

### LINKED ISSUE #455: rounding in Python 3 need not return a `float`
This is based on a [discussion in SymPy](https://github.com/sympy/sympy/issues/12056)

I think `round` should return an mpf rather then a `float`.  Consider:
```python
>>> mp.dps = 64
>>> x = mpmath.mpf('0.66666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666')

>>>  round(x, 48)      # is not going to give 48 digits
0.6666666666666666

>>> repr(_)            # its a float
'0.6666666666666666'
```
The current implementation of `__round__` just converts to `float` and rounds there.  I think we can do better---and are allowed to in Python 3.

skirpichev:
Probably this will look non-intuitive, but underlying floating-point arithmetic is binary, not decimal.  Thus, round(x, n) sometimes will pick up a decimal representation with more than n digits.  This is something common for CPython < 2.7 and < 3.1:
```pycon
Python 2.6.6 (r266:84292, Aug 12 2014, 07:57:07) [GCC 4.4.5] on linux2
>>> round(0.42754287856598971, 2)
0.42999999999999999
```
And with this patch as well:
```pycon
Python 3.12.5+ (heads/3.12:0181aa2e3e, Aug 29 2024, 14:55:08) [GCC 12.2.0] on linux
>>> from mpmath import *
>>> round(mp.mpf(0.42754287856598971), 2)
mpf('0.42999999999999999')
>>> mp.pretty = True
>>> round(mp.mpf(0.42754287856598971), 2)
0.43
```

See also https://github.com/python/cpython/issues/45921

Closes #455

### DIFF
--- a/mpmath/ctx_mp_python.py
+++ b/mpmath/ctx_mp_python.py
@@ -412,8 +412,13 @@ class _mpf(mpnumeric):
     def to_fixed(self, prec):
         return to_fixed(self._mpf_, prec)
 
-    def __round__(self, *args):
-        return round(float(self), *args)
+    def __round__(self, ndigits=0):
+        ctx = self.context
+        if ctx.isfinite(self):
+            frac = MPQ(*self.as_integer_ratio())
+            res = round(frac, ndigits)
+            return ctx.convert(res)
+        return self
 
 
 class _constant(_mpf):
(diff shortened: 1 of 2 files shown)


skirpichev:
I'm going to merge this after a week.  But I would appreciate @cbm755 feedback, as new feature might look surprising.

PR #830, opened 2024-08-02
Support gmpy2-like rounding modes in to_str()

### LINKED ISSUE #757: Implement directed rounding in to_str()
See #109.  Eventually, this could be used to support [gmpy2-like rounding modes](https://gmpy2.readthedocs.io/en/latest/mpfr.html#gmpy2.mpfr.__format__) in ``mpf/mpc.__format__()``.

javierelpianista:
Closes #757.


mpmath/libmp/libmpf.py
```diff
@@ -33,7 +33,8 @@ class ComplexResult(ValueError):
     pass
 
 # All supported rounding modes
-round_nearest = sys.intern('n')
+round_nearest = sys.intern('n')  # Round to nearest with tying away from zero
+round_nearest_even = sys.intern('m')  # Round to nearest with tying to even
```
skirpichev, line 37: No, it's wrong.  Currently, round_nearest is exactly "round to nearest with ties to even" (like roundTiesToEven from the IEEE 754 standard).  See e.g. round_int().

It seems, you don't add any code, that supports new meaning (outside of to_str()).  Anyway, this is a backward compatibility break, which I won't introduce.

mpmath/libmp/libmpf.py
```diff
@@ -1386,15 +1402,17 @@ def from_str(x, prec=0, rnd=round_fast, base=0):
...
     (?:\.(?P<precision>0|[1-9][0-9]*))?
-    (?P<rounding>[UDZN])?
+    (?P<rounding>[UDYZMN])?
     (?P<type>[eEfFgG])
 """, re.DOTALL | re.VERBOSE).fullmatch
 
 _GMPY_ROUND_CHAR_DICT = {
         'U': round_ceiling,
         'D': round_floor,
+        'Y': round_up,
```
skirpichev, line 1412: Better split out this to a separate patch.

mpmath/libmp/libmpf.py
```diff
@@ -1157,9 +1163,23 @@ def round_digits(digits, dps, base, rounding=round_nearest):
...
+    elif rounding == round_floor:
+        return round_up if sign else round_down
+    else:
+        return rounding
+
 
 def to_str(s, dps, strip_zeros=True, min_fixed=None, max_fixed=None,
-           show_zero_exponent=False, base=10, binary_exp=False):
+           show_zero_exponent=False, base=10, binary_exp=False,
+           rounding=round_nearest):
```
skirpichev, line 1182: So, lets not add new rounding modes.

Instead this argument should be set to some sentinel value (None, perhaps) per default.  With this you should emit DeprecationWarning and fallback on legacy code.  Else - use specified rounding mode.

skirpichev (changes requested):
Thus,
* ~~please add 'Y' support in a separate pr~~; done: see #831
* please don't introduce new rounding mode
* rounding=None in to_str() will be current (legacy) behaviour, that should be deprecated
* deprecation message should announce that new default will be round_nearest

mpmath/libmp/libmpf.py
javierelpianista: > No, it's wrong. Currently, round_nearest is exactly "round to nearest with ties to even" (like roundTiesToEven from the IEEE 754 standard). See e.g. round_int().

I incorrectly assumed that the default in mpmath was tie away from zero because that's what `to_str` does:
```python
>>> nstr(mp.mpf('0.25'), 1)
'0.3'
>>> nstr(mp.mpf('0.0625'), 2)
'0.063'
```

> It seems, you don't add any code, that supports new meaning (outside of to_str()). Anyway, this is a backward compatibility break, which I won't introduce.

I did, please see lines 1128 -- 1141 of libmpf.py (also the new tests in test_format.py).

mpmath/libmp/libmpf.py
javierelpianista: OK!

mpmath/libmp/libmpf.py
javierelpianista: So `to_str` should tie to even instead of away from zero?
We could also do something like
```python
round_nearest = sys.intern('n')  # Round to nearest with tying to even
round_nearest_az = sys.intern('m')  # Round to nearest with tying away from zero
```
and set the default value for `rounding` in `to_str` as `round_nearest_az`. This would keep compatibility with previous versions and not require to deprecate the old code.

mpmath/libmp/libmpf.py
skirpichev: > I incorrectly assumed that the default in mpmath was tie away from zero because that's what to_str does

Mpmath uses binary arithmetic.  Here defaults for the mp context:
```pycon
>>> from mpmath import mp
>>> mp._prec_rounding
[53, 'n']
```

to_str() for decimal output, probably, has historically different defaults.

> I did, please see lines 1128 -- 1141

round_digits() used only for formatting.

mpmath/libmp/libmpf.py
skirpichev: > So to_str should tie to even instead of away from zero?

No.  I think that to_str() should keep current behaviour, but with a deprecation notice.  If user supply a different rounding value (not None) - it should be used.

I don't think that we should add support for a new rounding mode beyond MPFR.

Also, using different defaults in the mp context and for decimal output looks odd.

javierelpianista:
@skirpichev The DeprecationWarning makes most test cases fail. 

To fix this I should modify all calls to `to_str` to include correct rounding.
This includes the `__str__` and `__repr__` methods for each type defined in mpmath.
I've made some of these modifications for mpf and mpc types, would you please tell me if you think this is the correct way to go forward?

skirpichev (commented):
> The DeprecationWarning makes most test cases fail.

It's easy to filter out this warning.  My second commit shows how to do this.

> To fix this I should modify all calls to to_str to include correct rounding.

Huh, tough decision.  It seems that we in any case will introduce a compatibility break.  Or - at least - a DeprecationWarning, which might surprise user, as it will come from the mpmath code, that calls to_str().

On another hand, removing legacy code and setting new defaults - leave us with just one test failure in test_str.py.  See https://github.com/mpmath/mpmath/pull/832. I'll run diofant's test suite on top of that pr, see https://github.com/diofant/diofant/pull/1430.

I'll appreciate @oscarbenjamin feedback on this.

> This includes the __str__ and __repr__ methods for each type defined in mpmath.

Hmm, maybe we won't such flexibility?  E.g. gmpy2 uses rounding to nearest, regardless of the context settings.  (But see https://github.com/aleaxit/gmpy/issues/503)

mpmath/libmp/libmpf.py
```diff
@@ -1157,9 +1161,23 @@ def round_digits(digits, dps, base, rounding=round_nearest):
     return digits, exponent
 
 
+def abs_rounding(rounding, sign):
```
skirpichev, line 1164: maybe this belongs to the round_digits?

skirpichev:
Now this has merge conflicts.

mpmath/libmp/libmpf.py
javierelpianista: Maybe, this would add a parameter to round_digits, but could make the code clearer.

skirpichev (commented):
This now lacks tests (don't forgot to test the warning too).

But I'm not sure if we should pursue this approach.  https://github.com/mpmath/mpmath/pull/832 seems to be working for the Diofant (not test failures).  I would appreciate testing on SymPy, however.

Lets wait few days for feedback from SymPy people.  If no objections follows - I think we should just switch to ``round_nearest``.

skirpichev (changes requested):
yeah, lest just switch to round_nearest per default

javierelpianista:
> yeah, lest just switch to round_nearest per default

OK, so if I understand correctly, I should modify all calls to `to_str` in the `__str__` and `__repr__` methods to use `round_nearest`, correct? Otherwise we would get deprecation warnings for simply printing a float, which is something regular users would not like (myself included)

skirpichev:
> OK, so if I understand correctly, I should modify all calls to to_str in the __str__ and __repr__ methods to use round_nearest, correct?

No.  Lets just change default in to_str() to round_nearest.

mpmath/libmp/libmpf.py
```diff
@@ -1207,6 +1215,11 @@ def to_str(s, dps, strip_zeros=True, min_fixed=None, max_fixed=None,
         if base not in (2, 16):
             raise ValueError("binary_exp option could be used for base 2 and 16")
 
+    if rounding is None:
+        warnings.warn("Calling to_str without specifying rounding mode is "
+                      "deprecated. The new default rounding mode will be "
+                  "round_nearest.", DeprecationWarning)
+
```
skirpichev, line 1222: No, lets just drop this:
```suggestion
```

skirpichev (changes requested):
Again.  Please just drop legacy code (rounding=None).  Don't test rounding=None.

javierelpianista:
@skirpichev I think we should assert inside `to_str` that `rounding` is one of the correct modes.

pyproject.toml
```diff
@@ -55,7 +55,8 @@ testpaths = ['mpmath', 'docs']
 doctest_optionflags = ['IGNORE_EXCEPTION_DETAIL', 'ELLIPSIS']
 addopts = "--durations=20 --doctest-modules --doctest-glob='*.rst'"
 norecursedirs = ['docs/plots', 'demo', '.eggs', '.git']
-filterwarnings = ['error::DeprecationWarning']
+filterwarnings = ['error::DeprecationWarning',
+                  'ignore:Calling to_str without specifying rounding mode:DeprecationWarning:mpmath']
```
skirpichev, line 59:
```suggestion
filterwarnings = ['error::DeprecationWarning']
```

skirpichev (changes requested):
> we should assert inside to_str that rounding is one of the correct modes.

No.  This is a public function and we should raise ValueError.

Please add basic tests for new option.

skirpichev:
Hmm, it seems some test(s?) loop forever with your recent changes.

javierelpianista:
> I don't think that just removing "bad test" is ok. Why it's failing?

The test I removed aimed to cover line 1237 of libmpf.py. It wasn't relevant for this PR so I removed it without much thought.

skirpichev:
I can reproduce this failure locally with python backend (MPMATH_NOGMPY=1).  Maybe it's a normalize() bug.

skirpichev (changes requested):
LGTM, but please update pr title/description

skirpichev (approved):

skirpichev:
Thanks.

PR #834, opened 2024-08-04
Correct formatting for `round_up` rounding mode

javierelpianista:
This PR improves the code in `round_digits`: now the first digit in the `stddigits` list is excluded from the list of digits that must be rounded up when the `round_up` rounding mode is used. When rounding to a 0 digit, all the digits following it are looked up, and if one of them is different from 0, the 0 is rounded up.

This fixes some wrongfully formatted floats as, for example, when running the following code in the master branch
```python
>>> from gmpy2 import mpfr
>>> from mpmath import mp
>>> print('{:.2Uf}'.format(mpfr('0.25')))
0.25
>>> print('{:.2Uf}'.format(mp.mpf('0.25')))
0.26
>>>
```

### DIFF
--- a/mpmath/tests/test_format.py
+++ b/mpmath/tests/test_format.py
@@ -645,33 +645,69 @@ def test_mpf_fmt():
         assert f"{mp.mpf('-123.456'):.5Ng}" == "-123.46"
 
         num = mp.mpf('0.1')
-        assert f"{-num:=.2Df}" == "-0.10"
-        assert f"{-num:=.3Df}" == "-0.100"
-        assert f"{-num:=.4Df}" == "-0.1000"
-        assert f"{-num:=.5Df}" == "-0.10000"
-        assert f"{-num:=.6Df}" == "-0.100000"
-        assert f"{-num:=.7Df}" == "-0.1000000"
+        assert f"{-num:=.2Df}" == "-0.11"
+        assert f"{-num:=.3Df}" == "-0.101"
+        assert f"{-num:=.4Df}" == "-0.1001"
+        assert f"{-num:=.5Df}" == "-0.10001"
+        assert f"{-num:=.6Df}" == "-0.100001"
+        assert f"{-num:=.7Df}" == "-0.1000001"
 
-        assert f"{-num:=.2De}" == "-1.00e-01"
-        assert f"{-num:=.3De}" == "-1.000e-01"
-        assert f"{-num:=.4De}" == "-1.0000e-01"
-        assert f"{-num:=.5De}" == "-1.00000e-01"
-        assert f"{-num:=.6De}" == "-1.000000e-01"
-        assert f"{-num:=.7De}" == "-1.0000000e-01"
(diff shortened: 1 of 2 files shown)


skirpichev (changes requested):
You should include tests.

mpmath/tests/test_format.py
```diff
@@ -644,6 +644,34 @@ def test_mpf_fmt():
         assert f"{mp.mpf('-123.456'):.5Zg}" == "-123.45"
         assert f"{mp.mpf('-123.456'):.5Ng}" == "-123.46"
 
+        num = mp.mpf('0.1')
+        assert f"{-num:=.2Df}" == "-0.10"
```
skirpichev, line 648: Are you sure?  Looks incorrect for me, c.f.:
```pycon
>>> import gmpy2
>>> num = gmpy2.mpfr('0.1')
>>> f"{-num:.2Df}"
'-0.11'
```

javierelpianista: I am sure about rounding 0 down. What you show me is because of a different issue:
```python
>>> from gmpy2 import mpfr
>>> print('{:.30f}'.format(mpfr(0.1)))
0.10000000000000005551115123126
>>> print('{:.2Uf}'.format(mpfr(0.1)))
0.11
```
Here 0.1 is rounded up because the incomplete floating point representation is slightly higher than 0.1.
On the other hand, if you format an exact float,
```python
>>> print('{:.30f}'.format(mpfr(0.25)))
0.250000000000000000000000000000
>>> print('{:.3Uf}'.format(mpfr(0.25)))
0.250
```
the 0 gets rounded down.

If I do the same with mpmath, leaving the code as it was before, I get
```python
>>> print('{:.2Uf}'.format(mp.mpf('0.1')))
0.11
>>> print('{:.3Uf}'.format(mp.mpf('0.25')))
0.251
```
whereas with the new one,
```python
>>> from mpmath import mp
>>> print('{:.2Uf}'.format(mp.mpf('0.1')))
0.10
>>> print('{:.3Uf}'.format(mp.mpf('0.25')))
0.250
```

The way I coded `round_digits()`, it only looks at the digit that is next to the one we round to. I should do something similar to what I did for tying `round_nearest` to even. The incorrect rounding up of the digit 0  was hiding this other problem.

skirpichev: > Here 0.1 is rounded up because the incomplete floating point representation is slightly higher than 0.1.

What you mean by "incomplete"?  Your rounding algorithm should work with binary floating-point numbers, not only with whose decimal numbers, that are binary fractions, like 0.25.

> If I do the same with mpmath, leaving the code as it was before, I get

Unfortunately, this fix breaks other tests, shown above.  We can't include wrong tests in CI.

> I should do something similar to what I did for tying round_nearest to even.

That sounds much better.

Yes, you should examine all digits after dps, not just assume they are ``0``.

skirpichev (changes requested):
Could you, please, adjust PR title and description?

skirpichev (approved):

PR #841, opened 2024-08-17
Implement 'a'/'A' formating types for mpf.__format__

skirpichev:
See #337

### DIFF
--- a/mpmath/ctx_mp_python.py
+++ b/mpmath/ctx_mp_python.py
@@ -398,6 +398,22 @@ class _mpf(mpnumeric):
             '-1.23456'
             >>> f'{x:.5Df}'
             '-1.23457'
+
+        Format types 'a' and 'A' (use uppercase digits) allow to represent
+        floating-point number as a C99-style hexadecimal string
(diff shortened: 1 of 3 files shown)


skirpichev:
See also https://github.com/python/cpython/issues/113804.  Potentially, we could reuse 'x'/'X' format codes, but I think it's a good idea to be compatible with gmpy2 (and C) here.

@javierelpianista ?

javierelpianista:
I

> See also [python/cpython#113804](https://github.com/python/cpython/issues/113804). Potentially, we could reuse 'x'/'X' format codes, but I think it's a good idea to be compatible with gmpy2 (and C) here.
> 
> @javierelpianista ?

Yes, I think it's better to use 'a' and 'A'.

skirpichev:
> print(f"{gmpy2.mpfr('1.234567890123456789'):.0a}") # This one is different

It's not.  The mpmath always use 1-normalization for normal numbers, while in the MPFR - the digit before decimal dot should be just non-zero.  This is something they wish to fix, but...  See https://sympa.inria.fr/sympa/arc/mpfr/2021-05/msg00002.html

PR #857, opened 2024-08-31
Fix formatting of mpf's when format type is missing

### LINKED ISSUE #856: Formatting mpf's is incompatible with Python if formatting type is missing
Documentation [says](https://docs.python.org/3/library/string.html#format-specification-mini-language): "For [float](https://docs.python.org/3/library/functions.html#float) this is the same as 'g', except that when fixed-point notation is used to format the result, it always includes at least one digit past the decimal point. The precision used is as large as needed to represent the given value faithfully."

While we have:
```pycon
>>> f'{mp.mpf(1.23456789):}'  # fixed precision=6
'1.23457'
>>> f'{mp.mpf(1):}'  # no trailing 0
'1'
```

Note, that it might be non-trivial to match exactly Python's output, because e.g. CPython and mpmath use different decimal approximations for some floats.  E.g.:
```pycon
>>> mp.mpf(0.3976135751111121)
mpf('0.39761357511111212')
>>> 0.3976135751111121
0.3976135751111121
>>> 0.39761357511111212
0.3976135751111121
```

skirpichev:
In this case it was incorrectly assumed, that this is an equivalent of 'g' formatting type.  It's not:
1) when no precision is specified - it should be equal to repr_deps(prec) to represent the given value faithfully.
2) when fixed-point is used to format the result, it always includes at least one digit past the decimal point.

Closes #856

### DIFF
--- a/mpmath/libmp/libmpf.py
+++ b/mpmath/libmp/libmpf.py
@@ -1450,9 +1450,9 @@ def read_format_spec(format_spec):
         'alternate': False,
         'thousands_separators': '',
         'width': -1,
-        'precision': 6,
+        'precision': -1,
         'rounding': round_nearest,
-        'type': 'g'
+        'type': ''
         }
 
     if match := _FLOAT_FORMAT_SPECIFICATION_MATCHER(format_spec):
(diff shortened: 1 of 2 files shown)


skirpichev:
Now on top of #859

skirpichev:
See also https://github.com/python/cpython/issues/123560

PR #859, opened 2024-09-01
Add hypothesis-based formatting tests and workaround #858

skirpichev:
* fix also few other minor issues, coming from tests, closes #858
* simplify handling '%' format type
* split out format_digits() and redo handling special numbers

PR #860, opened 2024-09-01
Other fix 858

javierelpianista:
This PR improves upon the fixes in #859 for fixed formatting of numbers like 0.99, 0.999, etc.

### DIFF
--- a/mpmath/libmp/libmpf.py
+++ b/mpmath/libmp/libmpf.py
@@ -1493,8 +1493,6 @@ def format_fixed(s,
                  base=10,
                  alternate=False,
                  no_neg_0=False,
-                 percent=False,
-                 prec=0,
                  rounding=round_nearest):
     '''
     Format a number into fixed point.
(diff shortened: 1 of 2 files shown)


mpmath/libmp/libmpf.py
```diff
@@ -1098,7 +1098,7 @@ def to_digits_exp(s, dps, base=10):
     exponent += len(digits) - fixdps - 1
     return sign, digits, exponent
 
-def round_digits(sign, digits, dps, base, rounding=round_nearest):
+def round_digits(sign, digits, dps, base, rounding=round_nearest, fixed=False):
```
skirpichev, line 1101: It seems to be the key.  Now I think that the fix should go to "Rounding up kills some instances" block and could be one-liner.

skirpichev:
https://github.com/mpmath/mpmath/pull/859 merged with fix from this pr.

PR #889, opened 2024-12-05
Use parity formula for besseli

### LINKED ISSUE #877: Negative integer order of the BesselI function
mpmath version: 1.3.0

I am using mpmath to generate values for verification of my own complex Bessel function calculation routine.
https://github.com/tk-yoshimura/ComplexBessel

I have discovered that mpmath cannot compute for large negative integers in Bessel I functions, even though it can use the inversion formula.
https://dlmf.nist.gov/10.27

Example:

```py
import mpmath as mpf

mpf.mp.dps = 64

x = 2

nus = [31, 32, 63, 64, 127, 128,
      -31, -31.25, -32, -32.25, 
      -63, -63.25, -64, -64.25, 
      -127, -127.25, -128, -128.25
]

for nu in nus:
    print(f"nu = {nu}, x = {x}")
    
    try:
        y = mpf.besseli(nu, x)
        print(f"{mpf.nstr(y, 40, strip_zeros=False, min_fixed=2, max_fixed=1)}")
    except ValueError as e:
        print(e)

    print("\n")
```

Error Message: 
```
nu = -127, x = 2

hypercomb() failed to converge to the requested 216 bits of accuracy
using a working precision of 4814 bits. The function value may be zero or
infinite; try passing zeroprec=N or infprec=M to bound finite values between
2^(-N) and 2^M. Otherwise try a higher maxprec or maxterms.
```

skirpichev:
See
https://functions.wolfram.com/Bessel-TypeFunctions/BesselI/04/02/01/0002/

Closes #877


mpmath/functions/bessel.py
```diff
@@ -82,6 +82,8 @@ def h(n):
 def besseli(ctx, n, z, derivative=0, **kwargs):
     n = ctx.convert(n)
     z = ctx.convert(z)
+    if ctx.nint(n) and n < 0:
+        return ctx.besseli(-n, z, derivative, **kwargs)
```
oscarbenjamin, line 86: This fails for non-real `n`:
```python
In [1]: import mpmath

In [2]: mpmath.besseli(1j, 1)
TypeError                                 Traceback (most recent call last)
Cell In[2], line 1
----> 1 mpmath.besseli(1j, 1)

File mpmath/mpmath/functions/bessel.py:85, in besseli(ctx, n, z, derivative, **kwargs)
     83 n = ctx.convert(n)
     84 z = ctx.convert(z)
---> 85 if ctx.nint(n) and n < 0:
     86     return ctx.besseli(-n, z, derivative, **kwargs)
     87 if not z:

TypeError: '<' not supported between instances of 'mpc' and 'int'
```

skirpichev: Thanks, should be fixed by 294e913.

oscarbenjamin: I think it also needs to be changed to `isint` e.g.:
```python
In [2]: mpmath.besseli(-1.5, 1)
Out[2]: mpf('0.2935253263474798')

In [8]: scipy.special.iv(-1.5, 1)
Out[8]: np.float64(-0.2935253263474799)
```

PR #894, opened 2024-12-31
Support negative indexes in matrix

### LINKED ISSUE #835: Matrix doesn't support negative indices
```python
from mpmath import matrix

# Create a 3x3 matrix
A = matrix([[1, 2, 3], [4, 5, 6], [7, 8, 9]])

# Access the last element of the first row
print(A[0, -1])
```

I expect negative index `-n` to indicate last n element, like in Python list or numpy array. But I got `0`.

After some digging, I found the related code:
https://github.com/mpmath/mpmath/blob/cf98bcf07d7d6ef8ebfc0575276107aac46519b9/mpmath/matrices/matrices.py#L499-L505

The same goes for index of columns, such as `A[-1,:]` returns `[0.0  0.0  0.0]`.

Are negative indices for other purposes? Can Python-like negative indices be supported?

ricor07:
closes #835

### DIFF
--- a/mpmath/matrices/matrices.py
+++ b/mpmath/matrices/matrices.py
@@ -107,6 +107,14 @@ class _matrix:
         mpf('2.0')
         >>> x[1,0]
         mpf('2.0')
+    
+    It is also possible to access matrices and vectors via negative index:
+        >>> x = matrix([[1, 2], [3, 4]])
+        >>> y = matrix([6, 7])
+        >>> x[-1, -2]
+        mpf('3.0')
+        >>> y[-2]
+        mpf('6.0')
 
     Other
     .....
@@ -467,7 +475,11 @@ class _matrix:
                 # Single row
                 if key[0] >= self._rows:
                     raise IndexError('Row index out of bounds')
+                
                 rows = [key[0]]
+                
+                if(key[0] < 0 and key[0] > - self._rows + 1):
+                    rows[0] += self._rows
 
             # Columns
             if isinstance(key[1],slice):
@@ -483,7 +495,11 @@ class _matrix:
                 # Single column
                 if key[1] >= self._cols:
                     raise IndexError('Column index out of bounds')
+                
                 columns = [key[1]]
+
+                if(key[1] < 0 and key[1] > - self._cols + 1):
+                    columns[0] += self._cols
 
             # Create matrix slice
             m = self.ctx.matrix(len(rows),len(columns))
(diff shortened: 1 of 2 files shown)


skirpichev:
What about setting?  This also lacks tests.

> why if key is not in self.data the output is 0 instead of raising an error.

perhaps, because it's a sparse matrix?

ricor07 (commented):
Added tests; fixed a syntax error present in my previous implementation

mpmath/matrices/matrices.py
```diff
@@ -496,6 +496,28 @@ def __getitem__(self, key):
             return m
 
         else:
+            """
+            Just as conventional numpy arrays, it is possible to select an element out of the matrix via negative index:
+            A = matrix([[1,2,3],[4,5,6],[7,8,9]])
+            print(A[0, -1]) # 3
+            print(A[-1, -2]) # 8
+            """
```
skirpichev, line 504:
```suggestion
```
Rather you can add examples in docstring.

mpmath/matrices/matrices.py
```diff
@@ -496,6 +496,28 @@ def __getitem__(self, key):
...
+            row, col = key  # Unpack the tuple
```
skirpichev, line 505:
```suggestion
            row, col = key
```

mpmath/matrices/matrices.py
```diff
@@ -496,6 +496,28 @@ def __getitem__(self, key):
...
+                else:
+                    raise IndexError('matrix index out of range')
+
+            if col < 0:
+                if -col <= self._cols:
+                    col = self._cols + col
+                else:
+                    raise IndexError('matrix index out of range')
+
+            key = (row, col)  # Repack into a new tuple
```
skirpichev, line 519:
```suggestion
            key = row, col
```

mpmath/tests/test_matrices.py
```diff
@@ -1,11 +1,16 @@
 import sys
-
 import pytest
 
 from mpmath import (convert, diag, extend, eye, fp, hilbert, inf, inverse, iv,
                     j, matrix, mnorm, mp, mpc, mpf, mpi, norm, nstr, ones,
                     randmatrix, sqrt, swap_row, zeros)
 
+def test_matrix_indexes():
```
skirpichev, line 8: Rather you could add this to test_matrix_basic().

skirpichev (commented):
Few comments.

You also should adjust `__setitem__()`.

ricor07:
Hello, I implemented the changes you asked for. Maybe there are a few comments that can be added but I'd like to hear your opinion. Best regards

skirpichev (commented):
Looks good, few notes added.

But please add tests, including all error cases.  See coverage reports in CI.

skirpichev:
Ah, you did your pr from the master branch of your fork.  So, I can't push changes to your repo.

Once this will be finished (please address review comments, fix formatting issues and add coverage tests), you should remove your fork and recreate it back (if you are planning to do something else for mpmath;)).  And next time please first create a branch for your patch.  Not use the master branch for this.

skirpichev:
Thanks, continued in #897

ricor07:
Ok, thanks for your job.

PR #900, opened 2025-01-19
Better introspection support for decorated functions

### LINKED ISSUE #899: Better introspection support for decorated functions
Currently we have something like this:
```pycon
>>> from mpmath import *
>>> import inspect
>>> help(polylog)
Help on method f_wrapped in module mpmath.ctx_mp_python:

f_wrapped(*args, **kwargs) method of mpmath.ctx_mp.MPContext instance
    Computes the polylogarithm, defined by the sum

    .. math ::

        \mathrm{Li}_s(z) = \sum_{k=1}^{\infty} \frac{z^k}{k^s}.

    This series is convergent only for `|z| < 1`, so elsewhere
    the analytic continuation is implied.

    The polylogarithm should not be confused with the logarithmic
    integral (also denoted by Li or li), which is implemented
    as :func:`~mpmath.li`.
[...]
>>> inspect.signature(polylog)
<Signature (*args, **kwargs)>
```
You can only guess in which order arguments appear in the function (sometimes it's very important, as there are different conventions in literature).  The `f_wrapped` name also doesn't help anyone.

A quick patch:
```diff
diff --git a/mpmath/ctx_mp_python.py b/mpmath/ctx_mp_python.py
index 1c11d17..e406f01 100644
--- a/mpmath/ctx_mp_python.py
+++ b/mpmath/ctx_mp_python.py
@@ -1108,6 +1108,9 @@ def f_wrapped(ctx, *args, **kwargs):
         else:
             f_wrapped = f
         f_wrapped.__doc__ = function_docs.__dict__.get(name, f.__doc__)
+        import inspect
+        f_wrapped.__signature__ = inspect.signature(f)
+        f_wrapped.__name__ = f.__name__
         setattr(cls, name, f_wrapped)
 
     def _convert_param(ctx, x):
```
make things slightly better:
```pycon
>>> help(polylog)
Help on method polylog in module mpmath.ctx_mp_python:

polylog(s, z) method of mpmath.ctx_mp.MPContext instance
    Computes the polylogarithm, defined by the sum

    .. math ::

        \mathrm{Li}_s(z) = \sum_{k=1}^{\infty} \frac{z^k}{k^s}.
[...]
>>> inspect.signature(polylog)
<Signature (s, z)>
```

I'm not sure that we should make arguments names to be a part of API, however:
```pycon
>>> polylog(2, -3)
mpf('-1.9393754207667089')
>>> polylog(2, z=-3)
mpf('-1.9393754207667089')
>>> polylog(s=2, z=-3)
mpf('-1.9393754207667089')
```
but there is not simple ways to avoid this.

skirpichev:
Closes #899

### DIFF
--- a/mpmath/ctx_fp.py
+++ b/mpmath/ctx_fp.py
@@ -1,5 +1,6 @@
 import cmath
 import functools
+import inspect
 import math
 import sys
 
@@ -59,6 +60,11 @@ class FPContext(StandardBaseContext):
         else:
             f_wrapped = f
         f_wrapped.__doc__ = function_docs.__dict__.get(name, f.__doc__)
+        try:
+            f_wrapped.__signature__ = inspect.signature(f)
+        except ValueError:  # pragma: no cover
+            pass
+        f_wrapped.__name__ = f.__name__
         setattr(cls, name, f_wrapped)
 
     @functools.lru_cache
--- a/mpmath/ctx_iv.py
+++ b/mpmath/ctx_iv.py
@@ -1,6 +1,8 @@
+import inspect
 import numbers
 import sys
 
+from . import function_docs
 from . import libmp
 from .libmp import (MPZ_ONE, ComplexResult, dps_to_prec, finf, fnan, fninf,
                     from_float, from_int, from_str, fzero, int_types, mpc_hash,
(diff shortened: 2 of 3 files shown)


skirpichev:
This can be reused in docs to reduce boilerplate (variable declarations) like
```rst
:func:`~mpmath.zeta`
^^^^^^^^^^^^^^^^^^^^^
.. autofunction:: mpmath.zeta(s,a=1,derivative=0)
```
But lets do this in another pr.

PR #933, opened 2025-03-30
Add pretty_dps context property to control number of printed digits

### LINKED ISSUE #921: str and repr with pretty=True

skirpichev:
Closes #921

### DIFF
--- a/docs/basics.rst
+++ b/docs/basics.rst
@@ -221,6 +221,8 @@ Setting the ``mp.pretty`` option will use the ``str()``-style output for ``repr(
     >>> mpf(0.6)
     mpf('0.59999999999999998')
 
+To use enough digits to be able recreate value exactly, set ``mp.pretty_dps`` to ``"repr"``.
+
 The number of digits with which numbers are printed by default is determined by
 the working precision.  To specify the number of digits to show without
 changing the working precision, use :func:`format syntax support
--- a/mpmath/__main__.py
+++ b/mpmath/__main__.py
@@ -57,6 +57,7 @@ def main():
         lines.append(f'mp.prec = {args.prec}')
     if not args.no_pretty:
         lines.append('mp.pretty = True')
+        lines.append('mp.pretty_dps = "repr"')
 
     try:
         import IPython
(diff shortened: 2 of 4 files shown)


skirpichev:
CC @fredrik-johansson

skirpichev:
I'm going to merge this over #927.  This solution has no compatibility breaks, though I think that #927 is better.

PR #954, opened 2025-05-09
Support bin/hex float literals in the CLI

skirpichev:
See https://github.com/skirpichev/peps/pull/4 and https://github.com/python/cpython/issues/114667.


skirpichev:
I'm going to merge that.  This looks helpful for interactive work.  Also, it's the only example of non-AST-based source transformations.

PR #956, opened 2025-05-13
log compute - taylor series

hnb22:


skirpichev (commented):
Don't forgot to add tests, probably to test_log().

PR #959, opened 2025-05-28
Switch to anderson method to speed up zetazero

### LINKED ISSUE #958: zetazero is suboptimal?
Here is my test code:
```py
#!/usr/bin/python3

import mpmath
import sys
import time

def method1(ctx, range_):
    for x in range_:
        ctx.nprint(ctx.im(ctx.zetazero(x)), ctx.dps+2, strip_zeros=False)

def method2(ctx, range_):
    for x in range_:
        initial = mpmath.fp.im(mpmath.fp.zetazero(x))
        k = initial * mpmath.fp.eps
        root = ctx.findroot(ctx.siegelz, [initial, initial+k], solver="secant")
        ctx.nprint(root, ctx.dps+2, strip_zeros=False)

with mpmath.mp.workprec(100):
    t1 = time.time()
    method1(mpmath.mp, range(9990, 10000))
    t2 = time.time()
    print()
    method2(mpmath.mp, range(9990, 10000))
    t3 = time.time()
    print()
    with mpmath.mp.extraprec(30):
        method2(mpmath.mp, range(9990, 10000))
    t4 = time.time()
    print(f"Method 1: {t2 - t1} seconds")
    print(f"Method 2: {t3 - t2} seconds")
    print(f"Method 2 (more digits): {t4 - t3} seconds")
```
On my machine (mpmath 1.3.0) it produces the following:
```
9868.714659330643201731352988040
9869.812878325986458201906576863
9870.332258556616445749494357358
9872.000623378132955731719993135
9872.558666059604231912379326542
9873.802220903648406406781798992
9874.323957629064029672232728985
9875.218994098846767145580371717
9875.600956248756888587220599532
9876.479017063783790065513504888

9868.714659330643201731352988036
9869.812878325986458201906576864
9870.332258556616445749494357359
9872.000623378132955731719993140
9872.558666059604231912379326545
9873.802220903648406406781798987
9874.323957629064029672232728989
9875.218994098846767145580371711
9875.600956248756888587220599535
9876.479017063783790065513504884

9868.714659330643201731352988035649035334
9869.812878325986458201906576863536161488
9870.332258556616445749494357359202336099
9872.000623378132955731719993139527864878
9872.558666059604231912379326545444598555
9873.802220903648406406781798986980788123
9874.323957629064029672232728988590912143
9875.218994098846767145580371710801805777
9875.600956248756888587220599534886719983
9876.479017063783790065513504884342609249
Method 1: 15.640349388122559 seconds
Method 2: 7.93834376335144 seconds
Method 2 (more digits): 11.716793060302734 seconds
```
----
Method 1 is the built-n `zetazero` at the requested precision. Method 2 uses `zetazero` in `fp` context, and then refines that guess to the requested precision using the secant method. Obviously the timings will be different for individual people, and for large enough zeros a lower-precision `mp` context would need to be used because `fp` would not be enough. But the speed difference is significant and (for me) consistent across a range of precisions and zeros.

I think what's going on here is that the secant method is just intrinsically more efficient than the custom Newton-based method that `zetazero` uses. It converges more slowly, but by avoiding the need for a derivative it avoids a lot of extra work which more than makes up the difference.

As a side note, I'm printing the values to 2 extra digits to show that Method 2 also is more accurate.

d0sboots:
For the smooth nature of the zeta function, it is an unambiguous win, while maintaining the bracketing guarantee.

This closes #958. I tried a variety of optimizations; some improved only specific areas of the parameter space, others were small wins in some cases but might be considered risky. This was unambiguously better without any risk of altering the existing behavior.


skirpichev:
Could you, please, provide some benchmark, e.g. with pyperf?

d0sboots:
I am not sure what you are saying, exactly.

> I make many checks, also about what method to use. When I see your checks, really do not compare. For example I compute the zero 10^15:
> 
> zero number 1000000000000000 is equal to
> (mpc(real='0.5', imag='208514052006405.46942460229754774510611'),
[999999999999989L,1000000000000051L], 10L,
'(1)(1)(1)(02)(1)(1)(02)(1)(1)(1)(1)(1)(1)(1)(1)(02)(1)(1)(1)(1)(1)
(22)(1)(1)(1)(1)(1)(02)(1)(22)(1)(2110)(20)(1)(22)(1)(1)(02)(02)(014)
(1)(1)(1)(1)(1)(1)(1)')
computed in time 121893.041
}
\bigskip
> 
> The time spent on the computation is
$$ 1 \text{ day } 9 \text{ hours } 51 \text{ minutes and } 33.041 \text{ seconds.}$$
> 
> It is clear that this result is not reliable.

Are you saying that the change I made here is flawed, and you determined this by checking the zero 10^15? But this seems impossible, because the time listed for the computation is 1 day 9 hours, which is much longer than my PR has been open. So, I must assume you mean something else.

I do not dispute that you put careful engineering work into this function. Indeed, that is why the change I made here is making a *minimal* disturbance, and should have no observable numerical impact. If it *does*, it would imply (to me) that the root-finding method itself is broken.

d0sboots:
I'd tried pegasus already, it's not as good.

I reran all the numbers to capture as json so I could use compare_to.

```
zetazero(9) @53
===============

Mean +- std dev: [bench_illinois] 12.7 ms +- 0.3 ms -> [bench_pegasus] 11.2 ms +- 0.1 ms: 1.13x faster
Mean +- std dev: [bench_illinois] 12.7 ms +- 0.3 ms -> [bench_anderson] 11.3 ms +- 0.3 ms: 1.12x faster

zetazero(999) @53
=================

Mean +- std dev: [bench_illinois] 209 ms +- 4 ms -> [bench_pegasus] 198 ms +- 3 ms: 1.06x faster
Mean +- std dev: [bench_illinois] 209 ms +- 4 ms -> [bench_anderson] 175 ms +- 3 ms: 1.19x faster

zetazero(9999) @53
==================

Mean +- std dev: [bench_illinois] 739 ms +- 4 ms -> [bench_pegasus] 646 ms +- 6 ms: 1.14x faster
Mean +- std dev: [bench_illinois] 739 ms +- 4 ms -> [bench_anderson] 645 ms +- 6 ms: 1.15x faster

zetazero(9) @100
================

Mean +- std dev: [bench_illinois] 17.9 ms +- 0.4 ms -> [bench_pegasus] 15.0 ms +- 0.4 ms: 1.19x faster
Mean +- std dev: [bench_illinois] 17.9 ms +- 0.4 ms -> [bench_anderson] 15.0 ms +- 0.2 ms: 1.19x faster

zetazero(99) @100
=================

Mean +- std dev: [bench_illinois] 118 ms +- 4 ms -> [bench_pegasus] 106 ms +- 1 ms: 1.11x faster
Mean +- std dev: [bench_illinois] 118 ms +- 4 ms -> [bench_anderson] 94.6 ms +- 1.1 ms: 1.25x faster

zetazero(999) @100
==================

Mean +- std dev: [bench_illinois] 389 ms +- 5 ms -> [bench_pegasus] 412 ms +- 10 ms: 1.06x slower
Mean +- std dev: [bench_illinois] 389 ms +- 5 ms -> [bench_anderson] 376 ms +- 5 ms: 1.03x faster

zetazero(9999) @100
===================

Mean +- std dev: [bench_illinois] 1.22 sec +- 0.02 sec -> [bench_pegasus] 1.18 sec +- 0.01 sec: 1.03x faster
Mean +- std dev: [bench_illinois] 1.22 sec +- 0.02 sec -> [bench_anderson] 1.18 sec +- 0.02 sec: 1.03x faster

zetazero(9) @200
================

Mean +- std dev: [bench_illinois] 20.2 ms +- 0.5 ms -> [bench_pegasus] 18.5 ms +- 0.4 ms: 1.09x faster
Mean +- std dev: [bench_illinois] 20.2 ms +- 0.5 ms -> [bench_anderson] 18.6 ms +- 0.2 ms: 1.09x faster

zetazero(99) @200
=================

Mean +- std dev: [bench_illinois] 166 ms +- 5 ms -> [bench_pegasus] 142 ms +- 3 ms: 1.17x faster
Mean +- std dev: [bench_illinois] 166 ms +- 5 ms -> [bench_anderson] 142 ms +- 2 ms: 1.17x faster

zetazero(999) @200
==================

Mean +- std dev: [bench_illinois] 332 ms +- 7 ms -> [bench_pegasus] 321 ms +- 6 ms: 1.04x faster
Mean +- std dev: [bench_illinois] 332 ms +- 7 ms -> [bench_anderson] 320 ms +- 4 ms: 1.04x faster

zetazero(9999) @200
===================

Mean +- std dev: [bench_illinois] 1.19 sec +- 0.01 sec -> [bench_pegasus] 1.17 sec +- 0.02 sec: 1.02x faster
Mean +- std dev: [bench_illinois] 1.19 sec +- 0.01 sec -> [bench_anderson] 1.16 sec +- 0.01 sec: 1.03x faster

zetazero(9) @400
================

Mean +- std dev: [bench_illinois] 28.4 ms +- 0.4 ms -> [bench_pegasus] 27.1 ms +- 0.6 ms: 1.05x faster
Mean +- std dev: [bench_illinois] 28.4 ms +- 0.4 ms -> [bench_anderson] 27.0 ms +- 0.7 ms: 1.05x faster

zetazero(99) @400
=================

Mean +- std dev: [bench_illinois] 187 ms +- 3 ms -> [bench_pegasus] 162 ms +- 1 ms: 1.15x faster
Mean +- std dev: [bench_illinois] 187 ms +- 3 ms -> [bench_anderson] 162 ms +- 2 ms: 1.16x faster

zetazero(999) @400
==================

Mean +- std dev: [bench_illinois] 533 ms +- 6 ms -> [bench_pegasus] 521 ms +- 7 ms: 1.02x faster
Mean +- std dev: [bench_illinois] 533 ms +- 6 ms -> [bench_anderson] 522 ms +- 9 ms: 1.02x faster

zetazero(9999) @400
===================

Mean +- std dev: [bench_illinois] 1.37 sec +- 0.01 sec -> [bench_pegasus] 1.28 sec +- 0.02 sec: 1.08x faster
Mean +- std dev: [bench_illinois] 1.37 sec +- 0.01 sec -> [bench_anderson] 1.27 sec +- 0.02 sec: 1.08x faster

Benchmark hidden because not significant (1): zetazero(99) @53

Geometric mean
==============

bench_pegasus: 1.07x faster
bench_anderson: 1.10x faster
```

skirpichev:
BTW, it seems that change in separate_zeros_in_block() doesn't affect your tests.

skirpichev:
>  I recall to select very carefully the method to use to compute to high precision the zeros.

That's something worth to be documented in code.

mpmath/functions/zetazeros.py
```diff
@@ -121,7 +121,7 @@ def separate_zeros_in_block(ctx, zero_number_block, T, V, limitloop=None,
                 f = lambda x: ctx.rs_z(x,derivative=1)
                 t0=T[kMax-1]
                 t1 = T[kMax]
-                t=ctx.findroot(f,  (t0,t1), solver ='illinois',verify=False, verbose=False)
+                t=ctx.findroot(f,  (t0,t1), solver ='anderson',verify=False, verbose=False)
```
skirpichev, line 124: Not sure why `verify=False` is used here.

mpmath/functions/zetazeros.py
```diff
@@ -159,7 +159,7 @@ def separate_my_zero(ctx, my_zero_number, zero_number_block, T, V, prec):
         index +=1
         precs = [precs[0] // 2 +3+2*index] + precs
     ctx.prec = precs[0] + guard
-    r = ctx.findroot(lambda x:ctx.siegelz(x), (t0,t1), solver ='illinois', verbose=False)
+    r = ctx.findroot(ctx.siegelz, (t0,t1), solver ='anderson', verbose=False)
```
skirpichev, line 162: I moved minor changes from this pr (lambda, etc) to https://github.com/mpmath/mpmath/pull/961

mpmath/functions/zetazeros.py
d0sboots: I don't think it has any effect, but I left it to avoid an unrelated change.

d0sboots:
> So, we have to use a method based on bisection: Illinois, Pegasus or Anderson. I recall checking all these methods (and other ones).
> 
> You have been checking some small zeros, this is not sufficient, the problem was when zeros were very near. In some cases the program returned a zero not in the interval [a',b'], that is not the n-th zero. You have to check this type of zero. Improving the time in small zeros some fraction of a second, when the main time is spent to locate the zero is not a great advance.

The reason I did not check large zeros is because there should *not* be any correctness implications switching between these methods; the only implication should be performance. All of these methods are based on bisection and maintain the invariant that they keep the root bracketed, as long as there is only a single root in the interval from the start. That is the same invariant that *all* of them require.

Let me be more plain: If you found a correctness difference between these methods when implementing zetazeros, it either speaks to a bug in findroot (or one of its solvers), or the original invariant of having an interval bracketing a single root wasn't holding. It sounds like in some cases the latter was true (the interval was previously too wide), but if zetazeros is correct now, that is no longer the case. 

If you are worried about particular difficult cases, the right answer is for those to be unittests. If they are too large to be run regularly, they can be added as skipped tests, or there are other solutions - I'm not sure how it's done in this project, but I am happy to add more tests in this PR. But relying on manual testing for correctness is not sustainable.

d0sboots:
> BTW, it seems that change in separate_zeros_in_block() doesn't affect your tests.

Yes, I changed that for consistency but it only triggers on specific values, which my benchmarks don't hit. I didn't expect it to be a significant source of the runtime for the ranges where this optimization is noticable.

mpmath/functions/zetazeros.py
skirpichev: Also added to https://github.com/mpmath/mpmath/pull/961

d0sboots:
> For example zetazero(1048449114) mpc(real='0.5', imag='388858886.0022851217767970582610330824019119') zetazero(1048449115) mpc(real='0.5', imag='388858886.0023936897027167200756700895162012')
> 
> Van de Lune gives me two near zeros zetazero(3570918901) mpc(real='0.5', imag='1239587702.547450313754819029745631023067728') zetazero(3570918902) mpc(real='0.5', imag='1239587702.547523871747088188555055346198066')
> 
> Or check the zeros Gourdon find closer for n<10^13
> 
> zero number 8637740722917 is (mpc(real='0.5', imag='2124447368584.3929646615152911269375') zero number 8637740722918 is (mpc(real='0.5', imag='2124447368584.3929817060386050127775')

In turns out that two of these were *already* in the unittests (the first one and the third one, although the latter is commented out for "potentially taking hours"). I added the middle one, and I'm running the commented-out ones to verify.

skirpichev:
> there is only a single root in the interval from the start.

@ariasdereyna, is this a valid assumption about your code, that calls findroot() or we miss something?

ariasdereyna:
@skirpichev

The short answer is yes.

The program determine first an interval (g_r, g_s] containing m zeros, we know also that our zero is the k zero of this interval. 

After we separate the m zeros in the interval by locating the adequate number of changes of sign, we determine an interval (a,b)\subset(g_r,g_s]  that contains only the zero we are computing. We have also a change of sign Z(a) Z(b) < 0.

At this moment is when we compute the zero to high precision. 

Note that in the process of separation of the m zeros, maybe we have to increase the precision to which we are computing. Because the zeros can be very near. Frequently we are computing to a higher precision, to the one we asked for. This explain that to compute a zero, the program is faster if we use from the start a higher precision. 

I do not recall why I preferred the method Illinois to Anderson. Theoretically both are good. Certainly I used Anderson on many trials.

skirpichev:
> The short answer is yes.

So, correctness shouldn't be an issue, right?

I tried some of your latest "uneasy cases":
| Benchmark                 | ref      | anderson               |
|---------------------------|:--------:|:----------------------:|
| zetazero(1048449114) @113 | 11.2 sec | 9.49 sec: 1.18x faster |
| zetazero(3570918901) @113 | 17.5 sec | 17.2 sec: 1.02x faster |
| zetazero(3570918902) @113 | 13.0 sec | 13.4 sec: 1.03x slower |
| Geometric mean            | (ref)    | 1.04x faster           |

Benchmark hidden because not significant (1): zetazero(1048449115) @113

So, we have some speedup here as well.

> I do not recall why I preferred the method Illinois to Anderson.

I understand that.

Though, I see no performance regressions so far.  @d0sboots, could you please test other posted cases?

<details>

```py
# bench.py
import pyperf
import mpmath
from mpmath import zetazero

runner = pyperf.Runner()
#for prec in [53, 100, 200, 400]:
#    with mpmath.mp.workprec(prec):
#        for x in [9, 99, 999, 9999]:
#            runner.bench_func(f"zetazero({x}) @{prec}", zetazero, x)
prec=113
with mpmath.mp.workprec(prec):
    for x in [1048449114, 1048449115, 3570918901,
              3570918902,
              # 8637740722917, 8637740722918,
              ]:
        runner.bench_func(f"zetazero({x}) @{prec}", zetazero, x)
```

</details>

skirpichev:
Here all tests for your latest "uneasy cases":
```
$ python a.py  # on the branch
(0.5 + 388858886.002285j)  in  8.106628894805908
(0.5 + 388858886.002394j)  in  11.133415460586548
(0.5 + 1239587702.54745j)  in  15.664631605148315
(0.5 + 1239587702.54752j)  in  12.10592269897461
(0.5 + 2124447368584.39j)  in  3744.5918004512787
(0.5 + 2124447368584.39j)  in  3611.094599723816
$ python a.py  # on master
(0.5 + 388858886.002285j)  in  9.89375615119934
(0.5 + 388858886.002394j)  in  12.0871901512146
(0.5 + 1239587702.54745j)  in  15.6492760181427
(0.5 + 1239587702.54752j)  in  12.52377438545227
(0.5 + 2124447368584.39j)  in  3731.0318746566772
(0.5 + 2124447368584.39j)  in  3596.809224843979
```

Original benchmark on my system:
| Benchmark           | ref      | patch                  |
|---------------------|:--------:|:----------------------:|
| zetazero(9) @53     | 74.5 ms  | 65.2 ms: 1.14x faster  |
| zetazero(999) @53   | 1.37 sec | 1.14 sec: 1.20x faster |
| zetazero(9999) @53  | 4.70 sec | 4.10 sec: 1.15x faster |
| zetazero(9) @100    | 103 ms   | 86.0 ms: 1.20x faster  |
| zetazero(99) @100   | 758 ms   | 609 ms: 1.24x faster   |
| zetazero(999) @100  | 2.51 sec | 2.41 sec: 1.04x faster |
| zetazero(9999) @100 | 7.62 sec | 7.34 sec: 1.04x faster |
| zetazero(9) @200    | 122 ms   | 113 ms: 1.08x faster   |
| zetazero(99) @200   | 1.06 sec | 914 ms: 1.16x faster   |
| zetazero(999) @200  | 2.18 sec | 2.10 sec: 1.04x faster |
| zetazero(9999) @200 | 7.63 sec | 7.43 sec: 1.03x faster |
| zetazero(9) @400    | 174 ms   | 165 ms: 1.05x faster   |
| zetazero(99) @400   | 1.20 sec | 1.04 sec: 1.16x faster |
| zetazero(999) @400  | 3.43 sec | 3.36 sec: 1.02x faster |
| zetazero(9999) @400 | 8.74 sec | 8.08 sec: 1.08x faster |
| Geometric mean      | (ref)    | 1.10x faster           |

Benchmark hidden because not significant (1): zetazero(99) @53

I think we can revert this small patch later, but for now I'll merge this.  It definitely improves things for small indexes.

<details>

```py
from time import time
from mpmath import zetazero
for n in [1048449114,
          1048449115,
          3570918901,
          3570918902,
          8637740722917,
          8637740722918,
          #1000000000000000
          ]:
    start = time()
    r = zetazero(n)
    end = time()
    print(r, " in ", end - start)
```

</details>

Issue #970, opened 2025-06-17
mpmath.factorial() and mpmath.gamma() should have better error messages


Mehdical-Man:
### The current error message is:
```py
>>> mpmath.factorial(-1)
    ~~~~~~~~~~~~~~~~^^^^
ValueError: gamma function pole
```
### A better error message should be:
```py
>>> mpmath.factorial(-1)
    ~~~~~~~~~~~~~~~~^^^^
ValueError: factorial not defined for negative integers
```
### The current error message is:
```py
>>> mpmath.gamma(0)
    ~~~~~~~~~~~~^^^
ValueError: gamma function pole
```
### A better error message should be:
```py
>>> mpmath.gamma(0)
    ~~~~~~~~~~~~^^^
ValueError: gamma not defined for 0 or negative integers
```

skirpichev:
Closed, as explained in the pr thread.

PR #971, opened 2025-06-17
Improve error messages for mpmath.factorial and mpmath.gamma and fixe…

Mehdical-Man:
…d docstring

1. if int(x) ≤ 0 in mpmath.gamma(x): the error message is "gamma not defined for 0 or negative integers"
2. if int(x) < 0 in mpmath.factorial(x): the error message is "factorial not defined for negative integers"
3. (x!) == xΓ(x) not (x!) == xΓ(x+1)

### DIFF
--- a/mpmath/libmp/gammazeta.py
+++ b/mpmath/libmp/gammazeta.py
@@ -1684,7 +1684,7 @@ def mpf_gamma(x, prec, rnd=round_fast, type=0):
     function, G(x), as well as the following versions of the same:
 
     type = 0 -- G(x)                    [standard gamma function]
-    type = 1 -- G(x+1) = x*G(x+1) = x!  [factorial]
+    type = 1 -- G(x+1) = x*G(x) = x!    [factorial]
     type = 2 -- 1/G(x)                  [reciprocal gamma function]
     type = 3 -- log(|G(x)|)             [log-gamma function, real part]
     """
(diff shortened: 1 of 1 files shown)


Mehdical-Man:
See #970

mpmath/libmp/gammazeta.py
```diff
@@ -1684,7 +1684,7 @@ def mpf_gamma(x, prec, rnd=round_fast, type=0):
     function, G(x), as well as the following versions of the same:
 
     type = 0 -- G(x)                    [standard gamma function]
-    type = 1 -- G(x+1) = x*G(x+1) = x!  [factorial]
+    type = 1 -- G(x+1) = x*G(x) = x!    [factorial]
```
skirpichev, line 1687: Good!

skirpichev (changes requested):
Please revert the rest.

I don't think that error messages should describe function domain.  There is an issue on what we should do in case of such errors (i.e. return a complex infinity as Mathematica), but that's a different story.

(BTW, new error message just wrong.)

mpmath/libmp/gammazeta.py
skirpichev: Ok, that included in https://github.com/mpmath/mpmath/pull/972, thanks.

PR #982, opened 2025-07-23
Add Fox H-function with rational A and B parameters

### LINKED ISSUE #980: Is it possible to integrate an implement of Fox H-function with rational scale parameter to mpmath
During my research (about product of random variables) I found Fox H-function become relevant and I need numerical values; mathematica has an implementation of Fox H-function; however, mathematica seems to have bug ([link](https://mathematica.stackexchange.com/questions/261563/numerical-values-of-fox-h-function)) and I've experienced bug for my parameter set; on the other hand, I appreciate mpmath (great work!) for having open-source implementation of Meijer G function. After some searching I found that it is possible to reduce Fox H-function to Meijer G-function as long as the A and B parameters are rational

Image
(Src: Integrals and Series, Vol 3, Prudnikov and Brychkov, p. 629)

For my case, all A/B are positive integers, so it can be simplified to

Image
(Src: my notebook)

And the implementation (generated by AI, with modification)
```python
import mpmath as mp

mp.dps = 256  # increase if needed

def foxh(z, a, A, b, B, m, n):
    """
    Compute the Fox H-function H_{p,q}^{m,n}(z) with integer scale parameters Ai and Bj
    based on formula (28) using mpmath's meijer G-function.

    Parameters:
    - z : argument at which to evaluate H
    - a : list of length p containing ai
    - A : list of length p containing integer Ai
    - b : list of length q containing bj
    - B : list of length q containing integer Bj
    - m : integer, number of bj terms in numerator
    - n : integer, number of ai terms in denominator

    Returns:
    - value of H_{p,q}^{m,n}(z)
    """
    # Validate integer scale parameters
    if not all(isinstance(Ai, int) and Ai > 0 for Ai in A):
        raise ValueError("All Ai must be positive integers")
    if not all(isinstance(Bj, int) and Bj > 0 for Bj in B):
        raise ValueError("All Bj must be positive integers")

    p = len(a)
    q = len(b)

    # Build the expanded parameter lists for the Meijer G-function
    a_tilde = []
    for ai, Ai in zip(a, A):
        for k in range(Ai):
            a_tilde.append(ai + k / Ai)

    b_tilde = []
    for bj, Bj in zip(b, B):
        for k in range(Bj):
            b_tilde.append(bj + k / Bj)

    # Compute beta = prod(B_j^B_j) / prod(A_i^A_i)
    beta = mp.mpf(1)
    for Ai in A:
        beta /= Ai**Ai
    for Bj in B:
        beta *= Bj**Bj

    # Compute m_tilde and n_tilde (lengths for numerator/denominator lists)
    m_tilde = sum(B[:m])
    n_tilde = sum(A[:n])

    # Compute a_star and c_star
    a_star = sum(A[:n]) - sum(A[n:]) + sum(B[:m]) - sum(B[m:])
    c_star = m + n - (p + q) / 2

    # Compute M factor = prod(B_j^(b_j - 1/2)) / prod(A_i^(a_i - 1/2))
    M = mp.mpf(1)
    for bj, Bj in zip(b, B):
        M *= Bj**(bj - mp.mpf(1)/2)
    for ai, Ai in zip(a, A):
        M /= Ai**(ai - mp.mpf(1)/2)

    # Prefactor from formula (28)
    prefactor = (2 * mp.pi)**(c_star - a_star/2) * M

    # Evaluate the Meijer G-function
    return prefactor * mp.meijerg(
        [a_tilde[:n_tilde], a_tilde[n_tilde:]],
        [b_tilde[:m_tilde], b_tilde[m_tilde:]],
        z / beta
    )

def chisq1(z):
    prefactor = mp.sqrt(1 / (2 * mp.pi)) * z**(-1/2)
    return prefactor * mp.exp(-z / 2)

def chisq1_foxH(z):
    b = [-1/2]
    B = [2]
    w = z * z / (4)
    prefactor = mp.mpf(1) / (mp.sqrt(mp.pi))
    return prefactor * foxh(w, [], [], b, B, m=1, n=0)

print(chisq1_foxH(4.0))  # Example for chi-squared distribution with 1 degree of freedom
print(chisq1(4.0))
```

I would like to contribute such implementation to mpmath for other people's convenience (and for checking my implementation/research results are correct and reasonable), and I think rational A and B would cover most use cases. My question is: **Will mpmath accept such implementation that covers only the rational parameter**. Of course, I will change the signature to match meijer G-function calling convention, change the integer part to deal with rationals, and add documentation.

For a full implementation, I think I need to deal with Fox–Wright function, which we seems to lack an implementation, and I may not have the bandwidth to investigate further.

I am new to the area of numerial computing so I may make mistake here. Really appreciate your time and attention!

ZenithalHourlyRate:
Closes #980

Please let me know if I missed anything.


mpmath/functions/hypergeometric.py
```diff
@@ -1062,6 +1063,82 @@ def h(*args):
...
+    z = ctx.convert(z)
+    r = ctx.convert(r)
+
+    # Convert integer to rationals in form Ai / 1
+    Afrac = [(Ai, 1) if isinstance(Ai, int) else Ai for Ai in A]
+    Bfrac = [(Bj, 1) if isinstance(Bj, int) else Bj for Bj in B]
+
+    if not all(isinstance(c, int) and isinstance(d, int) and
+               c > 0 and d > 0 for c, d in Afrac + Bfrac):
+        raise NotImplementedError("All A and B must be positive rationals")
```
skirpichev, line 1089: That's not true, you are accepting only integers, right?
BTW, mpf's ARE rationals, i.e. you can use MeijerG identify in general.  Though, not sure it worth.

ZenithalHourlyRate: I've adapted the code to accept either `A = n` an integer or `A = (p,q)` a rational `p/q`; the code above converts `A = n` to `A = (n, 1)`, and this line checks the condition that they are all in form `(p, q)`.

While mpf are rationals, I think letting user specify `(p,q)` instead of `mpf('p/q')` has the behefit that I could recover the true denominator more easily, and LCM would be small; I am not sure `mpf('p/q').to_rational` would exactly return `q` for me, if it is, I could change A to accept mpf as well.

skirpichev: > I've adapted the code to accept either A = n an integer or A = (p,q) a rational p/q; the code above converts A = n to A = (n, 1)

I see.  No, this is not good.  I don't think we have other function, accepting rationals that way.  Every parameter should be just some number.

> While mpf are rationals, I think letting user specify (p,q) instead of mpf('p/q') has the behefit that I could recover the true denominator more easily

Just as for a builtin float, you can use `as_integer_ratio()` method to recover numerator/denominator for a mpf.

ZenithalHourlyRate: > I don't think we have other function, accepting rationals that way. Every parameter should be just some number.

I found such usage here https://mpmath.org/doc/current/functions/zeta.html

>  The parameter a is usually a rational number `p/q`, and may be specified as such by passing an integer tuple `(p,q)`.

Anyway, I'll change it to use mpf

ZenithalHourlyRate:
Suggested changes applied.

PR #1017, opened 2025-10-22
Fix #1002: Raise ValueError when zero matrix is passed to logm()

### LINKED ISSUE #1002: `logm()` of zero matrix hangs instead of raising error.
See title. It is better to return a `ValueError` or `ZeroDivisionError`.

baranwalayush:
Closes #1002

### DIFF
--- a/mpmath/matrices/calculus.py
+++ b/mpmath/matrices/calculus.py
@@ -433,6 +433,8 @@ class MatrixCalculusMethods:
 
         """
         A = ctx.matrix(A)
+        if ctx.mnorm(A, 'inf') == 0:
+            raise ValueError("The logarithm is undefined for the zero matrix.")
         prec = ctx.prec
         try:
             ctx.prec += 10
--- a/mpmath/tests/test_calculus.py
+++ b/mpmath/tests/test_calculus.py
@@ -3,7 +3,7 @@ import pytest
 from mpmath import (arange, chebyfit, cos, cosm, differint, e, euler, exp,
                     expm, fourier, fourierval, inf, invertlaplace, j, limit,
                     log, matrix, mp, mpf, norm, pade, pi, polyroots, polyval,
-                    sin, sinm, sqrt)
+                    sin, sinm, sqrt, logm)
 
 
 def test_approximation():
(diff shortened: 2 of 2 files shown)


mpmath/matrices/calculus.py
```diff
@@ -433,6 +433,8 @@ def logm(ctx, A):
 
         """
         A = ctx.matrix(A)
+        if ctx.mnorm(A, 'inf') == 0:
+            raise ValueError("The logarithm is undefined for the zero matrix.")
```
skirpichev: It seems, that for some (all?) other singular matrices the algorithm works:
```pycon
>>> A = zeros(2)
>>> A[0, 1]=1
>>> logm(A)
  File "/home/sk/src/mpmath/mpmath/matrices/calculus.py", line 331, in sqrtm
    Y, Z = 0.5*(Y+ctx.inverse(Z)), 0.5*(Z+ctx.inverse(Y))
                                          ~~~~~~~~~~~^^^
  File "/home/sk/src/mpmath/mpmath/matrices/linalg.py", line 295, in inverse
    A, p = ctx.LU_decomp(A)
           ~~~~~~~~~~~~~^^^
  File "/home/sk/src/mpmath/mpmath/matrices/linalg.py", line 132, in LU_decomp
    raise ZeroDivisionError('matrix is numerically singular')
ZeroDivisionError: matrix is numerically singular

During handling of the above exception, another exception occurred:
[...]
  File "/home/sk/src/mpmath/mpmath/matrices/linalg.py", line 132, in LU_decomp
    raise ZeroDivisionError('matrix is numerically singular')
ZeroDivisionError: matrix is numerically singular
```

1. I would prefer similar error (type and message)
2. Can we avoid special case?

baranwalayush: Yes, I did mention that in the issue itself. The reason and ways to fix that. Please take a look there.

skirpichev: Did you try another way?

I would rather avoid check for very special case, that will affect all calls of this method.

baranwalayush: yes, we can have raise a value error with some error message in place of the convergence error.

PR #1034, opened 2026-02-20
polylog: z to positive infinity should return nan

### LINKED ISSUE #1033: polylog(s, +inf) returns nans for real and imaginary components
#904 fixed #473 (-inf input, correctly unit tested in #904).  But I think it is wrong for the +inf case.

```
In [90]: mpmath.polylog(2, inf)
Out[90]: mpf('-inf')
```

But the limit looks more like `-inf - infi` to me:
```
In [87]: mpmath.polylog(2, 1e10)
Out[87]: mpc(real='-261.80503739032343', imag='-72.337844124154643')

In [88]: mpmath.polylog(2, 1e20)
Out[88]: mpc(real='-1057.0897539619832', imag='-144.67568824830929')

In [89]: mpmath.polylog(2, 1e100)
Out[89]: mpc(real='-26506.200684258292', imag='-723.37844124154651')
```
(I think normally that would be NaN).

Further evidence that NaN is appropriate: we see a sign flip when flipping the sign of a small imaginary input (with large real): 
```
In [8]: mpmath.polylog(2, 1e20 + 1e-10j)
Out[8]: mpc(real='-1057.0897539619832', imag='144.67568824830929')

In [9]: mpmath.polylog(2, 1e20 - 1e-10j)
Out[9]: mpc(real='-1057.0897539619832', imag='-144.67568824830929')
```
(and Wikipedia mentions a branch cut along the positive real axes above 1).

- - - -

Perhaps too minor to block Issue #1015, although it does seem to be a regression from 1.3:
```
In [11]: mpmath.__version__
Out[11]: '1.3.0'

In [12]: mpmath.polylog(2, inf)
Out[12]: mpc(real='nan', imag='nan')
```

cbm755:
Fixes #1033.

### DIFF
--- a/mpmath/functions/zeta.py
+++ b/mpmath/functions/zeta.py
@@ -402,7 +402,7 @@ def polylog_series(ctx, s, z):
 def polylog_continuation(ctx, n, z):
     if n < 0:
         return z*0
-    if ctx._is_real_type(z) and ctx.isinf(z) and n > 0:
+    if ctx._is_real_type(z) and ctx.isinf(z) and n > 0 and z < 0:
         return ctx.ninf
     twopij = 2j * ctx.pi
     a = -twopij**n/ctx.fac(n) * ctx.bernpoly(n, ctx.ln(z)/twopij)
--- a/mpmath/tests/test_functions2.py
+++ b/mpmath/tests/test_functions2.py
@@ -2445,6 +2445,10 @@ def test_issue_473():
     assert mp.polylog(4, -mp.inf) == -mp.inf
     assert mp.polylog(5, -mp.inf) == -mp.inf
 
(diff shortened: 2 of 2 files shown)


skirpichev (approved):
LGTM as a quick fix.

But I think that a correct answer should be rather `-inf+nanj` (see mentioned identity).

BTW, WolframEngine 14.3 shows `-Infinity` even for `PolyLog[1,Infinity]`.

cbm755:
> But I think that a correct answer should be rather -inf+nanj (see mentioned identity).

I agree.  Happily the unit test added here would still be acceptable b/c `isnan(...)` should be true for `-inf + nanj` as well.

Issue #1044, opened 2026-02-26
mpmath 1.4.0 breaks Frozen Applications due to importlib.metadata usage


FlurinArner:
## Summary
mpmath 1.4.0 breaks frozen applications (PyInstaller, py2app, cx_Freeze, Nuitka) by switching from hardcoded version strings to dynamic lookup via `importlib.metadata.version()`.

## Root Cause

**mpmath 1.4.0** ([`mpmath/__init__.py` L1-2](https://github.com/mpmath/mpmath/blob/master/mpmath/__init__.py#L1-L2)):
```python
from importlib.metadata import version
__version__ = version(__name__)
```

**Previous versions (1.2.1, 1.3.0)**:
```python
__version__ = '1.3.0'  # Hardcoded
```

**Problem**: Frozen applications bundle Python code but not package metadata (`.dist-info` directories), causing `importlib.metadata.version()` to fail with `PackageNotFoundError`.

## Minimal Reproduction

Simulate frozen app behavior without requiring PyInstaller:

```python
#!/usr/bin/env python3
"""Simulates frozen app behavior where package metadata is unavailable"""

import importlib.metadata

# Mock metadata to simulate frozen environment
original_version = importlib.metadata.version

def mock_version(package_name):
    if package_name == "mpmath":
        raise importlib.metadata.PackageNotFoundError(
            f"No package metadata was found for {package_name}"
        )
    return original_version(package_name)

importlib.metadata.version = mock_version

import mpmath  # ✗ FAILS on 1.4.0
print(f"mpmath version: {mpmath.__version__}")
```

## Workarounds

**Downgrade to mpmath 1.3.0:**
```bash
pip install mpmath==1.3.0
```
or in Poetry:
```toml
mpmath = "1.3.0"
```

## Proposed Solution

Use a fallback for when metadata is unavailable:

```python
try:
    from importlib.metadata import version
    __version__ = version(__name__)
except Exception:
    # Fallback for frozen apps or when metadata unavailable
    __version__ = "1.4.0"  # Update this string in each release
```

## Environment
- **mpmath**: 1.4.0
- **Python**: 3.10
  - expected on all 3.9+ versions supported by mpmath 1.4.0
- **Freezing tools**: PyInstaller 5.4.1
  - expected on all others also (py2app, cx_Freeze, Nuitka)
- **OS**: Windows

oscarbenjamin:
It also slows down import time e.g. here with 1.3.0:
```
[...]

PR #1045, opened 2026-02-26
Fix/frozen metadata version

FlurinArner:
This PR fixes a regression in `mpmath 1.4.0` where importing mpmath in frozen applications (PyInstaller, cx_Freeze, etc.) fails with `importlib.metadata.PackageNotFoundError` due to missing package metadata.

- Adds a try/except fallback to set `__version__` to a hardcoded string if metadata is unavailable (e.g., in frozen apps).
- Adds a regression test that mocks missing metadata to ensure the fallback works and the hardcoded version is updated in each release.

**Closes**: #1044

**Motivation:**
`mpmath 1.4.0` switched to dynamic version lookup via `importlib.metadata.version(__name__)`, which breaks in frozen apps that do not bundle `.dist-info` metadata. This PR restores compatibility for those environments.

### DIFF
--- a/mpmath/tests/test_issue_metadata_fallback.py
+++ b/mpmath/tests/test_issue_metadata_fallback.py
@@ -5,11 +5,7 @@ import importlib.metadata
 import importlib
 
 def test_issue_metadata_fallback(monkeypatch):
-    # Import mpmath normally first to get the expected version
-    import mpmath
-    expected_version = mpmath.__version__
-
-    # Save original version function
+    # Mock metadata.version to simulate frozen app where metadata is unavailable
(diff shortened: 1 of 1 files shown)


skirpichev:
Sorry, but I don't see any progress here.  There is an alternative pr, that aims to fix same problem.  I will appreciate your review.

It looks like the pr was created by AI assistant.  Proper reloading of modules is tricky.  I have such tests in the diofant's test suite, maybe it helps: https://github.com/diofant/diofant/blob/master/diofant/tests/core/test_cache.py.

Meanwhile, I'm closing this.  Feel free to reopen, if you will made tests working.

PR #1055, opened 2026-03-02
Test/frozen package version

### LINKED ISSUE #1044: mpmath 1.4.0 breaks Frozen Applications due to importlib.metadata usage

FlurinArner:
- Separate job to test the `pyinstaller` frozen application generation and `mpmath.__version__` availablity.
- Add bash script `mpmath/tests/test_version_frozen.sh`
- Ignore the `test_local/` folder generated in the `test_version_frozen.sh` bash script.
- Resolves the testing requirement for #1044

closes #1044


mpmath/tests/test_version_frozen.sh
```diff
@@ -0,0 +1,66 @@
+#!/bin/bash
```
skirpichev, line 1: I think you are not using any bashisms.  Why not sh?

Please also add a small commentary here with brief description and point to issue number.

mpmath/tests/test_version_frozen.sh
```diff
@@ -0,0 +1,66 @@
...
+set -e
+
+# Clean up any previous test artifacts
+echo "Cleaning up previous test artifacts..."
+rm -rf test_local build dist *.spec .venv
+
+mkdir -p test_local
```
skirpichev: You can use mktemp utility to manage temporary directory outside of the mpmath tree.

mpmath/tests/test_version_frozen.sh
```diff
@@ -0,0 +1,66 @@
...
+
+echo "Creating virtual environment..."
+python3 -m venv .venv
+source .venv/bin/activate
+
+
+cd .. # Leave the test_local directory to build the tarball from the local repo
+echo "Building tarball from local repo..."
+# Build the source distribution
+python3 -m pip install --quiet build
```
skirpichev: Please install all deps here.  And please don't use ``--quiet`` and so on to silence output.

mpmath/tests/test_version_frozen.sh
```diff
@@ -0,0 +1,66 @@
...
+
+
+# Create version_script.py that imports mpmath and prints the package version
+cat << EOF > version_script.py
+import mpmath
+print("mpmath version:", mpmath.__version__)
+EOF
+
+# Save version_script.py output to DIRECT_VERSION for later comparison
+DIRECT_VERSION=$(python version_script.py | grep "mpmath version:" | awk '{print $3}')
```
skirpichev: You can get version by `python -m mpmath --version`.

mpmath/tests/test_version_frozen.sh
```diff
@@ -0,0 +1,66 @@
...
+# Assert that the version from the frozen/bundled executable matches the direct version
+if [ "$DIRECT_VERSION" == "$FROZEN_VERSION" ]; then
+    echo "Test passed: Version matches in frozen (bundled) executable."
+else
+    echo "Test failed: Version mismatch."
+    echo "Direct version: $DIRECT_VERSION"
+    echo "Frozen version: $FROZEN_VERSION"
+    exit 1
+fi
+
```
skirpichev:
```suggestion
```

.gitignore
```diff
@@ -72,3 +72,6 @@ htmlcov/
 # per-project directories for virtual environments
 venv/
 .venv/
+
+# Temporary test directory
+test_local/
```
skirpichev: And this will be unnecessary with mktemp.

mpmath/tests/test_version_frozen.sh
```diff
@@ -0,0 +1,66 @@
...
+python3 -m pip install --quiet build
+python3 -m build --sdist
+
+# Find and install the generated tarball
+TARBALL=$(ls -t dist/*.tar.gz | head -1)
+echo "Generated tarball: $TARBALL"
+
+# Install mpmath from the generated tarball
+echo "Installing mpmath from tarball..."
+pip install --quiet dist/"$(basename $TARBALL)"
```
skirpichev: Don't be quiet here as well.

mpmath/tests/test_version_frozen.sh
```diff
@@ -0,0 +1,66 @@
...
+echo "Building tarball from local repo..."
+# Build the source distribution
+python3 -m pip install --quiet build
+python3 -m build --sdist
+
+# Find and install the generated tarball
+TARBALL=$(ls -t dist/*.tar.gz | head -1)
+echo "Generated tarball: $TARBALL"
+
+# Install mpmath from the generated tarball
```
skirpichev: This looks redundant.

mpmath/tests/test_version_frozen.sh
```diff
@@ -0,0 +1,66 @@
...
+./dist/version_script 2>&1 | tee output.txt
+
+# Extract the version from the output
+FROZEN_VERSION=$(grep "mpmath version:" output.txt | awk '{print $3}')
+
+cd ..
+deactivate
+
+
+# Assert that the version from the frozen/bundled executable matches the direct version
```
skirpichev:
```suggestion
```
Also useless.

mpmath/tests/test_version_frozen.sh
```diff
@@ -0,0 +1,66 @@
...
+# Create version_script.py that imports mpmath and prints the package version
+cat << EOF > version_script.py
+import mpmath
+print("mpmath version:", mpmath.__version__)
+EOF
+
+# Save version_script.py output to DIRECT_VERSION for later comparison
+DIRECT_VERSION=$(python version_script.py | grep "mpmath version:" | awk '{print $3}')
+
+# Build with PyInstaller
```
skirpichev: Redundant

mpmath/tests/test_version_frozen.sh
```diff
@@ -0,0 +1,66 @@
...
+echo "Building version_script with PyInstaller..."
+pip install --quiet pyinstaller
+pyinstaller --onefile --clean version_script.py
+
+# Run the frozen executable and capture output
```
skirpichev: Useless.

.github/workflows/test.yml
```diff
@@ -12,6 +12,18 @@ jobs:
     uses: ./.github/workflows/coverage.yml
   docs:
     uses: ./.github/workflows/docs.yml
+  frozen-version:
+    runs-on: ubuntu-24.04
+    steps:
+    - uses: actions/checkout@v6
+      with:
+        fetch-depth: 0
+    - name: Set up Python (latest)
```
skirpichev: Latest stable.  I suggest just drop this name.  (Edit: I meant the "name" field.)

skirpichev (changes requested):
Please see some comments.

Test is quick enough, so please also add it to the "needs" list of the tests job.

mpmath/tests/test_version_frozen.sh
FlurinArner: You mean just "the comment is redundant", correct?

skirpichev: Yes.  Right below we see: "echo "Installing mpmath from tarball...""

mpmath/tests/test_version_frozen.sh
FlurinArner: I added a commentary, but actually the `source .venv/bin/activate` works only specifically with `bash`. 

There are alternatives for `fish` or `csh`, but I couldn't find how to get it to work in pure `sh`.

.github/workflows/test.yml
FlurinArner: Let us try directly with `- uses: actions/setup-python@v6` then 👍

mpmath/tests/test_version_frozen.sh
```diff
@@ -1,32 +1,38 @@
...
 echo "Cleaning up previous test artifacts..."
-rm -rf test_local build dist *.spec .venv
+rm -rf build dist *.spec .venv
 
-mkdir -p test_local
-cd test_local
+# Get the directory of the current repository
+MPMATH_DIR=$(pwd)
+# # Working directory independent version: 
+# MPMATH_DIR=$(realpath "$(dirname "$(realpath "$0")")/../..")
```
FlurinArner: @skirpichev Let me know if you prefer the slighly complexer version (but safer), instead of using `pwd`

mpmath/tests/test_version_frozen.sh
```diff
@@ -0,0 +1,67 @@
+#!/bin/bash
+# Test that the version number is provided correctly in a frozen (bundled) executable (see #1044).
+set -e
+
+# Clean up any previous test artifacts
```
skirpichev:
```suggestion
```

mpmath/tests/test_version_frozen.sh
```diff
@@ -0,0 +1,67 @@
...
+echo "Repo mpmath directory: $MPMATH_DIR"
+TEMP_DIR=$(mktemp -d)
+echo "Created temporary directory: $TEMP_DIR"
+cd "$TEMP_DIR"
+
+echo "Creating virtual environment..."
+python3 -m venv .venv
+source .venv/bin/activate
+
+
```
skirpichev:
```suggestion
```

mpmath/tests/test_version_frozen.sh
```diff
@@ -0,0 +1,67 @@
...
+cd "$MPMATH_DIR"
+echo "Building tarball from local repo..."
+# Build the source distribution
```
skirpichev:
```suggestion
echo "Building the source distribution from local repo..."
```

mpmath/tests/test_version_frozen.sh
```diff
@@ -0,0 +1,67 @@
...
+
+# Find and install the generated tarball
+TARBALL=$(ls -t dist/*.tar.gz | head -1)
+echo "Generated tarball: $TARBALL"
+
+echo "Installing mpmath from tarball..."
+pip install dist/"$(basename $TARBALL)"
+cd "$TEMP_DIR" # Return to temporary folder to run the version script
+
+
```
skirpichev:
```suggestion
```

mpmath/tests/test_version_frozen.sh
```diff
@@ -0,0 +1,67 @@
...
+# Create version_script.py that imports mpmath and prints the package version
+cat << EOF > version_script.py
+import mpmath
+print("mpmath version:", mpmath.__version__)
```
skirpichev:
```suggestion
print(mpmath.__version__)
```

mpmath/tests/test_version_frozen.sh
```diff
@@ -0,0 +1,67 @@
...
+set -e
+
+# Clean up any previous test artifacts
+echo "Cleaning up previous test artifacts..."
+rm -rf build dist *.spec .venv
+
+# Get the directory of the current repository
+MPMATH_DIR=$(pwd)
+# # Working directory independent version: 
+# MPMATH_DIR=$(realpath "$(dirname "$(realpath "$0")")/../..")
```
skirpichev: Assuming bash, you can use pushd/popd and get rid of this variable.  Though, I don't see why you need one.

mpmath/tests/test_version_frozen.sh
```diff
@@ -0,0 +1,67 @@
...
+DIRECT_VERSION=$(python -m mpmath --version)
+
+echo "Building version_script with PyInstaller..."
+pyinstaller --onefile --clean version_script.py
+
+echo "Running frozen executable..."
+./dist/version_script 2>&1 | tee output.txt
+
+# Extract the version from the output
+FROZEN_VERSION=$(grep "mpmath version:" output.txt | awk '{print $3}')
```
skirpichev:
```suggestion
echo "Run frozen executable and extract the version from the output"
FROZEN_VERSION=$(./dist/version_script 2>&1)
```

mpmath/tests/test_version_frozen.sh
```diff
@@ -0,0 +1,67 @@
...
+
+# Return to the original repository directory
+cd "$MPMATH_DIR" 
+deactivate
```
skirpichev: In principle, it's useless.

mpmath/tests/test_version_frozen.sh
```diff
@@ -0,0 +1,67 @@
+#!/bin/bash
+# Test that the version number is provided correctly in a frozen (bundled) executable (see #1044).
+set -e
+
+# Clean up any previous test artifacts
+echo "Cleaning up previous test artifacts..."
+rm -rf build dist *.spec .venv
```
skirpichev: *.spec ?

mpmath/tests/test_version_frozen.sh
skirpichev: You are right, this version is better.  But I don't think you need a separate variable for this.

mpmath/tests/test_version_frozen.sh
FlurinArner: Ah, this is also from `pyinstaller` and was in the `temp_folder`.. Actually this didn't work, since it was generated inside that folder.

Removed

mpmath/tests/test_version_frozen.sh
FlurinArner: I kept the `MPMATH_DIR` variable definition, because `realpath $0` returns `No such file or directory` after performing a `cd` to a different directory.

mpmath/tests/test_version_frozen.sh
FlurinArner: Ah, you're saying here that I should remove `deactivate`, because nobody will be using the script in a manual way and the shell session will be closed once the test completes?

skirpichev (approved):
I did some final edits, instead of trying to do yet another review turn, sorry.

Now looks good for me.

FlurinArner:
> I did some final edits, instead of trying to do yet another review turn, sorry.

That's good for me! Thanks for the positive feedback and quick resolution. 

Just to be sure (since [you moved](https://github.com/mpmath/mpmath/pull/1055/changes/54f2aa705fb06f3d2919017634614b3a03e5d3ec#diff-2688cec3c1b113e004a5f57e3104f1fbe1768cbd6bff9c5d32daf7b96d95a00eR25) the `venv` creation and activation to **after** installing the tarball), I re-tested the `test_version_frozen.sh` against `v1.4.0` and can confirm that it would have failed. Now on `master` (after #1048) the test succeeds. :+1:

PR #1058, opened 2026-03-04
Drop support for CPython 3.9

skirpichev:
(Following https://github.com/sympy/sympy/pull/28816)


skirpichev:
BTW, I would prefer to drop also support for 3.10, following NEP 29.  Based on previous history timeline, v1.5 will be released after EOL for the CPython 3.10.

But it seems, SymPy prefers more conservative policy.

PR #1066, opened 2026-03-15
Add executable bit to test_version_frozen.sh

d-torrance:
This fixes the "[script-not-executable](https://udd.debian.org/lintian-tag/script-not-executable)" warning in the Debian package.


skirpichev:
I don't see a real problem here.  Though, lets at least *use* this bit in CI.

skirpichev (approved):

PR #1083, opened 2026-05-17
Fix qr_solve() failure on well-conditioned matrices with zero pivot

### LINKED ISSUE #983: qr_solve(): unexpected failure on 4x4 system
Attempting to solve the following system of equations using `qr_solve(A, b)` results in an error, although it appears the solution should exist:

$$\mathbf{A} = \begin{bmatrix} 
1 & -π/20 & (-π/20)^2 & (-π/20)^3 \\
1 & 0 & 0 & 0 \\
1 & π/20 & (π/20)^2 & (π/20)^3 \\
1 & π/10 & (π/10)^2 & (π/10)^3 \\
\end{bmatrix}$$

$$\mathbf{x} =  \begin{bmatrix} 
sin(-π/20) \\ 
0 \\ 
sin(π/20) \\ 
sin(π/20) 
\end{bmatrix}$$

Reproducing code example, `cpython` version = 3.12.1 `mpmath` version = 1.3.0:
```
from mpmath import mp

A = mp.matrix([[mp.one, -mp.pi / 20, (-mp.pi / 20) ** 2, (-mp.pi / 20) ** 3],
               [mp.one, mp.zero, mp.zero, mp.zero],
               [mp.one, mp.pi / 20, (mp.pi / 20) ** 2, (mp.pi / 20) ** 3],
               [mp.one, mp.pi / 10, (mp.pi / 10) ** 2, (mp.pi / 10) ** 3]])

b = mp.matrix([[mp.sin(-mp.pi / 20)],
               [mp.zero],
               [mp.sin(mp.pi / 20)],
               [mp.sin(mp.pi / 20)]])

# Uncommenting the line below allows the QR householder method to work. I
# think this is because it allows a slight non-zero result to appear so
# execution can continue. Any 'dps' value greater than 10 causes
# failure.
#
# mp.dps = 10

x_qr, residual_qr = mp.qr_solve(A, b)  # <- Gives error.

print(f"\nSolution using QR decomposition x_qr:\n{x_qr}")
print(f"Residual Ax_qr - b:\n{residual_qr}")
```

Error message:
```
  File "...\mpmath_issue.py", line 48, in <module>
    x_qr, residual_qr = mp.qr_solve(A, b)
                        ^^^^^^^^^^^^^^^^^
  File "...\mpmath\matrices\linalg.py", line 407, in qr_solve
    H, p, x, r = ctx.householder(ctx.extend(A, b))
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "...\mpmath\matrices\linalg.py", line 341, in householder
    raise ValueError('matrix is numerically singular')
ValueError: matrix is numerically singular
```

By comparison, `lu_solve(A, b)` has no problem:
```
print(f"Condition number of A: {mp.cond(A)}")

# Solution using LU decomposition -> Works fine.
x_lu = mp.lu_solve(A, b)
residual_lu = A * x_lu - b
print(f"\nSolution using LU decomposition x_lu:\n{x_lu}")
print(f"\nResidual Ax_lu - b:\n{residual_lu}")
```

Gives result:
```
Condition number of A: 694.870840206284

Solution using LU decomposition x_lu:
[             0.0]
[1.16187485778415]
[             0.0]
[-6.7270020477122]

Residual Ax_lu - b:
[0.0]
[0.0]
[0.0]
[0.0]
```

Cross-checking with *Wolfram|alpha* confirms the system is invertible / solvable, and lines up with the `lu_solve()` solution:

$$\mathbf{A}^{-1} = \begin{bmatrix} 
0 & 1 & 0 & 0 \\
-20/(3π) & -10/π  & 20/π & -10/(3π) \\
200/π^2 & -400/π^2 & 200/π^2 & 0 \\
-4000/(3π^3) & 4000/π^3 & -4000/π^3 & 4000/(3π^3)
\end{bmatrix}$$

$$\mathbf{x} = \mathbf{A}^{-1} \mathbf{b} = \begin{bmatrix} 
0 \\
70 sin(π/20)/(3π) \\
0 \\
-4000 sin(π/20)/(3π^3)
\end{bmatrix}$$

I believe the problem is likely this line in `householder(ctx, A)` in [mpmath/matrices/linalg.py](https://github.com/mpmath/mpmath/blob/3788ac73204c0b75662c746208e2114184b7dba4/mpmath/matrices/linalg.py#L333):
```
if not abs(s) > ctx.eps:  # <- Problem area?
    raise ValueError('matrix is numerically singular')
```
I suspect this check might be in the wrong location... it may not be required here at all?  From memory this should check that if the sum is zero, then nothing further needs to be done with that particular column?

JamBalaya56562:
Fixes mpmath/mpmath#983.

## Reproduction

Independent reproduction / analysis: https://aletheia-works.github.io/vivarium/repro/mpmath/983/

```python
from mpmath import mp, pi

A = mp.matrix([[1, -pi/20, (-pi/20)**2, (-pi/20)**3],
               [1, 0, 0, 0],
               [1, pi/20, (pi/20)**2, (pi/20)**3],
               [1, pi/10, (pi/10)**2, (pi/10)**3]])
b = mp.matrix([[mp.sin(-pi/20)], [0],
               [mp.sin(pi/20)], [mp.sin(pi/20)]])

mp.qr_solve(A, b)   # before: ValueError: matrix is numerically singular
                    # after:  solves correctly (matches lu_solve)
```

The matrix is well-conditioned (cond ≈ 695, det ≈ 1.8e-4) and `lu_solve()` handles it without issue.

## Root cause

The issue reporter suspected the singularity guard `if not abs(s) > ctx.eps` in `householder()`. That guard is only the *symptom* — the actual bug is one line below it, in the sign convention used to pick the Householder reflection:

```python
p.append(-ctx.sign(ctx.re(A[j,j])) * ctx.sqrt(s))
```

When `A[j,j]` is exactly zero, `ctx.sign(0) == 0`, so `p[j]` collapses to zero. Then:

```python
kappa = ctx.one / (s - p[j] * A[j,j])   # = 1/s instead of 2/||v||²
```

The Householder reflection is computed with the wrong scaling, the trailing submatrix is corrupted, and a few iterations later the now-noisy column trips the "numerically singular" check.

The issue's matrix triggers this because row 1 is `[1, 0, 0, 0]`: after the j=0 step the second pivot `A[1,1]` lands on exact zero.

## Fix

Default `sign` to `+1` when `Re(A[j,j])` is zero. This matches the convention used by LAPACK's `dlarfg` and other standard QR implementations.

Verified: with the fix, `householder()` produces `p = [-2, -0.351, 0.0493, -0.0052]`, matching `numpy.linalg.qr` exactly.

## Tests

- Added `test_qr_solve_issue_983` using the exact matrix from the issue, asserting `||A x − b||_∞ < 1e-14`.
- All 19 existing tests in `mpmath/tests/test_linalg.py` (including `test_householder`, `test_solve`, `test_singular`) continue to pass.

### DIFF
--- a/mpmath/tests/test_linalg.py
+++ b/mpmath/tests/test_linalg.py
@@ -201,6 +201,18 @@ def test_solve_overdet_complex():
     b = matrix([1 + j, 2, -j])
     assert norm(residual(A, lu_solve(A, b), b)) < 1.0208
 
+def test_qr_solve_issue_983():
+    A = matrix([[mp.one, -mp.pi / 20, (-mp.pi / 20) ** 2, (-mp.pi / 20) ** 3],
+                [mp.one, mp.zero, mp.zero, mp.zero],
+                [mp.one, mp.pi / 20, (mp.pi / 20) ** 2, (mp.pi / 20) ** 3],
+                [mp.one, mp.pi / 10, (mp.pi / 10) ** 2, (mp.pi / 10) ** 3]])
+    b = matrix([[mp.sin(-mp.pi / 20)],
+                [mp.zero],
+                [mp.sin(mp.pi / 20)],
+                [mp.sin(mp.pi / 20)]])
+    x, _ = qr_solve(A, b)
+    assert norm(residual(A, x, b), inf) < 1e-14
+
 def test_singular():
     A = [[5.6, 1.2], [7./15, .1]]
     B = repr(zeros(2))
--- a/mpmath/tests/test_linalg.py
+++ b/mpmath/tests/test_linalg.py
@@ -202,14 +202,14 @@ def test_solve_overdet_complex():
     assert norm(residual(A, lu_solve(A, b), b)) < 1.0208
 
 def test_qr_solve_issue_983():
-    A = matrix([[mp.one, -mp.pi / 20, (-mp.pi / 20) ** 2, (-mp.pi / 20) ** 3],
-                [mp.one, mp.zero, mp.zero, mp.zero],
-                [mp.one, mp.pi / 20, (mp.pi / 20) ** 2, (mp.pi / 20) ** 3],
-                [mp.one, mp.pi / 10, (mp.pi / 10) ** 2, (mp.pi / 10) ** 3]])
-    b = matrix([[mp.sin(-mp.pi / 20)],
-                [mp.zero],
-                [mp.sin(mp.pi / 20)],
-                [mp.sin(mp.pi / 20)]])
+    A = matrix([[1, -pi/20, (-pi/20)**2, (-pi/20)**3],
+                [1, 0, 0, 0],
+                [1, pi / 20, (pi/20)**2, (pi/20)**3],
+                [1, pi/10, (pi/10)**2, (pi/10)**3]])
+    b = matrix([[mp.sin(-pi/20)],
+                [0],
+                [mp.sin(pi/20)],
+                [mp.sin(pi/20)]])
     x, _ = qr_solve(A, b)
     assert norm(residual(A, x, b), inf) < 1e-14
 
--- a/mpmath/matrices/linalg.py
+++ b/mpmath/matrices/linalg.py
@@ -371,7 +371,10 @@ class LinearAlgebraMethods:
             s = ctx.fsum(abs(A[i,j])**2 for i in range(j, m))
             if not abs(s) > ctx.eps:
                 raise ValueError('matrix is numerically singular')
-            p.append(-ctx.sign(ctx.re(A[j,j])) * ctx.sqrt(s))
+            sign = ctx.sign(ctx.re(A[j,j]))
+            if sign == 0:
+                sign = ctx.one
+            p.append(-sign * ctx.sqrt(s))
             kappa = ctx.one / (s - p[j] * A[j,j])
             A[j,j] -= p[j]
             for k in range(j+1, n):


mpmath/tests/test_linalg.py
```diff
@@ -201,6 +201,18 @@ def test_solve_overdet_complex():
...
 
+def test_qr_solve_issue_983():
+    A = matrix([[mp.one, -mp.pi / 20, (-mp.pi / 20) ** 2, (-mp.pi / 20) ** 3],
+                [mp.one, mp.zero, mp.zero, mp.zero],
+                [mp.one, mp.pi / 20, (mp.pi / 20) ** 2, (mp.pi / 20) ** 3],
+                [mp.one, mp.pi / 10, (mp.pi / 10) ** 2, (mp.pi / 10) ** 3]])
+    b = matrix([[mp.sin(-mp.pi / 20)],
+                [mp.zero],
+                [mp.sin(mp.pi / 20)],
+                [mp.sin(mp.pi / 20)]])
```
skirpichev:
```suggestion
    A = matrix([[1, -pi/20, (-pi/20)**2, (-pi/20)**3],
                [1, 0, 0, 0],
                [1, pi / 20, (pi/20)**2, (pi/20)**3],
                [1, pi/10, (pi/10)**2, (pi/10)**3]])
    b = matrix([[mp.sin(-pi/20)],
                [0],
                [mp.sin(pi/20)],
                [mp.sin(pi/20)]])
```

JamBalaya56562: Done in beae0aa — applied the suggestion verbatim and updated the PR description's reproduction snippet to match the simplified style. All 19 tests in test_linalg.py still pass.

skirpichev (approved):

PR #1087, opened 2026-05-24
Clarify to_float() docstring

### LINKED ISSUE #1085: Change in to_float overflow handling
After 87cfbc42 the `to_float` function will convert large mpfs to float_max rather than inf:

Test code:
```python
import sys
from mpmath import mp, mpf
from mpmath.libmp import to_float, round_down, round_nearest

mp.dps = 1000
x = mpf("1e500")

print(float(x))
print(to_float(x._mpf_))
print(to_float(x._mpf_, rnd=round_down))
print(to_float(x._mpf_, rnd=round_nearest))
print(sys.float_info.max)
```
With mpmath 1.4.0:
```
inf
inf
inf
inf
1.7976931348623157e+308
```
With current mpmath master:
```
inf
1.7976931348623157e+308
1.7976931348623157e+308
inf
1.7976931348623157e+308
```
The docstring for `to_float` still says:

> If the magnitude of number is too large to represent as a regular float, it will be converted to infinity.

That makes me think that this change is unintentional.

skirpichev:
Second paragraph rephrased to:
> If the magnitude of *rounded* number is too large to represent as
> a regular float, it will be converted to infinity.

Closes #1085

### DIFF
--- a/mpmath/libmp/libmpf.py
+++ b/mpmath/libmp/libmpf.py
@@ -339,15 +339,13 @@ def from_Decimal(x, prec=0, rnd=round_fast):
 
 def to_float(s, strict=False, rnd=round_fast):
     """
-    Convert a raw mpf to a Python float.  The result is
-    exact if s.bit_length() <= sys.float_info.mant_dig
-    and no underflow/overflow occurs.  Else result is
-    correctly rounded.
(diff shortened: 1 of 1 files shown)


skirpichev:
CC @oscarbenjamin

PR #1092, opened 2026-06-01
Raise ValueError when same sign at interval boundaries in bisection rootfinding algorithm

baranwalayush:
While working on #1075, I came across this TODO in the codebase. 
- Implemented the requirement
- Added unit test for the same

### DIFF
--- a/mpmath/calculus/optimization.py
+++ b/mpmath/calculus/optimization.py
@@ -283,7 +283,6 @@ class Muller:
             error = abs(x2 - x1)
             yield x2, error
 
-# TODO: consider raising a ValueError when there's no sign change in a and b
 class Bisection:
     """
     1d-solver generating pairs of approximative root and error.
@@ -311,13 +310,19 @@ class Bisection:
         self.b = x0[1]
 
     def __iter__(self):
+        ctx = self.ctx
         f = self.f
         a = self.a
         b = self.b
         l = b - a
+        fa = f(a)
         fb = f(b)
+
+        if ctx.sign(fa) == ctx.sign(fb):
+            raise ValueError("Function must have opposite signs at interval boundaries.")
+        
         while True:
-            m = self.ctx.ldexp(a + b, -1)
+            m = ctx.ldexp(a + b, -1)
             fm = f(m)
             sign = fm * fb
             if sign < 0:
@@ -326,7 +331,7 @@ class Bisection:
                 b = m
                 fb = fm
             else:
-                yield m, self.ctx.zero
+                yield m, ctx.zero
             l /= 2
             yield (a + b)/2, abs(l)
 
--- a/mpmath/calculus/optimization.py
+++ b/mpmath/calculus/optimization.py
@@ -318,7 +318,7 @@ class Bisection:
         fa = f(a)
         fb = f(b)
 
-        if ctx.sign(fa) == ctx.sign(fb):
+        if fa*fb > 0:
             raise ValueError("Function must have opposite signs at interval boundaries.")
 
         while True:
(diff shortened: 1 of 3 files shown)


skirpichev:
Fails CI.

mpmath/calculus/optimization.py
```diff
@@ -311,13 +310,19 @@ def __init__(self, ctx, f, x0, **kwargs):
...
     def __iter__(self):
+        ctx = self.ctx
         f = self.f
         a = self.a
         b = self.b
         l = b - a
+        fa = f(a)
         fb = f(b)
+
+        if ctx.sign(fa) == ctx.sign(fb):
```
skirpichev:
```suggestion
        if f(a)*f(b) > 0:
```
?

mpmath/tests/test_rootfinding.py
```diff
@@ -50,6 +50,10 @@ def test_bisection():
     # issue 273
     assert findroot(lambda x: x**2-1,(0,2),solver='bisect') == 1
 
+def test_bisection2():
+    with pytest.raises(ValueError):
+        findroot(lambda x: x**2-1,(4,2),solver='bisect') == 1
```
skirpichev:
```suggestion
        findroot(lambda x: x**2-1, (4, 2), solver='bisect') == 1
```

mpmath/function_docs.py
```diff
@@ -7092,9 +7092,10 @@
 investigating the zeros of the Riemann zeta function.
 For example, one can use a root-finding algorithm based
 on sign changes::
-
-    >>> findroot(siegelz, [100, 200], solver='bisect')
-    176.4414342977104188888926
+    >>> g10 = grampoint(10)
+    >>> g11 = grampoint(11)
+    >>> findroot(siegelz, [g10, g11], solver='bisect')
+    56.44624769706339480436776
```
skirpichev: That looks unrelated to change.

baranwalayush: The CI test was failing earlier because of this `findroot(siegelz, [100, 200], solver='bisect')` line, as the function has same sign at both the boundaries. So, I changed it to not throw the ValueError.

skirpichev: Then at least use appropriate boundaries to keep old root.

skirpichev (commented):
I'm not sure about this.  TODO says "consider".

Should we filter out possibility of using this method for intervals with multiple roots?

baranwalayush:
> Should we filter out possibility of using this method for intervals with multiple roots?

I think yes, because in case of `fa*fb > 0`, we might filter the possibility of having even number of roots in between the intervals, but not having it breaks the logic of bisection method. If the algorithm doesn't explicitly check for valid brackets, it can easily zoom into a sub-interval that contains absolutely no roots.

mpmath/function_docs.py
```diff
@@ -7092,8 +7092,7 @@
 investigating the zeros of the Riemann zeta function.
 For example, one can use a root-finding algorithm based
 on sign changes::
-
-    >>> findroot(siegelz, [100, 200], solver='bisect')
+    >>> findroot(siegelz, [176, 177], solver='bisect')
```
skirpichev, line 7096:
```suggestion

    >>> findroot(siegelz, [176, 177], solver='bisect')
```

Lets keep the space.

mpmath/tests/test_rootfinding.py
```diff
@@ -50,6 +50,10 @@ def test_bisection():
     # issue 273
     assert findroot(lambda x: x**2-1,(0,2),solver='bisect') == 1
 
+def test_bisection2():
```
skirpichev:
```suggestion
```
Why extra function?

skirpichev (approved):

skirpichev:
Thank you.

PR #1100, opened 2026-06-14
Implement direct series for lerchphi() base case with |z| < 0.5

### LINKED ISSUE #1099: lerchphi loses hundreds of digits for 0<z<<1, complex a: error plateaus across wide dps ranges (e.g. 2.7e-127 for dps=300..500)
## Summary

For small real `z` in (0,1) and complex `a = 1/4 + i*t` with `t` roughly in 10..50,
`lerchphi(z, 2, a)` returns values whose error stays **hundreds of digits above the working
precision**, and the error is locally *stable under precision increase*: it plateaus over wide
`dps` ranges (e.g. frozen at 2.7e-127 for every `dps` from 300 to 500). At `dps=500` the
returned value has only ~124 correct digits.

In this regime the defining series Φ(z,s,a) = Σ_{k≥0} z^k/(a+k)^s converges geometrically
(|z| ≈ 3.6e-4 → ~3.4 digits per term), so the reference value is trivially computable to any
precision using exact arithmetic only.

## Minimal reproducible example

```python
import mpmath as mp

def lerch_series(z, s, a, dps):
    """Defining series with 20 guard digits; exact for |z|<1 (geometric tail)."""
    old = mp.mp.dps
    mp.mp.dps = dps + 20
    K = int(dps * mp.log(10) / mp.log(1 / abs(z))) + 10
    acc = mp.mpf(0); zk = mp.mpf(1)
    for k in range(K + 1):
        acc += zk / (a + k) ** s
        zk *= z
    mp.mp.dps = old
    return +acc

z = mp.mpf(1) / 2809
for dps in [50, 100, 150, 200, 250, 300, 350, 400, 500, 600]:
    mp.mp.dps = dps
    a = mp.mpc(mp.mpf(1) / 4, mp.pi * 32 / mp.log(53))   # a ≈ 0.25 + 25.32j
    lib = mp.lerchphi(z, 2, a)
    ref = lerch_series(z, 2, a, dps)
    print(dps, mp.nstr(abs(lib - ref), 4))
```

Output (mpmath 1.3.0, Python 3.13, macOS arm64):

```
 50  0.0
100  3.931e-82
150  4.737e-97
200  4.737e-97     <-- plateau: identical error for 150..250
250  4.737e-97
300  2.722e-127
350  2.722e-127
400  2.722e-127    <-- plateau: identical error for 300..500
500  2.722e-127
600  4.0e-188      <-- still ~412 digits short of requested precision
```

A second instance, same signature: `z = 1/10000`, `a = 1/4 + i*pi*37/log(100)` →
error frozen at 3.657e-127 for dps=300 and dps=400.

## Dependence on Im(a) (dps=300, z=1/2809, a=1/4+i·t)

```
t =  0.79 : 0.0
t =  6.33 : 0.0
t = 12.66 : 2.709e-167
t = 18.99 : 5.338e-134
t = 25.32 : 2.722e-127   <-- worst band
t = 31.65 : 9.037e-131
t = 47.48 : 1.672e-156
t = 79.13 : 3.252e-222
t = 110.8 : 4.183e-306
```

## Expected behavior

Error ~ 10^-dps. The parameters are inside the textbook domain of the defining series
(|z| < 1, Re(a) > 0) — no analytic continuation is involved.

## Why this is nasty in practice

Because the error plateaus across wide precision ranges, the standard validation idiom
"recompute at higher dps and compare" reports full agreement (e.g. dps=300 vs dps=400
match to all shown digits) while both values are wrong beyond ~124 digits. We hit this in
a high-precision spectral computation where the e-127 contamination of matrix entries
produced spurious tiny negative eigenvalues of a positive matrix; only a comparison
against the defining series exposed it.

## Possibly related

#252 reports branch/continuity problems of lerchphi for complex `a` (different regime,
z > 1, O(1) errors). The docs state lerchphi forces Re(a) > 0 via a recurrence and then
uses an integral representation; the plateau structure here looks like the quadrature
degree tracking dps in steps while the integrand's accuracy is limited by something
else (parameter-dependent, worst around Im(a) ≈ 25 for these z). For |z| << 1 a direct
series evaluation would be both correct and faster.

skirpichev:
Closes #1099

### DIFF
--- a/mpmath/functions/zeta.py
+++ b/mpmath/functions/zeta.py
@@ -1162,6 +1162,8 @@ def lerchphi(ctx, z, s, a):
             v += zpow / (a+n)**s
             zpow *= z
         return zpow * ctx.lerchphi(z,s, a+m) + v
+    if abs(z) < 0.5:
+        return ctx.nsum(lambda k: z**k/(a+k)**s, [0, ctx.inf])
     g = ctx.ln(z)
     v = 1/(2*a**s) + ctx.gammainc(1-s, -a*g) * (-g)**(s-1) / z**a
     h = s / 2
--- a/mpmath/tests/test_functions2.py
+++ b/mpmath/tests/test_functions2.py
@@ -7,15 +7,16 @@ from mpmath import (agm, airyai, airybi, appellf1, bei, ber, besseli, besselj,
                     besseljzero, besselk, bessely, besselyzero, betainc,
                     chebyt, chebyu, chi, ci, clsin, convert, coulombg, e, e1,
                     ei, ellipe, ellipk, eps, erf, erfc, erfi, erfinv, exp,
-                    expint, fadd, fmul, foxh, fp, fraction, fresnelc, fresnels,
-                    fsub, fsum, gamma, gammainc, gegenbauer, hankel1, hankel2,
-                    hermite, hyp0f1, hyp1f1, hyp1f2, hyp2f0, hyp2f1, hyp2f2,
-                    hyp2f3, hyper, hypercomb, hyperu, inf, isnan, j, j0, j1,
-                    jacobi, kei, ker, laguerre, lambertw, ldexp, legendre,
-                    legenp, legenq, lerchphi, li, log, lower_gamma, meijerg,
-                    mp, mpc, mpf, nan, ncdf, npdf, nthroot, pi, polylog, qp,
-                    quadts, shi, si, spherharm, spherical_jn, spherical_yn,
-                    sqrt, struveh, struvel, upper_gamma, whitm, whitw, zeta)
+                    expint, extradps, fadd, fmul, foxh, fp, fraction, fresnelc,
+                    fresnels, fsub, fsum, gamma, gammainc, gegenbauer, hankel1,
+                    hankel2, hermite, hyp0f1, hyp1f1, hyp1f2, hyp2f0, hyp2f1,
+                    hyp2f2, hyp2f3, hyper, hypercomb, hyperu, inf, isnan, j,
+                    j0, j1, jacobi, kei, ker, laguerre, lambertw, ldexp,
+                    legendre, legenp, legenq, lerchphi, li, log, lower_gamma,
+                    meijerg, mp, mpc, mpf, nan, ncdf, npdf, nthroot, pi,
+                    polylog, qp, quadts, shi, si, spherharm, spherical_jn,
+                    spherical_yn, sqrt, struveh, struvel, upper_gamma, whitm,
+                    whitw, zeta)
 from mpmath.libmp import BACKEND, NoConvergence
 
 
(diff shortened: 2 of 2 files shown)


fredrik-johansson:
Have you checked how this performs for |z| very close to 1?

skirpichev:
> Have you checked how this performs for |z| very close to 1?

Yes, quick tests suggest that quad() performs worse, e.g.:

```
In [1]: mp.dps=200

In [2]: z = 1e-8

In [3]: a = mpc(mpf(1)/4, pi*32/log(53))

In [4]: %time r1 = lerchphi(z, 2, a)
CPU times: user 23.4 s, sys: 37.3 ms, total: 23.4 s
Wall time: 23.5 s
```
on the master vs:
```
In [4]: %time r1 = lerchphi(z, 2, a)
CPU times: user 25.8 ms, sys: 0 ns, total: 25.8 ms
Wall time: 26 ms
```
on the branch.

fredrik-johansson:
Your z is very close to 0. What about 0.9999 or 0.9999 * exp(i*theta) for various theta.

fredrik-johansson:
I expect `polylog_general` to do well for `z` close to 1 since it uses a power series expansion in `log(z)`, while `nsum` should perform well for `z` close to -1.

skirpichev:
> Your z is very close to 0.

Huh, indeed.  Typo turned into a wrong claim.  Thanks for catching.

> I expect polylog_general to do well for z close to 1 since it uses a power series expansion in log(z), while nsum should perform well for z close to -1.

I'm thinking about something like `|z|<0.5`.  The nsum() here already better:
```
In [1]: mp.dps=200; z=1-0.5; a = mpc(mpf(1)/4, pi*32/log(53))

In [2]: %time r1 = lerchphi(z, 2, a)
CPU times: user 4.16 s, sys: 140 ms, total: 4.3 s
Wall time: 4.4 s

In [3]: %time r1 = lerchphi(-z, 2, a)
CPU times: user 2.3 s, sys: 8.53 ms, total: 2.31 s
Wall time: 2.31 s
```
vs
```
In [2]: %time r1 = lerchphi(z, 2, a)
CPU times: user 13.7 s, sys: 69.9 ms, total: 13.8 s
Wall time: 13.8 s

In [3]: %time r1 = lerchphi(-z, 2, a)
CPU times: user 20.9 s, sys: 8.22 ms, total: 20.9 s
Wall time: 20.9 s
```

skirpichev:
@fredrik-johansson, are you ok with modification?

skirpichev:
Ok, lets close this.  Count silence as a negative review.

skirpichev:
Actually, no.  Looking at FLINT sources, it seems that similar strategy is used, with bound 0.875 (vs 0.5).

fredrik-johansson:
Sorry; yes, this looks fine.

BTW, when the series is not applicable one could also use the algorithm `lerch_hankel` from https://fredrikj.net/blog/2022/02/computing-the-lerch-transcendent/ which I believe avoids some problems with the present integration fallback used in mpmath.

skirpichev:
> Sorry; yes, this looks fine.

Ok.

I guess, that "direct" method will perform better and allows to increase the bound.  Though, maybe not as in the FLINT.

> when the series is not applicable one could also use the algorithm lerch_hankel from https://fredrikj.net/blog/2022/02/computing-the-lerch-transcendent/ 

Thanks.  I'll try this on related issue with lerchphi.  I was assuming, that acb_dirichlet_lerch_phi_integral() uses similar algorithm, as mpmath.

PR #1103, opened 2026-06-17
Add Brent root-finding algorithm

### LINKED ISSUE #1095: Add Brent root-finding algorithm
This will address following TODO:
https://github.com/mpmath/mpmath/blob/e10416f9971cb4ab450f9b26e18a9b12e9878dc9/mpmath/calculus/optimization.py#L575

Note, there is semi-finished pr:
https://github.com/mpmath/mpmath/pull/999

baranwalayush:
closes #1095 
- Implemented the Brent root-finding algorithm
- Followed mpmath implementation style
- Added unit tests for the same

Reference : https://en.wikipedia.org/wiki/Brent%27s_method

### DIFF
--- a/mpmath/calculus/optimization.py
+++ b/mpmath/calculus/optimization.py
@@ -672,10 +672,6 @@ class Brent:
                 a, b = b, a
                 fa, fb = fb, fa
 
-            # Check for exact root hit
-            if abs(fb) < self.tol:
-                yield b, abs(fb)
-
 class ModAB:
     """
     1d-solver generating pairs of approximative root and error.
--- a/mpmath/calculus/optimization.py
+++ b/mpmath/calculus/optimization.py
@@ -601,7 +601,6 @@ class Brent:
 
         self.f = f
         self.a, self.b = x0
-        self.verbose = kwargs['verbose']
         self.tol = kwargs['tol']
 
     def __iter__(self):
--- a/mpmath/calculus/optimization.py
+++ b/mpmath/calculus/optimization.py
@@ -613,11 +613,9 @@ class Brent:
         fa = f(a)
         fb = f(b)
 
-        # Check for initial bracketing
         if fa*fb > 0:
             raise ValueError("Function must have opposite signs at interval boundaries.")
 
-        # Enforce that b holds the closest estimate to the root
         if abs(fa) < abs(fb):
             a, b = b, a
             fa, fb = fb, fa
@@ -628,10 +626,9 @@ class Brent:
         mflag = True
 
         while True:
-            # Yield current best guess (b) and interval error boundary
+
             yield b, abs(b - a)
 
-            # Check if Inverse Quadratic Interpolation can be used
             if fa != fc and fb != fc:
                 # Inverse Quadratic Interpolation formula
                 s = (a * fb * fc) / ((fa - fb) * (fa - fc)) + \
@@ -661,12 +658,10 @@ class Brent:
 
             fs = f(s)
 
-            # Update tracking states
             d = c
             c = b
             fc = fb
 
-            # Retain bracketing properties
             if fa*fs < 0:
                 b = s
                 fb = fs
@@ -674,14 +669,13 @@ class Brent:
                 a = s
                 fa = fs
 
-            # Enforce that b holds the closest estimate to the root
             if abs(fa) < abs(fb):
                 a, b = b, a
                 fa, fb = fb, fa
 
             # Check for exact root hit
-            if fb == ctx.zero:
-                yield b, ctx.zero
+            if abs(fb) < self.tol:
+                yield b, abs(fb)
 
 class ModAB:
     """
--- a/mpmath/calculus/optimization.py
+++ b/mpmath/calculus/optimization.py
@@ -575,7 +575,113 @@ class ANewton:
                     print('accelerating convergence')
             yield x0, error
 
-# TODO: add Brent
+class Brent:
+    """
+    1d-solver generating pairs of approximative root and error.
+
+    Uses Brent's method to find a root of f in [a, b]. It combines
+    Bisection, the Secant method, and Inverse Quadratic Interpolation (IQI)
+    for robust and superlinear convergence.
+
+    Pro:
+    * Guaranteed to converge if a root is bracketed (like Bisection).
+    * Can converge much faster than Bisection on smooth functions.
+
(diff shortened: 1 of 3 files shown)


mpmath/calculus/optimization.py
```diff
@@ -575,7 +575,113 @@ def __iter__(self):
...
+                fa = fs
+
+            # Enforce that b holds the closest estimate to the root
+            if abs(fa) < abs(fb):
+                a, b = b, a
+                fa, fb = fb, fa
+
+            # Check for exact root hit
+            if fb == ctx.zero:
+                yield b, ctx.zero
```
skirpichev: Here you have error=0.  Does it make sense to continue?

BTW, maybe you exit if abs(fb) < self.tol instead?

mpmath/calculus/optimization.py
```diff
@@ -575,7 +575,113 @@ def __iter__(self):
...
+    def __iter__(self):
+        ctx = self.ctx
+        f = self.f
+
+        a = self.a
+        b = self.b
+        fa = f(a)
+        fb = f(b)
+
+        # Check for initial bracketing
```
skirpichev:
```suggestion
```

mpmath/calculus/optimization.py
```diff
@@ -575,7 +575,113 @@ def __iter__(self):
...
+        if fa*fb > 0:
+            raise ValueError("Function must have opposite signs at interval boundaries.")
+
+        # Enforce that b holds the closest estimate to the root
```
skirpichev:
```suggestion
```
Add instead a comment on variable b above.

mpmath/calculus/optimization.py
```diff
@@ -575,7 +575,113 @@ def __iter__(self):
...
+            a, b = b, a
+            fa, fb = fb, fa
+
+        c = a
+        fc = fa
+        d = c # will be assigned properly on the first interation
+        mflag = True
+
+        while True:
+            # Yield current best guess (b) and interval error boundary
```
skirpichev:
```suggestion
```

mpmath/calculus/optimization.py
```diff
@@ -575,7 +575,113 @@ def __iter__(self):
...
+            yield b, abs(b - a)
+
+            # Check if Inverse Quadratic Interpolation can be used
```
skirpichev:
```suggestion
```

mpmath/calculus/optimization.py
```diff
@@ -575,7 +575,113 @@ def __iter__(self):
...
+
+            if cond1 or cond2 or cond3 or cond4 or cond5:
+                s = ctx.ldexp(a + b, -1)
+                mflag = True
+            else:
+                mflag = False
+
+            fs = f(s)
+
+            # Update tracking states
```
skirpichev:
```suggestion
```

mpmath/calculus/optimization.py
```diff
@@ -575,7 +575,113 @@ def __iter__(self):
...
+
+            # Retain bracketing properties
+            if fa*fs < 0:
+                b = s
+                fb = fs
+            else:
+                a = s
+                fa = fs
+
+            # Enforce that b holds the closest estimate to the root
```
skirpichev:
```suggestion
```

mpmath/calculus/optimization.py
baranwalayush: either we continue or not is handled by the `findroot()` function itself, it compares the error and tolerance and returns if the tolerance is higher.

Ok, changed it to use `tol` instead of `ctx.zero`.

mpmath/calculus/optimization.py
```diff
@@ -575,7 +575,107 @@ def __iter__(self):
...
+            a, b = b, a
+            fa, fb = fb, fa
+
+        c = a
+        fc = fa
+        d = c # will be assigned properly on the first interation
+        mflag = True
+
+        while True:
+
```
skirpichev, line 628:
```suggestion
```

mpmath/calculus/optimization.py
```diff
@@ -575,7 +575,107 @@ def __iter__(self):
...
+    maxsteps = 100
+
+    def __init__(self, ctx, f, x0, **kwargs):
+        self.ctx = ctx
+        if len(x0) != 2:
+            raise ValueError('expected interval of 2 points, got %i' % len(x0))
+
+        self.f = f
+        self.a, self.b = x0
+        self.verbose = kwargs['verbose']
```
skirpichev: It's not used.

mpmath/calculus/optimization.py
```diff
@@ -575,7 +575,107 @@ def __iter__(self):
...
+                a = s
+                fa = fs
+
+            if abs(fa) < abs(fb):
+                a, b = b, a
+                fa, fb = fb, fa
+
+            # Check for exact root hit
+            if abs(fb) < self.tol:
+                yield b, abs(fb)
```
skirpichev: Should you continue in this case?

baranwalayush: No, we shouldn't, but adding a break statement would not add anything as this is already handled via `findroot()`.
```
        # use solver
        iterations = solver(ctx, f, x0, **kwargs)
        maxsteps = kwargs.get('maxsteps', iterations.maxsteps)
        i = 0
        for x, error in iterations:
            if verbose:
                print('x:    ', x)
                print('error:', error)
            i += 1
            if error < tol * max(1, norm(x)) or i >= maxsteps:
                break
```

baranwalayush: Please correct me if I'm wrong, the root-finding algorithms are always used with `findroot()` function and when to break the loop is already handled in it, so it doesn't matter whether we add another break statement in the algorithm's implementations. Besides, adding the `break` statement is causing the 100% test coverage failure, as the program is never reaching that part.

skirpichev: Yes, probably you are right.

mpmath/calculus/optimization.py
```diff
@@ -575,7 +575,106 @@ def __iter__(self):
...
```
skirpichev:
```suggestion
```

skirpichev (changes requested):
And please add new solver to the docs (docs/calculus/optimization.rst).

skirpichev (approved):

Issue #612, opened 2021-10-06
Weierstrass P-function


stla:
Hello,

  It would be nice to have the [Weierstrass P-function](https://en.wikipedia.org/wiki/Weierstrass_elliptic_function) in **mpmath**.

Hyrumveneno:
Agreed. Do you have a algorithm in mind to use?

stla:
It is implemented in the R package [elliptic](https://github.com/RobinHankin/elliptic/blob/master/R/elliptic.R#L1208). Does it help?

fredrik-johansson:
It can be computed using Jacobi theta functions: https://fungrim.org/entry/af0dfc/

Hyrumveneno:
> It is implemented in the R package [elliptic](https://github.com/RobinHankin/elliptic/blob/master/R/elliptic.R#L1208). Does it help?

Naw, I was just asking because if you know the algorithm, most of the time, it's not too hard to implement.
> It can be computed using Jacobi theta functions: https://fungrim.org/entry/af0dfc/

It looks like you could implement it if you wanted to.

stla:
I seriously believe this formula is wrong. This is much more complicated. This took me a long time but I've finally done an implementation. I compared the printed results with Wolfram and they are the same.

```python
from mpmath import jtheta, pi, exp, sqrt, polyroots, agm

def p_weierstrass_from_g2_g3(g2, g3):
    r1, r2, r3 = polyroots([4, 0, -g2, -g3])
    e3 = r3
    a = sqrt(r1 - r3)
    b = sqrt(r1 - r2)
    c = sqrt(r2 - r3)
    if abs(a + b) < abs(a - b):
        b *= 1j
    if abs(a + c) < abs(a - c):
        c *= 1j
    if abs(c + 1j*b) < abs(c - 1j*b):
        e3 = r1
        a *= 1j 
        tmp = b
        b = 1j * c
        c = 1j * tmp
    w1 = 1 / agm(a, b)
    w3 = 1j / agm(a, c)
    q = exp(1j * pi * w3/w1)
    return lambda z: (
        e3 + (pi * jtheta(2,0,q) * jtheta(3,0,q) * jtheta(4,z/w1,q)
              / (pi * w1 * jtheta(1,z/w1,q)))**2    
    )

z = 1+1j
g2 = 5 - 3j
g3 = 2 - 7j
print(p_weierstrass_from_g2_g3(g2, g3)(z))
[...]

Issue #1056, opened 2026-03-04
Functions should use explicit arguments (including keyword arguments) when possible


skirpichev:
For example:
```pycon
>>> inspect.signature(hyp2f1)
<Signature (a, b, c, z, **kwargs)>
```
should rather explicitly list valid keyword arguments like `eliminate`, `eliminate_all`, `force_series` and so on.

And I think we should declare such arguments as keyword-only, i.e.:
```py
def hyp2f1(a, b, c, z, *, eliminate=False):
    ...
```
Though, this will introduce a compatibility break, affecting people, who prefer unreadable code mess.

See also https://github.com/sympy/sympy/issues/14030 and https://github.com/sympy/sympy/issues/17944 for rationale.

skirpichev:
On top of https://github.com/mpmath/mpmath/pull/1124:
```pycon
>>> import inspect
... with_args = []  
... with_kwargs = []
... for n in dir(mpmath):
...     m = getattr(mpmath, n)
...     try:
...         s = inspect.signature(m)
...     except:
...         continue
...     if any(_.kind == inspect._ParameterKind.VAR_POSITIONAL for _ in s.parameters.values()):
...         with_args.append(n)
...     if any(_.kind == inspect._ParameterKind.VAR_KEYWORD for _ in s.parameters.values()):
...         with_kwargs.append(n)
... 
>>> with_args
['arange', 'ellipe', 'ellippi', 'invlapdehoog', 'invlapstehfest', 'invlaptalbot', 'linspace', 'matrix', 'nsum', 'ones', 'quad', 'quadgl', 'quadts', 'timing', 'zeros']
>>> with_kwargs
[...]

PR #1113, opened 2026-06-24
Add Weierstrass elliptic functions

HeskethGD:
Add Weierstrass wp, wpprime, sigma, zeta, and inverse wp functions with support for invariants, half-periods, and tau parameterizations.

Add conversions between Weierstrass invariants and half-periods, including special lemniscatic and equianharmonic cases. Add tests for differential identities, parameter roundtrips, Klein invariant consistency, cubic roots at half-periods, periodicity, inverse wp, and invalid parameterizations.

Document the new functions in the elliptic functions reference.

Based on (a slightly modified fork) of the pyweierstrass package by @stla / Stéphane Laurent
https://pyweierstrass.readthedocs.io/en/latest/ 
https://github.com/stla/pyweierstrass?tab=GPL-3.0-1-ov-file

All tests pass. In addition to tests in this suite, also confirmed it reproduces expected results when solving known nonlinear diff eqns in various notebooks by comparing against scipy numeric solutions.

**AI use declaration:** Codex was used to assist with mpmath style conversion. All lines of code were manually read by a human (me) and iteratively adjusted. Flake ran. 

Co-authored-by: Stéphane Laurent <stla@users.noreply.github.com>

### DIFF
--- a/mpmath/__init__.py
+++ b/mpmath/__init__.py
@@ -48,6 +48,15 @@ jtheta = mp.jtheta
 kleinj = mp.kleinj
 eta = mp.eta
 
+# Weierstrass elliptic functions
+w_invariants = mp.w_invariants
+w_half_periods = mp.w_half_periods
+wp = mp.wp
+wpprime = mp.wpprime
+wsigma = mp.wsigma
+wzeta = mp.wzeta
+invwp = mp.invwp
+
 qp = mp.qp
 qhyper = mp.qhyper
 qgamma = mp.qgamma
--- a/mpmath/__init__.py
+++ b/mpmath/__init__.py
@@ -49,13 +49,13 @@ kleinj = mp.kleinj
 eta = mp.eta
 
 # Weierstrass elliptic functions
-w_invariants = mp.w_invariants
-w_half_periods = mp.w_half_periods
-wp = mp.wp
-wpprime = mp.wpprime
-wsigma = mp.wsigma
-wzeta = mp.wzeta
-invwp = mp.invwp
+weierinvariants = mp.weierinvariants
+weierhalfperiods = mp.weierhalfperiods
+weierp = mp.weierp
+weierpprime = mp.weierpprime
+weiersigma = mp.weiersigma
+weierzeta = mp.weierzeta
+weierpinv = mp.weierpinv
 
 qp = mp.qp
 qhyper = mp.qhyper
--- a/mpmath/functions/elliptic.py
+++ b/mpmath/functions/elliptic.py
@@ -1767,7 +1767,7 @@ def _wzeta_from_tau_omega(ctx, z, omega1, tau_or_omega2):
 # ============================================================================
 
 @defun
-def weierinvariants(ctx, omega1, omega2=None):
+def weierinvariants(ctx, omega):
     r"""
     Returns the Weierstrass invariants `(g_2, g_3)` corresponding to
     the half-periods `(\omega_1, \omega_2)`::
@@ -1775,24 +1775,17 @@ def weierinvariants(ctx, omega1, omega2=None):
         >>> from mpmath import mp, chop, weierinvariants
         >>> mp.dps = 15
         >>> mp.pretty = True
-        >>> g2, g3 = weierinvariants(1, 0.5j)
+        >>> g2, g3 = weierinvariants((1, 0.5j))
         >>> chop(g2)
         129.987495088848
         >>> chop(g3)
         -284.355330876541
 
-    The half-periods may also be passed as a single pair::
-
-        >>> weierinvariants((1, 0.5j))[1]
-        (-284.355330876541 + 0.0j)
-
     """
-    if omega2 is None:
-        try:
-            omega1, omega2 = omega1
-        except (TypeError, ValueError):
-            raise ValueError("weierinvariants: expected "
-                             "omega1, omega2 or a pair of half-periods")
+    try:
+        omega1, omega2 = omega
+    except (TypeError, ValueError):
+        raise ValueError("weierinvariants: expected a pair of half-periods")
     prec = ctx.prec
     try:
         ctx.prec += 10
@@ -1807,7 +1800,7 @@ def weierinvariants(ctx, omega1, omega2=None):
     return +g2, +g3
 
 @defun
-def weierhalfperiods(ctx, g2, g3=None):
+def weierhalfperiods(ctx, g):
     r"""
     Returns a pair of fundamental half-periods `(\omega_1, \omega_2)`
     corresponding to the Weierstrass invariants `(g_2, g_3)`::
@@ -1816,24 +1809,18 @@ def weierhalfperiods(ctx, g2, g3=None):
         >>> from mpmath import weierhalfperiods, weierinvariants
         >>> mp.dps = 15
         >>> mp.pretty = True
-        >>> omega1, omega2 = weierhalfperiods(60, 140)
-        >>> g2, g3 = weierinvariants(omega1, omega2)
+        >>> omega1, omega2 = weierhalfperiods((60, 140))
+        >>> g2, g3 = weierinvariants((omega1, omega2))
         >>> chop(g2), chop(g3)
         (60.0, 140.0)
-
-    The invariants may also be passed as a single pair::
-
-        >>> omega1, omega2 = weierhalfperiods((60, 140))
         >>> chop(omega2/omega1)
         (0.5 + 0.209032224450873j)
 
     """
-    if g3 is None:
-        try:
-            g2, g3 = g2
-        except (TypeError, ValueError):
-            raise ValueError("weierhalfperiods: expected "
-                             "g2, g3 or a pair of invariants")
+    try:
+        g2, g3 = g
+    except (TypeError, ValueError):
+        raise ValueError("weierhalfperiods: expected a pair of invariants")
     prec = ctx.prec
     try:
         ctx.prec += 10
--- a/mpmath/functions/elliptic.py
+++ b/mpmath/functions/elliptic.py
@@ -1476,3 +1476,548 @@ def ellippi(ctx, *args):
         RJ = ctx.elliprj(x, y, 1, 1-n*s**2)
         return s*RF, n*s**3*RJ/3
     return ctx.sum_accurately(terms) + P
+
+
+# Weierstrass Elliptic Functions
+# ============================================================================
+
+def _complex_sort(ctx, list_of_mpcs, ascending=True):
+    """
+    Sorts a list of complex numbers first by real part,
+    then by imaginary part where real parts are equal.
+    """
+    sorted_list = sorted([(c.real, c.imag) for c in list_of_mpcs],
+                         reverse=not ascending)
+    return [ctx.mpc(real=t[0], imag=t[1]) for t in sorted_list]
+
+def _roots_from_omega(ctx, omega1, omega2):
+    """
+    Compute roots e1, e2, e3 of 4*z^3 - g2*z - g3 = 0 using theta functions.
+    This is ~10x faster than solving the cubic directly.
+    """
+    tau = omega2 / omega1
+    q = ctx.qfrom(tau=tau)
+    j24 = ctx.jtheta(2, 0, q)**4
+    j44 = ctx.jtheta(4, 0, q)**4
+    c = ctx.pi**2 / omega1**2 / 12
+    e1 = c * (j24 + 2*j44)
+    e2 = c * (j24 - j44)
+    e3 = -c * (2*j24 + j44)
+    return _complex_sort(ctx, [e1, e2, e3], ascending=False)
+
+def _eisenstein_E4_E6(ctx, tau):
+    """
+    Eisenstein E-series of weight 4 and 6.
+    Uses theta function formula to avoid numerical errors.
+    """
+    q = ctx.qfrom(tau=tau)
+    j2 = ctx.jtheta(2, 0, q)
+    j3 = ctx.jtheta(3, 0, q)
+    j4 = ctx.jtheta(4, 0, q)
+    E4 = (j2**8 + j3**8 + j4**8) / 2
+    E6 = (-3*j2**8 * (j3**4 + j4**4) + (j3**12 + j4**12)) / 2
+    return E4, E6
+
+def _eisenstein_G4_G6(ctx, tau):
+    """
+    Eisenstein G-series of weight 4 and 6.
+    """
+    E4, E6 = _eisenstein_E4_E6(ctx, tau)
+    G4 = 2 * ctx.zeta(4) * E4
+    G6 = 2 * ctx.zeta(6) * E6
+    return G4, G6
+
+def _inverse_kleinj(ctx, _j):
+    """
+    Compute tau from j-invariant using the inverse j-function.
+    See: https://en.wikipedia.org/wiki/J-invariant
+    """
+    _j = ctx.convert(_j)
+    sqrt_arg = 3*(1728*_j**2 - _j**3)
+    exponent = ctx.mpf(1) / ctx.mpf(3)
+    t = (-_j**3 + 2304*_j**2 - 884736*_j +
+         12288*ctx.sqrt(sqrt_arg))**exponent
+    x = ctx.mpf(1)/768*t + (1 - _j/768) - (1536*_j - _j**2) / (768*t)
+
+    lbd = (1 + ctx.sqrt(1 - 4*x)) / 2
+    tau = ctx.j * ctx.agm(1, ctx.sqrt(1-lbd)) / ctx.agm(1, ctx.sqrt(lbd))
+    return tau
+
+def _kleinj_from_g2g3(ctx, g2, g3):
+    """
+    Klein-j invariant from g2, g3 (the 1728-normalized version).
+    """
+    g2 = ctx.convert(g2)
+    g3 = ctx.convert(g3)
+    return 1728 / (1 - 27*g3**2/g2**3)
+
+def _tau_from_g(ctx, g2, g3):
+    """
+    Compute tau (half-period ratio) from g2, g3.
+    """
+    g2 = ctx.convert(g2)
+    g3 = ctx.convert(g3)
+    j = _kleinj_from_g2g3(ctx, g2, g3)
+    tau = _inverse_kleinj(ctx, j)
+    return tau
+
+def _omega_from_g(ctx, g2, g3, tolerance=None):
+    """
+    Compute half-periods (omega1, omega2) from invariants g2, g3.
+    """
+    if tolerance is None:
+        # Start with relaxed tolerance; can be tightened with higher precision
+        tolerance = 0.2
+
+    g2 = ctx.convert(g2)
+    g3 = ctx.convert(g3)
+
+    # Special cases
+    if g2 == 0:
+        omegaA = (g3 ** (ctx.mpf(-1)/ctx.mpf(6)) *
+                  ctx.gamma(ctx.mpf(1)/ctx.mpf(3))**3 / (4*ctx.pi))
+        tau = ctx.mpc(ctx.mpf(1)/ctx.mpf(2), ctx.sqrt(3)/2)
+    elif g3 == 0:
+        tau = _tau_from_g(ctx, g2, g3)
+        G4, G6 = _eisenstein_G4_G6(ctx, tau)
+        omegaA = ctx.j * (ctx.mpf(15)/(4*g2) * G4) ** (ctx.mpf(1)/ctx.mpf(4))
+    else:
+        tau = _tau_from_g(ctx, g2, g3)
+        G4, G6 = _eisenstein_G4_G6(ctx, tau)
+        _sqrt_arg = g2/g3 * G6/G4 * ctx.mpf(7)/ctx.mpf(12)
+        omegaA = ctx.sqrt(_sqrt_arg)
+
+    omegaB = tau * omegaA
+    omegaC = omegaA + omegaB
+
+    # Match omegas to roots using Weierstrass P function
+    omegas = [omegaA, omegaB, omegaC]
+    index_combos = [(0,1,2), (0,2,1), (1,0,2), (1,2,0), (2,0,1), (2,1,0)]
+
+    e1, e2, e3 = _roots_from_omega(ctx, omegaA, omegaB)
+
+    # Compute wp at each omega to match with roots
+    wps = []
+    for omegaN in omegas:
+        wp_val = _wp_from_tau_omega(ctx, omegaN, omegaA, tau)
+        wps.append(wp_val)
+
+    maes = []
+    for ic in index_combos:
+        mae = (ctx.fabs(e1 - wps[ic[0]]) + ctx.fabs(e2 - wps[ic[1]]) +
+               ctx.fabs(e3 - wps[ic[2]])) / 3
+        maes.append(mae)
+
+    mae = min(maes)
+    min_index = maes.index(mae)
+
+    if mae > tolerance:
+        raise ValueError("omega_from_g: convergence tolerance not met")
+
+    omega1, omega2, omega3 = [omegas[k] for k in index_combos[min_index]]
+
+    # Ensure Im(tau) > 0
+    if ctx.im(omega2/omega1) <= 0:
+        omega2 = -omega2
+
+    return omega1, omega2, omega3
+
+def _g_from_omega(ctx, omega1, omega2):
+    """
+    Compute g2, g3 from half-periods omega1, omega2.
+    """
+    omega1 = ctx.convert(omega1)
+    omega2 = ctx.convert(omega2)
+    tau = omega2 / omega1
+    q = ctx.qfrom(tau=tau)
+    j2 = ctx.jtheta(2, 0, q)
+    j3 = ctx.jtheta(3, 0, q)
+    g2 = (ctx.mpf(4)/3) * (ctx.pi/(2*omega1))**4 * (j2**8 - (j2*j3)**4 + j3**8)
+    g3 = ((ctx.mpf(8)/27) * (ctx.pi/(2*omega1))**6 *
+          (j2**12 - (ctx.mpf(3)/2*j2**8*j3**4 +
+                     ctx.mpf(3)/2*j2**4*j3**8) +
+           j3**12))
+    return g2, g3
+
+def _weierstrass_omega_tau(ctx, funcname, g2=None, g3=None,
+                           tau=None, omega=None):
+    """
+    Resolve one Weierstrass parameterization to (omega1, tau).
+    """
+    if (g2 is None) != (g3 is None):
+        raise ValueError("%s: must provide both g2 and g3" % funcname)
+    parameter_count = (int(g2 is not None) + int(tau is not None) +
+                       int(omega is not None))
+    if parameter_count != 1:
+        raise ValueError("%s: must provide exactly one of g2, g3; "
+                         "omega; or tau" % funcname)
+    if omega is not None:
+        omega1, omega2 = omega
+        omega1 = ctx.convert(omega1)
+        omega2 = ctx.convert(omega2)
+        tau = omega2 / omega1
+        if ctx.im(tau) <= 0:
+            raise ValueError("%s: omega ratio must be in upper half-plane" %
+                             funcname)
+        return omega1, tau
+    if tau is not None:
+        tau = ctx.convert(tau)
+        if ctx.im(tau) <= 0:
+            raise ValueError("%s: tau must be in upper half-plane" % funcname)
+        return ctx.one, tau
+    omega1, omega2 = _omega_from_g(ctx, g2, g3)[:2]
+    return omega1, omega2 / omega1
+
+def _wp_from_tau_omega(ctx, z, omega1, tau_or_omega2):
+    """
+    Internal: Weierstrass P function using tau or omega.
+    Computes wp(z; omega1, tau or omega2)
+
+    Following the convention:
+    wp(z, omega1, omega2) = wp_theta(z/(2*omega1), tau) / (omega1^2/4)
+    """
+    z = ctx.convert(z)
+    omega1 = ctx.convert(omega1)
+    tau = ctx.convert(tau_or_omega2)
+
+    # Normalize z by the period: z_norm = z / (2 * omega1)
+    z_norm = z / (2 * omega1)
+
+    q = ctx.qfrom(tau=tau)
+    j1z = ctx.jtheta(1, ctx.pi*z_norm, q)
+    j2 = ctx.jtheta(2, 0, q)
+    j3 = ctx.jtheta(3, 0, q)
+    j4z = ctx.jtheta(4, ctx.pi*z_norm, q)
+
+    # Theta-based Weierstrass P (no omega normalization)
+    wp_theta = ((ctx.pi*j2*j3*j4z/j1z)**2 -
+                ctx.pi**2 * (j2**4 + j3**4) / 3)
+
+    # Apply omega normalization
+    return wp_theta / omega1**2 / 4
+
+def _wpprime_from_tau_omega(ctx, z, omega1, tau_or_omega2):
+    """
+    Internal: Weierstrass P' function using tau or omega.
+    Computes wp'(z; omega1, tau or omega2)
+    """
+    z = ctx.convert(z)
+    omega1 = ctx.convert(omega1)
+    tau = ctx.convert(tau_or_omega2)
+
+    # Normalize z by the period
+    z_norm = z / (2 * omega1)
+
+    q = ctx.qfrom(tau=tau)
+    z1 = ctx.pi * z_norm
+    j10p = ctx.jtheta(1, 0, q, 1)
+    j20 = ctx.jtheta(2, 0, q)
+    j30 = ctx.jtheta(3, 0, q)
+    j40 = ctx.jtheta(4, 0, q)
+    k0 = j10p**3 / (j20 * j30 * j40)
+
+    j1z1 = ctx.jtheta(1, z1, q)
+    j2z1 = ctx.jtheta(2, z1, q)
+    j3z1 = ctx.jtheta(3, z1, q)
+    j4z1 = ctx.jtheta(4, z1, q)
+    kz = j2z1 * j3z1 * j4z1 / j1z1**3
+
+    wpprime_val = -ctx.pi**3 / (4 * omega1**3) * k0 * kz
+    return wpprime_val
+
+def _wsigma_from_tau_omega(ctx, z, omega1, tau_or_omega2):
+    """
+    Internal: Weierstrass sigma function.
+    """
+    z = ctx.convert(z)
+    omega1 = ctx.convert(omega1)
+    tau = ctx.convert(tau_or_omega2)
+
+    z1 = ctx.pi * z / (2 * omega1)
+    q = ctx.qfrom(tau=tau)
+    j10p = ctx.jtheta(1, 0, q, 1)
+    j10ppp = ctx.jtheta(1, 0, q, 3)
+    j1z1 = ctx.jtheta(1, z1, q)
+
+    w_sigma = (2 * omega1 / (ctx.pi * j10p) *
+               ctx.exp(-z1**2 * j10ppp / (6 * j10p)) * j1z1)
+    return w_sigma
+
+def _wzeta_from_tau_omega(ctx, z, omega1, tau_or_omega2):
+    """
+    Internal: Weierstrass zeta function.
+    """
+    z = ctx.convert(z)
+    omega1 = ctx.convert(omega1)
+    tau = ctx.convert(tau_or_omega2)
+
+    w1 = -omega1 / ctx.pi
+    q = ctx.qfrom(tau=tau)
+    p = 1 / 2 / w1
+    eta1 = p / 6 / w1 * ctx.jtheta(1, 0, q, 3) / ctx.jtheta(1, 0, q, 1)
+
+    j1pz = ctx.jtheta(1, p*z, q, 1)
+    j1z = ctx.jtheta(1, p*z, q)
+
+    return -eta1 * z + p * j1pz / j1z
+
+
+# ============================================================================
+# Weierstrass parameter conversion functions
+# ============================================================================
+
+@defun
+def w_invariants(ctx, omega1, omega2=None):
+    r"""
+    Returns the Weierstrass invariants `(g_2, g_3)` corresponding to
+    the half-periods `(\omega_1, \omega_2)`::
+
+        >>> from mpmath import mp, chop, w_invariants
+        >>> mp.dps = 15
+        >>> mp.pretty = True
+        >>> g2, g3 = w_invariants(1, 0.5j)
+        >>> chop(g2)
+        129.987495088848
+        >>> chop(g3)
+        -284.355330876541
+
+    The half-periods may also be passed as a single pair::
+
+        >>> w_invariants((1, 0.5j))[1]
+        (-284.355330876541 + 0.0j)
+
+    """
+    if omega2 is None:
+        try:
+            omega1, omega2 = omega1
+        except (TypeError, ValueError):
+            raise ValueError("w_invariants: expected "
+                             "omega1, omega2 or a pair of half-periods")
+    prec = ctx.prec
+    try:
+        ctx.prec += 10
+        omega1 = ctx.convert(omega1)
+        omega2 = ctx.convert(omega2)
+        if ctx.im(omega2/omega1) <= 0:
+            raise ValueError("w_invariants: omega ratio must be "
+                             "in upper half-plane")
+        g2, g3 = _g_from_omega(ctx, omega1, omega2)
+    finally:
+        ctx.prec = prec
+    return +g2, +g3
+
+@defun
+def w_half_periods(ctx, g2, g3=None):
+    r"""
+    Returns a pair of fundamental half-periods `(\omega_1, \omega_2)`
+    corresponding to the Weierstrass invariants `(g_2, g_3)`::
+
+        >>> from mpmath import mp, chop
+        >>> from mpmath import w_half_periods, w_invariants
+        >>> mp.dps = 15
+        >>> mp.pretty = True
+        >>> omega1, omega2 = w_half_periods(60, 140)
+        >>> g2, g3 = w_invariants(omega1, omega2)
+        >>> chop(g2), chop(g3)
+        (60.0, 140.0)
+
+    The invariants may also be passed as a single pair::
+
+        >>> omega1, omega2 = w_half_periods((60, 140))
+        >>> chop(omega2/omega1)
+        (0.5 + 0.209032224450873j)
+
+    """
+    if g3 is None:
+        try:
+            g2, g3 = g2
+        except (TypeError, ValueError):
+            raise ValueError("w_half_periods: expected "
+                             "g2, g3 or a pair of invariants")
+    prec = ctx.prec
+    try:
+        ctx.prec += 10
+        omega1, omega2 = _omega_from_g(ctx, g2, g3)[:2]
+    finally:
+        ctx.prec = prec
+    return +omega1, +omega2
+
+
+# ============================================================================
+# Main Weierstrass Elliptic Functions
+# ============================================================================
+
+@defun_wrapped
+def wp(ctx, z, g2=None, g3=None, tau=None, omega=None):
+    r"""
+    Weierstrass elliptic function `\wp(z; g_2, g_3)`.
+
+    Computes the Weierstrass P-function, a doubly-periodic elliptic function
+    satisfying the differential equation:
+
+    .. math::
+
+        (\wp')^2 = 4\wp^3 - g_2 \wp - g_3
+
+    The function can be parameterized in multiple ways:
+    - via `g_2, g_3` (elliptic invariants) - recommended
+    - via `\omega_1, \omega_2` (half-periods)
+    - via `\tau` (half-period ratio)
+
+    **Examples**
+
+    Direct computation with invariants::
+
+        >>> from mpmath import mp, wp, chop
+        >>> mp.dps = 15
+        >>> mp.pretty = True
+        >>> chop(wp(0.5, g2=60, g3=140))
+        5.12943876105856
+
+    Using tau parameterization::
+
+        >>> chop(wp(0.5, tau=0.5j))
+        5.15638936351528
+
+    **References**
+
+    - DLMF Section 23.2: https://dlmf.nist.gov/23.2
+    - Weierstrass, K. (1895)
+
+    """
+    z = ctx.convert(z)
+    omega1, tau = _weierstrass_omega_tau(ctx, "wp", g2, g3, tau, omega)
+    return _wp_from_tau_omega(ctx, z, omega1, tau)
+
+@defun_wrapped
+def wpprime(ctx, z, g2=None, g3=None, tau=None, omega=None):
+    r"""
+    Derivative of Weierstrass elliptic function `\wp'(z; g_2, g_3)`.
+
+    Computes the derivative of the Weierstrass P-function.
+
+    **Examples**
+
+    Compute derivative::
+
+        >>> from mpmath import mp, wpprime, chop
+        >>> mp.dps = 15
+        >>> mp.pretty = True
+        >>> chop(wpprime(0.5, g2=60, g3=140))
+        -9.5957928748663
+
+    Verify differential equation::
+
+        >>> from mpmath import mp, wp, wpprime
+        >>> mp.dps = 15
+        >>> z = 0.5
+        >>> g2, g3 = 60, 140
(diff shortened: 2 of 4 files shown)


HeskethGD:
Just saw the workflow said a few lines were not covered by tests and I have updated now so they are.

fredrik-johansson:
At a first glance, this looks reasonable.

mpmath/__init__.py
```diff
@@ -48,6 +48,15 @@
...
 eta = mp.eta
 
+# Weierstrass elliptic functions
+w_invariants = mp.w_invariants
+w_half_periods = mp.w_half_periods
+wp = mp.wp
+wpprime = mp.wpprime
+wsigma = mp.wsigma
+wzeta = mp.wzeta
+invwp = mp.invwp
```
skirpichev: What about naming?  How it was chosen?

wp looks especially odd, a common variable for working precision.  Something like `weierstrass_p` or `weierstrassp`, too verbose?  On another hand, we have shortcuts like `whitw` (https://functions.wolfram.com/HypergeometricFunctions/WhittakerW/).

For inverse functions, we have `erfinv`.

mpmath/tests/test_elliptic.py
```diff
@@ -681,3 +683,192 @@ def test_issue_1104():
...
+# Weierstrass Elliptic Functions
+# ============================================================================
+
+def test_weierstrass_tau_values():
+    mp.dps = 30
+
+    z = mpf('0.3')
+    tau = j/2
+    assert mpc_ae(wp(z, tau=tau),
+                  mpf('11.6229960825738115058320196726'),
```
skirpichev: How reference values were computed?

mpmath/functions/elliptic.py
```diff
@@ -1476,3 +1476,552 @@ def terms():
...
+
+    return -eta1 * z + p * j1pz / j1z
+
+
+# ============================================================================
+# Weierstrass parameter conversion functions
+# ============================================================================
+
+@defun
+def w_invariants(ctx, omega1, omega2=None):
```
skirpichev: I think this should be either two-arguments method of the ctx, or one-argument, accepting a 2-tuple.  Same for w_half_periods().

mpmath/functions/elliptic.py
```diff
@@ -1476,3 +1476,552 @@ def terms():
...
+        tau = ctx.convert(tau)
+        if ctx.im(tau) <= 0:
+            raise ValueError("%s: tau must be in upper half-plane" % funcname)
+        return ctx.one, tau
+    omega1, omega2 = _omega_from_g(ctx, g2, g3)[:2]
+    return omega1, omega2 / omega1
+
+def _wp_from_tau_omega(ctx, z, omega1, tau_or_omega2):
+    """
+    Internal: Weierstrass P function using tau or omega.
```
skirpichev:
```suggestion
    Weierstrass P function using tau or omega.
```
Underscore-prefixed functions are private, per default.

mpmath/functions/elliptic.py
```diff
@@ -1476,3 +1476,552 @@ def terms():
...
+        ctx.prec = prec
+    return +omega1, +omega2
+
+
+# ============================================================================
+# Main Weierstrass Elliptic Functions
+# ============================================================================
+
+@defun_wrapped
+def wp(ctx, z, g2=None, g3=None, tau=None, omega=None):
```
skirpichev: Maybe other functions, like this, also should have simpler signatures?

Like e.g. https://reference.wolfram.com/language/ref/WeierstrassP.html.

mpmath/__init__.py
fredrik-johansson: It could be `weierp`.

mpmath/tests/test_elliptic.py
fredrik-johansson: Reference values don't match flint:
```
>>> CC.weierstrass_p("0.3", tau="I/2")
[16.9966858369455667449328609546283921 +/- 2.10e-35]
```

mpmath/__init__.py
HeskethGD: it was following some of the existing elliptic function style in the code base. eg jtheta for jacobi theta. happy to make it more verbose.

mpmath/functions/elliptic.py
HeskethGD: It was trying to align with existing signatures of elliptic functions elsewhere in the file:

image

mpmath/tests/test_elliptic.py
HeskethGD: The long decimal reference values were calculated with pyweierstrass so they were checking the conversions was consistent but I will check and improve where appropriate. Also included known closed form cases for equianharmonic and lemniscatic cases for half-periods.

HeskethGD: The difference with FLINT is a normalisation choice. This current code the `tau` path treats `tau` as giving half-periods `(1, tau)`, hence periods `(2, 2*tau)`, while FLINT appears to use the normalised period lattice `(1, tau)`, i.e. half-periods `(1/2, tau/2)`. The latter FLINT normalisation is likely the better convention for `tau`, consistent with `q = exp(pi*i*tau)`, so I’ll change `tau=` to use half-periods `(1/2, tau/2)`. 

With the current code the FLINT value is obtained by calling `wp(0.3, omega=(1/2, I/4))` or equivalently `4*wp(0.6, tau=I/2)`.

HeskethGD: Updated the tau convention to match FLINT so now we get:
```
mp.weierp(0.3, tau=1j/2)
(16.9966858369455672988823111976 + 0.0j)
```

HeskethGD: Moved away from the pre-computed reference values in tests. Added more identity test including legendre relations for Weierstrass zeta and the addition theorem for Weierstrass P. Also added comparison with mpmath jacobi sn is consistent with known relation to weierstrass p.

mpmath/functions/elliptic.py
HeskethGD: Removed

mpmath/__init__.py
HeskethGD: Adopted the weierp suggestion and used weier in all others too. Looked at other names in mpmath and coudl see some variation in style but hopefully this is not too offensive or ambiguous.

mpmath/tests/test_elliptic.py
skirpichev: > Moved away from the pre-computed reference values in tests.

Why?  Nothing is wrong with testing for reference values, assuming they are correct.

HeskethGD: The trouble here is they were calculated with the pyweierstrass package which is basically what is being implemented here so it was a bit circular. It served the purpose of checking the port was successful. I don't have access to another tool to calculate other reference values. I added a check that compares against jacobi sn values from mpmath instead.

skirpichev: > it was a bit circular.

I think you can use e.g. FLINT or Wolfram Engine.

mpmath/functions/elliptic.py
HeskethGD: Done.

Made weierinvariants tuple-only: weierinvariants((omega1, omega2))
Made weierhalfperiods tuple-only: weierhalfperiods((g2, g3))

mpmath/tests/test_elliptic.py
HeskethGD: Added tests against Wolfram Engine for P, P prime, zeta and sigma. Two value sets for each. All pass.
Also added extra degenerate case that collapses to sinh.

mpmath/functions/elliptic.py
HeskethGD: I've not changed this as to me it is consistent with the style of other elliptic functions in this file, although not with wolfram, but if you have a specific change request happy to look at it.

HeskethGD:
Think all comments have been addressed/responded to. Let me know if anything outstanding.

mpmath/functions/elliptic.py
skirpichev: I think that direct arguments are better for API.  You can't say, using Python syntax (well, maybe with annotations), that it should be 2-tuple.

mpmath/functions/elliptic.py
```diff
@@ -1476,3 +1476,544 @@ def terms():
...
+
+# Weierstrass Elliptic Functions
+# ============================================================================
+
+def _complex_sort(ctx, list_of_mpcs, ascending=True):
+    """
+    Sorts a list of complex numbers first by real part,
+    then by imaginary part where real parts are equal.
+    """
+    sorted_list = sorted([(c.real, c.imag) for c in list_of_mpcs],
```
skirpichev:
```suggestion
    sorted_list = sorted(((c.real, c.imag) for c in list_of_mpcs),
```

mpmath/functions/elliptic.py
```diff
@@ -1476,3 +1476,544 @@ def terms():
         RJ = ctx.elliprj(x, y, 1, 1-n*s**2)
         return s*RF, n*s**3*RJ/3
     return ctx.sum_accurately(terms) + P
+
+
+# Weierstrass Elliptic Functions
+# ============================================================================
+
+def _complex_sort(ctx, list_of_mpcs, ascending=True):
```
skirpichev: It seems the only use of this helper is in _roots_from_omega(), and with ascending=False.  I suggest inline this helper or drop the option.

mpmath/functions/elliptic.py
```diff
@@ -1476,3 +1476,544 @@ def terms():
...
+         12288*ctx.sqrt(sqrt_arg))**exponent
+    x = ctx.mpf(1)/768*t + (1 - _j/768) - (1536*_j - _j**2) / (768*t)
+
+    lbd = (1 + ctx.sqrt(1 - 4*x)) / 2
+    tau = ctx.j * ctx.agm(1, ctx.sqrt(1-lbd)) / ctx.agm(1, ctx.sqrt(lbd))
+    return tau
+
+def _kleinj_from_g2g3(ctx, g2, g3):
+    """
+    Klein-j invariant from g2, g3 (the 1728-normalized version).
```
skirpichev:
```suggestion
    Klein-j invariant from g2, g3 (the 1728-normalized version).
    See: https://en.wikipedia.org/wiki/J-invariant
```

mpmath/functions/elliptic.py
```diff
@@ -1476,3 +1476,544 @@ def terms():
...
+    """
+    Compute tau (half-period ratio) from g2, g3.
+    """
+    g2 = ctx.convert(g2)
+    g3 = ctx.convert(g3)
+    j = _kleinj_from_g2g3(ctx, g2, g3)
+    tau = _inverse_kleinj(ctx, j)
+    return tau
+
+def _omega_from_g(ctx, g2, g3, tolerance=None):
```
skirpichev: tolerance - always None, isn't?  //  Coverage tests not check branches yet.

Then, probably this could be a local variable.  BTW, I'm surprised that it's too big.  Will this work for high precision?

mpmath/functions/elliptic.py
```diff
@@ -1476,3 +1476,544 @@ def terms():
...
+    # Compute wp at each omega to match with roots
+    wps = []
+    for omegaN in omegas:
+        wp_val = _wp_from_tau_omega(ctx, omegaN, omegaA, tau)
+        wps.append(wp_val)
+
+    maes = []
+    for ic in index_combos:
+        mae = (ctx.fabs(e1 - wps[ic[0]]) + ctx.fabs(e2 - wps[ic[1]]) +
+               ctx.fabs(e3 - wps[ic[2]])) / 3
```
skirpichev:
```suggestion
        mae = (abs(e1 - wps[ic[0]]) + abs(e2 - wps[ic[1]]) +
               abs(e3 - wps[ic[2]])) / 3
```

mpmath/functions/elliptic.py
```diff
@@ -1476,3 +1476,544 @@ def terms():
...
+    Derivative of Weierstrass elliptic function `\wp'(z; g_2, g_3)`.
+
+    Computes the derivative of the Weierstrass P-function.
+
+    **Examples**
+
+    Compute derivative::
+
+        >>> from mpmath import mp, weierpprime, chop
+        >>> mp.dps = 15
```
skirpichev:
```suggestion
```

mpmath/functions/elliptic.py
```diff
@@ -1476,3 +1476,544 @@ def terms():
...
+        >>> mp.pretty = True
+        >>> chop(weierpprime(0.5, g2=60, g3=140))
+        -9.5957928748663
+
+    Verify differential equation::
+
+        >>> from mpmath import mp, weierp, weierpprime
+        >>> mp.dps = 15
```
skirpichev:
```suggestion
```

mpmath/functions/elliptic.py
```diff
@@ -1476,3 +1476,544 @@ def terms():
...
+    .. math::
+
+        \wp(z) = -\frac{d^2}{dz^2} \ln \sigma(z)
+
+    **Examples**
+
+    Compute sigma function::
+
+        >>> from mpmath import mp, weiersigma, chop
+        >>> mp.dps = 15
```
skirpichev:
```suggestion
```

mpmath/functions/elliptic.py
```diff
@@ -1476,3 +1476,544 @@ def terms():
...
+
+    The Weierstrass zeta function is related to the P and sigma functions.
+    It is quasi-periodic (not doubly periodic).
+
+    **Examples**
+
+    Compute zeta function::
+
+        >>> from mpmath import mp, weierzeta, chop
+        >>> mp.dps = 15
```
skirpichev:
```suggestion
```

mpmath/functions/elliptic.py
```diff
@@ -1476,3 +1476,544 @@ def terms():
...
+# ============================================================================
+
+@defun
+def weierinvariants(ctx, omega):
+    r"""
+    Returns the Weierstrass invariants `(g_2, g_3)` corresponding to
+    the half-periods `(\omega_1, \omega_2)`::
+
+        >>> from mpmath import mp, chop, weierinvariants
+        >>> mp.dps = 15
```
skirpichev:
```suggestion
```
I think we shouldn't mention defaults.

mpmath/functions/elliptic.py
```diff
@@ -1476,3 +1476,544 @@ def terms():
...
+
+@defun
+def weierhalfperiods(ctx, g):
+    r"""
+    Returns a pair of fundamental half-periods `(\omega_1, \omega_2)`
+    corresponding to the Weierstrass invariants `(g_2, g_3)`::
+
+        >>> from mpmath import mp, chop
+        >>> from mpmath import weierhalfperiods, weierinvariants
+        >>> mp.dps = 15
```
skirpichev:
```suggestion
```

mpmath/functions/elliptic.py
```diff
@@ -1476,3 +1476,544 @@ def terms():
...
+    For repeated evaluation with the same invariants, it is faster to compute
+    the half-periods once with :func:`~mpmath.weierhalfperiods` and pass them
+    using the `omega` keyword.
+
+    **Examples**
+
+    Direct computation with invariants::
+
+        >>> from mpmath import mp, weierp, chop
+        >>> mp.dps = 15
```
skirpichev:
```suggestion
```

mpmath/tests/test_elliptic.py
```diff
@@ -681,3 +684,342 @@ def test_issue_1104():
...
+
+# Weierstrass Elliptic Functions
+# ============================================================================
+
+def test_weierstrass_tau_uses_normalized_periods():
+    mp.dps = 30
+
+    z = mpf('0.3')
+    tau = j/2
+    omega = (mpf(1)/2, tau/2)
```
skirpichev:
```suggestion
    omega = (0.5, tau/2)
```
Lets use literals, where it's possible.  (1/2, if you prefer.)

mpmath/tests/test_elliptic.py
```diff
@@ -681,3 +684,342 @@ def test_issue_1104():
...
+
+    for f in [weierp, weierpprime, weiersigma, weierzeta]:
+        assert mpc_ae(f(z, tau=tau), f(z, omega=omega), eps=eps*1000)
+
+def test_weierstrass_g2g3_differential_equation():
+    mp.dps = 30
+
+    z = mpf('0.3')
+    for g2, g3 in [(mpf(60), mpf(140)), (mpf(0), mpf(140)),
+                   (mpf(60), mpf(0))]:
```
skirpichev:
```suggestion
    for g2, g3 in [(60, 140), (0, 140), (60, 0)]:
```

mpmath/tests/test_elliptic.py
```diff
@@ -681,3 +684,342 @@ def test_issue_1104():
...
+        p = weierp(z, g2=g2, g3=g3)
+        pp = weierpprime(z, g2=g2, g3=g3)
+        assert mpc_ae(pp**2, 4*p**3 - g2*p - g3, eps=eps*1000)
+
+def test_weierstrass_parameter_conversions():
+    mp.dps = 30
+
+    omega = (mpf(1), j/2)
```
skirpichev:
```suggestion
    omega = (1, j/2)
```

mpmath/tests/test_elliptic.py
```diff
@@ -681,3 +684,342 @@ def test_issue_1104():
...
+        m*omega1 + n*omega2
+        for m in [-1, 0, 1]
+        for n in [-1, 0, 1]
+        if m or n
+    ]
+    assert min(abs(point - lemniscatic) for point in lattice_points) < eps*1000
+    assert min(abs(point - j*lemniscatic) for point in lattice_points) < eps*1000
+
+    equianharmonic = gamma(mpf(1)/3)**3/(4*pi)
+    tau = mpf(1)/2 + sqrt(3)*j/2
```
skirpichev:
```suggestion
    tau = 0.5 + sqrt(3)*j/2
```

mpmath/tests/test_elliptic.py
```diff
@@ -681,3 +684,342 @@ def test_issue_1104():
...
+    omega1, omega2 = weierhalfperiods((mpf(0), mpf(1)))
```
skirpichev:
```suggestion
    omega1, omega2 = weierhalfperiods((0, 1))
```

mpmath/tests/test_elliptic.py
```diff
@@ -681,3 +684,342 @@ def test_issue_1104():
...
+
+def test_weierstrass_half_periods_convergence_failure():
+    old_dps = mp.dps
+    try:
+        mp.dps = 5
+        pytest.raises(ValueError,
+                      lambda: weierhalfperiods((mpf('1e-100'),
+                                                mpf('1e-50'))))
+    finally:
+        mp.dps = old_dps
```
skirpichev:
```suggestion
    mp.dps = 5
    pytest.raises(ValueError,
                  lambda: weierhalfperiods((mpf('1e-100'), mpf('1e-50'))))
```

In tests you might freely change working precision.  It will be restored to defaults *before* each test.

mpmath/tests/test_elliptic.py
```diff
@@ -681,3 +684,342 @@ def test_issue_1104():
...
+
+def test_weierstrass_parameter_conversions_with_kleinj():
+    mp.dps = 30
+
+    tau = mpf('0.625') + mpf('0.75')*j
```
skirpichev:
```suggestion
    tau = 0.625 + 0.75j
```

mpmath/tests/test_elliptic.py
```diff
@@ -681,3 +684,342 @@ def test_issue_1104():
...
+    g2, g3 = weierinvariants((mpf(1)/2, tau/2))
```
skirpichev:
```suggestion
    g2, g3 = weierinvariants((0.5, tau/2))
```

mpmath/tests/test_elliptic.py
```diff
@@ -681,3 +684,342 @@ def test_issue_1104():
...
+    recovered_omega1, recovered_omega2 = weierhalfperiods((g2, g3))
+    recovered_tau = recovered_omega2/recovered_omega1
+    j_from_invariants = g2**3/(g2**3 - mpf(27)*g3**2)
```
skirpichev:
```suggestion
    j_from_invariants = g2**3/(g2**3 - 27*g3**2)
```

mpmath/tests/test_elliptic.py
```diff
@@ -681,3 +684,342 @@ def test_issue_1104():
...
+
+    assert mpc_ae(kleinj(tau), j_from_invariants, eps=eps*1000)
+    assert mpc_ae(kleinj(recovered_tau), kleinj(tau), eps=eps*1000)
+
+def test_weierstrass_half_period_values_are_cubic_roots():
+    mp.dps = 30
+
+    omega1 = mpf(1)
```
skirpichev:
```suggestion
    omega1 = 1
```

mpmath/tests/test_elliptic.py
```diff
@@ -681,3 +684,342 @@ def test_issue_1104():
...
+    omega2 = j/2
+    omega = (omega1, omega2)
+    g2, g3 = weierinvariants(omega)
+
+    roots = polyroots([-g3, -g2, mpf(0), mpf(4)], maxsteps=50)
```
skirpichev:
```suggestion
    roots = polyroots([-g3, -g2, 0, 4], maxsteps=50)
```

mpmath/tests/test_elliptic.py
```diff
@@ -681,3 +684,342 @@ def test_issue_1104():
...
+    half_period_values = [
+        weierp(omega1, omega=omega),
+        weierp(omega2, omega=omega),
+        weierp(omega1 + omega2, omega=omega),
+    ]
+
+    for value in half_period_values:
+        assert mpc_ae(4*value**3 - g2*value - g3, mpf(0),
```
skirpichev:
```suggestion
        assert mpc_ae(4*value**3 - g2*value - g3, 0,
```

mpmath/tests/test_elliptic.py
```diff
@@ -681,3 +684,342 @@ def test_issue_1104():
...
+                      eps=eps*1000)
+        assert min(abs(value - root) for root in roots) < eps*1000
+    for root in roots:
+        assert min(abs(value - root) for value in half_period_values) < eps*1000
+
+def test_weierstrass_conversions_with_weierp():
+    mp.dps = 30
+
+    z = mpf('0.3')
+    g2, g3 = mpf(60), mpf(140)
```
skirpichev:
```suggestion
    g2, g3 = 60, 140
```

mpmath/tests/test_elliptic.py
```diff
@@ -681,3 +684,342 @@ def test_issue_1104():
...
+    omega = weierhalfperiods((g2, g3))
+    assert mpc_ae(weierp(z, g2=g2, g3=g3), weierp(z, omega=omega),
+                  eps=eps*1000)
+
+def test_weierstrass_periodicity():
+    mp.dps = 30
+
+    z = mpf('0.3')
+    omega1 = mpf(1)
```
skirpichev:
```suggestion
    omega1 = 1
```

mpmath/tests/test_elliptic.py
```diff
@@ -681,3 +684,342 @@ def test_issue_1104():
...
+                  eps=eps*1000)
+    assert mpc_ae(weierpprime(z + 2*omega2, omega=omega), pp,
+                  eps=eps*1000)
+
+def test_weierstrass_scaling_laws():
+    mp.dps = 30
+
+    z = mpf('0.3')
+    scale = mpf('1.7')
+    omega = (mpf(1), j/2)
```
skirpichev:
```suggestion
    omega = (1, j/2)
```

mpmath/tests/test_elliptic.py
```diff
@@ -681,3 +684,342 @@ def test_issue_1104():
...
+                  scale*weiersigma(z, omega=omega), eps=eps*1000)
+    assert mpc_ae(weierzeta(scale*z, omega=scaled_omega),
+                  weierzeta(z, omega=omega)/scale, eps=eps*1000)
+
+def test_weierstrass_tau_omega_parameterizations():
+    mp.dps = 30
+
+    z = mpf('0.3')
+    tau = j/2
+    omega = (mpf(1)/2, tau/2)
```
skirpichev:
```suggestion
    omega = (0.5, tau/2)
```

mpmath/tests/test_elliptic.py
```diff
@@ -681,3 +684,342 @@ def test_issue_1104():
...
+    for f in [weierp, weierpprime, weiersigma, weierzeta]:
+        assert mpc_ae(f(z, tau=tau), f(z, omega=omega))
+
+def test_weierstrass_addition_theorem():
+    mp.dps = 30
+
+    z = mpf('0.3')
+    w = mpf('0.4') + j/10
+    omega = (mpf(1), j/2)
```
skirpichev:
```suggestion
    omega = (1, j/2)
```

mpmath/tests/test_elliptic.py
```diff
@@ -681,3 +684,342 @@ def test_issue_1104():
...
+    ppw = weierpprime(w, omega=omega)
+    rhs = ((ppz - ppw)/(pz - pw))**2/4 - pz - pw
+
+    assert mpc_ae(weierp(z + w, omega=omega), rhs, eps=eps*1000)
+
+def test_weierstrass_zeta_legendre_relation():
+    mp.dps = 30
+
+    z = mpf('0.3') + j/10
+    omega1 = mpf(1)
```
skirpichev:
```suggestion
    omega1 = 1
```

mpmath/tests/test_elliptic.py
```diff
@@ -681,3 +684,342 @@ def test_issue_1104():
...
+                  weiersigma(z, tau=tau), weierzeta(z, tau=tau),
+                  eps=eps*1000)
+    assert mpc_ae(diff(lambda t: weierzeta(t, tau=tau), z),
+                  -weierp(z, tau=tau), eps=eps*1000)
+
+def test_weierstrass_weierpinv():
+    mp.dps = 30
+
+    z = mpf('0.3')
+    g2, g3 = mpf(60), mpf(140)
```
skirpichev:
```suggestion
    g2, g3 = 60, 140
```

mpmath/tests/test_elliptic.py
```diff
@@ -681,3 +684,342 @@ def test_issue_1104():
...
+
+def test_weierstrass_values_from_wolfram_engine():
+    """
+    Test values computed with Wolfram Engine at 50 decimal digits.
+    """
+    mp.dps = 30
+
+    z = mpf(1)/5 + j/10
+    g2 = mpf(23)
+    g3 = -mpf(6)
```
skirpichev:
```suggestion
    g2 = 23
    g3 = -6
```

mpmath/tests/test_elliptic.py
```diff
@@ -681,3 +684,342 @@ def test_issue_1104():
...
+    assert mpc_ae(result, res, eps=eps*1000)
+
+    # Wolfram Engine N[WeierstrassSigma[1/5 + I/10, {23, -6}], 50]
+    res = (mpf('0.20003622045198197835660834749697373254229661740043') +
+           mpf('0.09996069154774003218065176170970294010303412640179')*j)
+    result = weiersigma(z, g2=g2, g3=g3)
+    assert mpc_ae(result, res, eps=eps*1000)
+
+    z = mpf(23)/7 + j/19
+    g2 = mpf(4)
```
skirpichev:
```suggestion
    g2 = 4
```

mpmath/tests/test_elliptic.py
```diff
@@ -681,3 +684,342 @@ def test_issue_1104():
...
+    pytest.raises(ValueError, lambda: weierp(z, g2=mpf(1)))
+    pytest.raises(ValueError, lambda: weierp(z, tau=-j))
+    pytest.raises(ValueError, lambda: weierp(z, omega=(mpf(1), -j)))
+    pytest.raises(ValueError,
+                  lambda: weierp(z, g2=mpf(60), g3=mpf(140), tau=j/2))
+    pytest.raises(TypeError, lambda: weierinvariants(mpf(1), -j))
+    pytest.raises(ValueError, lambda: weierinvariants((mpf(1), -j)))
+    pytest.raises(ValueError, lambda: weierinvariants(mpf(1)))
+    pytest.raises(TypeError, lambda: weierhalfperiods(mpf(1), mpf(0)))
+    pytest.raises(ValueError, lambda: weierhalfperiods(mpf(1)))
```
skirpichev:
```suggestion
    pytest.raises(ValueError, lambda: weierp(z))
    pytest.raises(ValueError, lambda: weierp(z, g2=1))
    pytest.raises(ValueError, lambda: weierp(z, tau=-j))
    pytest.raises(ValueError, lambda: weierp(z, omega=(1, -j)))
    pytest.raises(ValueError,
                  lambda: weierp(z, g2=60, g3=140, tau=j/2))
    pytest.raises(TypeError, lambda: weierinvariants(1, -j))
    pytest.raises(ValueError, lambda: weierinvariants((1, -j)))
    pytest.raises(ValueError, lambda: weierinvariants(1))
    pytest.raises(TypeError, lambda: weierhalfperiods(1, 0))
    pytest.raises(ValueError, lambda: weierhalfperiods(1))
```

mpmath/tests/test_elliptic.py
```diff
@@ -681,3 +684,342 @@ def test_issue_1104():
...
+
+    g2_roundtrip, g3_roundtrip = weierinvariants((omega1, omega2))
+    assert mpc_ae(g2, g2_roundtrip, eps=eps*10000)
+    assert mpc_ae(g3, g3_roundtrip, eps=eps*10000)
+    assert (omega2/omega1).imag > 0
+
+def test_weierstrass_special_half_periods():
+    mp.dps = 30
+
+    lemniscatic = gamma(mpf(1)/4)**2/(4*sqrt(pi))
```
skirpichev:
```suggestion
    lemniscatic = gamma(0.25)**2/(4*sqrt(pi))
```

mpmath/tests/test_elliptic.py
```diff
@@ -681,3 +684,342 @@ def test_issue_1104():
...
+    # and 4*(x - e1)*(x - e2)*(x - e3) = 4*x**3 - g2*x - g3.
+    e1 = mpf(2)
+    e2 = -mpf(1)/2
+    e3 = -mpf(3)/2
+
+    g2 = -4*(e1*e2 + e1*e3 + e2*e3)
+    g3 = 4*e1*e2*e3
+
+    scale = sqrt(e1 - e3)
+    m = (e2 - e3)/(e1 - e3)
```
skirpichev:
```suggestion
    e1 = 2
    e2 = -0.5
    e3 = -1.5

    g2 = -4*(e1*e2 + e1*e3 + e2*e3)
    g3 = 4*e1*e2*e3

    scale = sqrt(e1 - e3)
    m = mpf(e2 - e3)/(e1 - e3)
```

mpmath/tests/test_elliptic.py
```diff
@@ -681,3 +684,342 @@ def test_issue_1104():
...
+    mp.dps = 30
+
+    z = mpf('0.3')
+    tau = j/2
+    omega = (mpf(1)/2, tau/2)
+
+    for f in [weierp, weierpprime, weiersigma, weierzeta]:
+        assert mpc_ae(f(z, tau=tau), f(z, omega=omega), eps=eps*1000)
+
+def test_weierstrass_g2g3_differential_equation():
```
skirpichev, line 811: Maybe you could add some references for tested identities, say [https://dlmf.nist.gov/23.3.E10](https://dlmf.nist.gov/23.3.E10).

mpmath/functions/elliptic.py
```diff
@@ -1476,3 +1476,544 @@ def terms():
...
+
+    return -eta1 * z + p * j1pz / j1z
+
+
+# ============================================================================
+# Weierstrass parameter conversion functions
+# ============================================================================
+
+@defun
+def weierinvariants(ctx, omega):
```
skirpichev: I would prefer signature like weierinvariants(ctx, omega1, omega2), same for weierhalfperiods.

OR qfrom-like API.

mpmath/functions/elliptic.py
```diff
@@ -1476,3 +1476,544 @@ def terms():
...
+
+    Using tau parameterization::
+
+        >>> chop(weierp(0.5, tau=0.5j))
+        13.7503716360407
+
+    **References**
+
+    - DLMF Section 23.2: https://dlmf.nist.gov/23.2
+    - Weierstrass, K. (1895)
```
skirpichev:
```suggestion
    - [DLMF]_ Chapter 23: Weierstrass Elliptic and Modular Functions (23.2.4)
```
Probably, similar for other main functions.

mpmath/functions/elliptic.py
```diff
@@ -1476,3 +1476,544 @@ def terms():
...
+        (\wp')^2 = 4\wp^3 - g_2 \wp - g_3
+
+    The function can be parameterized in multiple ways:
+    - via `g_2, g_3` (elliptic invariants)
+    - via `\omega_1, \omega_2` (half-periods)
+    - via `\tau`, using normalized periods `(1, \tau)`
+
+    For repeated evaluation with the same invariants, it is faster to compute
+    the half-periods once with :func:`~mpmath.weierhalfperiods` and pass them
+    using the `omega` keyword.
```
skirpichev: As more efficient seems to be "omega1, tau" parametrization - maybe you can just use one in function signatures and inline helpers like _wp_from_tau_omega()?

Or just add explicit omega1/omega2 parameters.

Also, consider adding qfrom/taufrom-like functions instead of weierinvariants/weierhalfperiods.

skirpichev (commented):
Looks good, here is mostly a bunch of nitpicks.

Major points is functions API.  I think you should either:
1. simplify function signatures to "tau, omega1" parametrization
2. add explicit omega1, omega2 arguments

Probably, you should also use qfrom-like API to construct either g2, g3 or omega1, omega2.  Maybe weiergfrom/weieromegafrom instead of current conversion functions.  Though, current naming seems acceptable too.

Maybe you could also expand a bit sphinx docs for elliptic functions to explain parametrization for new set of functions.

mpmath/tests/test_elliptic.py
HeskethGD: I was finding that if I passed decimals or fractions in places I did not always get the required precision in test comparisons. Hence the use of mpf(1)/2 etc in several places in tests.

For example this passes at high precision:
```
def test_weierstrass_degenerate_sinh_case():

    z = mp.mpf('2.3456')
    g2 = mp.mpf(1)/12
    g3 = -mp.mpf(1)/216

    expected = mp.mpf(1)/12 + mp.mpf(1)/(4*mp.sinh(z/2)**2)
    actual = mp.weierp(z, g2=g2, g3=g3)

    assert mpc_ae(actual, expected)
 ```
 but this does not:
 ```
 def test_weierstrass_degenerate_sinh_case():

    z = mp.mpf('2.3456')
    g2 = 1/12
    g3 = -1/216

    expected = mp.mpf(1)/12 + mp.mpf(1)/(4*mp.sinh(z/2)**2)
    actual = mp.weierp(z, g2=g2, g3=g3)

    assert mpc_ae(actual, expected)
```

skirpichev: There should be no difference for exactly representable numbers like 0.5.

mpmath/functions/elliptic.py
HeskethGD: Reverted back to omega1, omega2

mpmath/functions/elliptic.py
HeskethGD: Moved inline

mpmath/functions/elliptic.py
HeskethGD: Done

mpmath/functions/elliptic.py
HeskethGD: Made it adaptive. Made local variable.

mpmath/functions/elliptic.py
HeskethGD: Done

mpmath/functions/elliptic.py
HeskethGD: Removed all

mpmath/tests/test_elliptic.py
HeskethGD: Done

mpmath/tests/test_elliptic.py
HeskethGD: Done

mpmath/functions/elliptic.py
HeskethGD: Done omega1, omega2

mpmath/functions/elliptic.py
HeskethGD: Done

mpmath/functions/elliptic.py
HeskethGD: Added omega1, omega2. Reluctant to only have omega1, tau or similar because it can be a bit annoying to have to construct a lattice just to check a few wp values by g2, g3. Mathematica only provides g2, g3 I believe.

mpmath/tests/test_elliptic.py
HeskethGD: OK fair enough. I've not changed these. There only in tests and it runs in 4.5s. If it needs changing can take a look on Monday.

mpmath/functions/elliptic.py
HeskethGD: Done

mpmath/functions/elliptic.py
```diff
@@ -1476,3 +1475,593 @@ def terms():
...
+    j1z1 = ctx.jtheta(1, z1, q)
+    j2z1 = ctx.jtheta(2, z1, q)
+    j3z1 = ctx.jtheta(3, z1, q)
+    j4z1 = ctx.jtheta(4, z1, q)
+    kz = j2z1 * j3z1 * j4z1 / j1z1**3
+
+    wpprime_val = -ctx.pi**3 / (4 * omega1**3) * k0 * kz
+    return wpprime_val
+
+def _wsigma_from_tau_omega(ctx, z, omega1, tau_or_omega2):
```
skirpichev: This helper seems redundant, it's used in one place.

mpmath/functions/elliptic.py
```diff
@@ -1476,3 +1475,593 @@ def terms():
...
+    q = ctx.qfrom(tau=tau)
+    j10p = ctx.jtheta(1, 0, q, 1)
+    j10ppp = ctx.jtheta(1, 0, q, 3)
+    j1z1 = ctx.jtheta(1, z1, q)
+
+    w_sigma = (2 * omega1 / (ctx.pi * j10p) *
+               ctx.exp(-z1**2 * j10ppp / (6 * j10p)) * j1z1)
+    return w_sigma
+
+def _wzeta_from_tau_omega(ctx, z, omega1, tau_or_omega2):
```
skirpichev: This too.

mpmath/functions/elliptic.py
```diff
@@ -1476,3 +1475,593 @@ def terms():
...
+        ctx.prec += 10
+        omega1 = ctx.convert(omega1)
+        omega2 = ctx.convert(omega2)
+        if ctx.im(omega2/omega1) <= 0:
+            raise ValueError("weierinvariants: omega ratio must be "
+                             "in upper half-plane")
+        g2, g3 = _g_from_omega(ctx, omega1, omega2)
+    finally:
+        ctx.prec = prec
+    return +g2, +g3
```
skirpichev: This could be:
```suggestion
    with ctx.extraprec(10):
        omega1 = ctx.convert(omega1)
        omega2 = ctx.convert(omega2)
        if ctx.im(omega2/omega1) <= 0:
            raise ValueError("weierinvariants: omega ratio must be "
                             "in upper half-plane")
        g2, g3 = _g_from_omega(ctx, omega1, omega2)
    return +g2, +g3
```

This will not work for fp/iv contexts, but I think it's a separate issue, that must be fixed.

Or even:
```suggestion
    omega1 = ctx.convert(omega1)
    omega2 = ctx.convert(omega2)
    if ctx.im(omega2/omega1) <= 0:
        raise ValueError("weierinvariants: omega ratio must be "
                         "in upper half-plane")
    return ctx.extraprec(10, normalize_output=True)(_g_from_omega)(ctx, omega1, omega2)
```

Same for weierhalfperiods().

skirpichev (commented):
In the end I'll squash commits.  You can provide a draft for the commit message.

mpmath/tests/test_elliptic.py
HeskethGD: Made most of these changes. Also found the mp_ae() funciton did not like integers in this case so kept mpf(60) mpf(140):
```
def test_weierstrass_half_periods_high_precision():
    mp.dps = 80

    g2 = mpf(60)
    g3 = mpf(140)
    omega1, omega2 = weierhalfperiods(g2, g3)
    g2_roundtrip, g3_roundtrip = weierinvariants(omega1, omega2)

    assert mpc_ae(g2, g2_roundtrip, eps=eps*10000)
    assert mpc_ae(g3, g3_roundtrip, eps=eps*10000)
    
E       AttributeError: 'int' object has no attribute 'ae'
```

mpmath/functions/elliptic.py
HeskethGD: Did the first option.

mpmath/functions/elliptic.py
HeskethGD: Done

mpmath/functions/elliptic.py
HeskethGD: Done

HeskethGD:
The approach taken so far was cautious in regards to not modifying existing functions. There is some scope to eliminate some more helpers if we do modify some existing functions in the elliptic.py file. 

For example we could unify the handling of tau and kleinj:

- absorb the _tau_from_g functionality into the existing taufrom function by adding g2, g3 args.
- the _tau_from_g currently depends on _kleinj_from_g2g3 and _inverse_kleinj. We could modify the existing kleinj to take g2, g3. We'd need to adjust how the new functions use the 1728 j vs J scaling in appropriate places but not too tricky, we'd favour J over j like the existing kleinj.
- the _inverse_kleinj function could become a public function kleinjinv under the modular functions and adjusted to take J not j (i.e. change 1728 norm). It would only take g2, g3 and return tau.

Let me know if you think this is a good idea and I will add the change in this PR.

mpmath/tests/test_elliptic.py
skirpichev: Looks as a bug in mpc_ae.  Lets fix this in a separate pr.

skirpichev:
> The approach taken so far was cautious in regards to not modifying existing functions. There is some scope to eliminate some more helpers if we do modify some existing functions in the elliptic.py file.

I think it's fine as a first step: it's more or less a direct integration of some external package.  Until beta, we are free to change newly added API.

But someone (maybe you) can do a better integration in following prs:
1. I'm not very happy with current qfrom-like API for functions.  IMO, it's better to have more simple function signatures, that handle basic case of parametrization + set of qfrom-like functions for parameter conversion.
2. I think we must expand our prose docs (docs/functions/elliptic.rst) to describe new parameters.

> absorb the _tau_from_g functionality into the existing taufrom function by adding g2, g3 args.

That definitely does make sense.  Probably also for qfrom.  Make sense for following prs.

> the _tau_from_g currently depends on _kleinj_from_g2g3 and _inverse_kleinj. We could modify the existing kleinj to take g2, g3.

Wrt API change - seems fine.  I think it's also a good opportunity to address #1056 in this module.  IIUIC, there is no compatibility break.

> We'd need to adjust how the new functions use the 1728 j vs J scaling in appropriate places but not too tricky, we'd favour J over j like the existing kleinj.

Hmm, maybe this is something you can address in the *current* pr.

> the _inverse_kleinj function could become a public function kleinjinv under the modular functions and adjusted to take J not j (i.e. change 1728 norm). It would only take g2, g3 and return tau.

I think that kleinjinv() could be added later.  Though, you might adjust _inverse_kleinj() to use J in the current PR.  BTW, this eliminates name clash with ``j``.

skirpichev (commented):
LGTM

Let me know if you wish to address use the 1728 j vs J scaling for new functions in the current PR.

Please also add a draft for the commit message as a pr comment.  Added code is slightly different from the pyweierstrass package, lets document this (e.g. using normalized period lattice like FLINT).

Lets then also wait few days if Fredrik would like to review this.  Gentle ping: @fredrik-johansson 

Thanks for your work!

skirpichev (approved):

HeskethGD:
Thanks! I will push an update re J vs j on here shortly.


Draft commit message:


**Add Weierstrass elliptic functions**

Adds Weierstrass elliptic functions, namely:

* Weierstrass $\wp$: `weierp`
* inverse Weierstrass $\wp$: `weierpinv`
* derivative of Weierstrass $\wp$: `weierpprime`
* Weierstrass zeta: `weierzeta`
* Weierstrass sigma: `weiersigma`

Also adds parameter conversion convenience functions:

* Weierstrass invariants $g_2, g_3$: `weierinvariants`
* half-periods $\omega_1, \omega_2$: `weierhalfperiods`

Tests and documentation are updated to cover the new functions.

The idea for this PR was inspired by the `pyweierstrass` package by @stla / Stéphane Laurent:

https://pyweierstrass.readthedocs.io/en/latest/
https://github.com/stla/pyweierstrass

The code implementation in this PR is different to that in `pyweierstrass` although both packages derive from similar well known mathematical formulas. Differences between `pyweierstrass` and this `mpmath` implementation include but are not limited to:

* function names adapted to `mpmath` conventions;
* tau normalization: in this implementation, `tau` specifies the normalized period lattice `(1, tau)`, corresponding to half-periods `(1/2, tau/2)`; this differs from `pyweierstrass`, where `tau` denotes half-periods `(1, tau)`;
* support for disambiguating inverse values of Weierstrass $\wp$ by optionally passing the corresponding derivative value;
* internal refactoring of helper functions for integration with `mpmath`.

AI use declaration: Codex GPT-5.5 was used to assist with conversion to `mpmath` style. All lines of code were manually reviewed by the PR author and iteratively adjusted.

Dedicated to Stéphane Laurent who opened the original PR to request this functionality in `mpmath` and whose `pyweierstrass` package provided great utility to the author of this PR in the years since.

Co-authored-by: Stéphane Laurent [stla@users.noreply.github.com](mailto:stla@users.noreply.github.com)

Edit:
closes https://github.com/mpmath/mpmath/issues/612

skirpichev (approved):

skirpichev:
I worry, that pyweierstrass has a very different license.  (Unfortunately, any licenses except maybe Public Domain are beyond of my comprehension.)

@stla, are you ok with the current patch, can it be included in the mpmath on it's license terms?  // https://github.com/mpmath/mpmath/blob/master/LICENSE

Are you ok with how your work will be credited?  See commit message draft above: https://github.com/mpmath/mpmath/pull/1113#issuecomment-4826279880

HeskethGD:
> I worry, that pyweierstrass has a very different license. (Unfortunately, any licenses except maybe Public Domain are beyond of my comprehension.)
> 
> @stla, are you ok with the current patch, can it be included in the mpmath on it's license terms? // https://github.com/mpmath/mpmath/blob/master/LICENSE
> 
> Are you ok with how your work will be credited? See commit message draft above: [#1113 (comment)](https://github.com/mpmath/mpmath/pull/1113#issuecomment-4826279880)

If no reply is received I may close and do a new PR that does not reuse any code from `pyweierstrass` and derives directly from academic sources. In truth, after much refactoring and the already existing academic derivation, that is largely the case already but I will make certain if we cannot make contact.

skirpichev:
> If no reply is received I may close and do a new PR that does not reuse any code from pyweierstrass and derives directly from academic sources.

No, please don't do this.

It would be a good thing from your co-author to accept mpmath license terms.  But if it is impossible, lets merge this as is.  I think it's already just an implementation of well known academic sources.

I will merge this tomorrow.  BTW, now there is a merge conflict.

HeskethGD:
> > If no reply is received I may close and do a new PR that does not reuse any code from pyweierstrass and derives directly from academic sources.
> 
> No, please don't do this.
> 
> It would be a good thing from your co-author to accept mpmath license terms. But if it is impossible, lets merge this as is. I think it's already just an implementation of well known academic sources.
> 
> I will merge this tomorrow. BTW, now there is a merge conflict.

As far as I can tell conflict is caused by this change which adds `inf, nsum` to test elliptic file. I could fix by importing those but as far as i can tell they are not used in the test file:

https://github.com/mpmath/mpmath/commit/8e31d101203eda0065ffcc011cb247a6e2c0ecd8#diff-d5e63243685c5f877d204d1882313e0217bec01a80b2cb58414be1f322f07247

If you agree, I think the fix could be to accept my current change as shown in this console which would remove the imports of `inf, nsum`:
https://github.com/mpmath/mpmath/pull/1113/conflicts

HeskethGD:
FYI, I have updated the above draft commit message.

skirpichev:
Thanks!

PR #1114, opened 2026-06-25
Fix chebyfit IndexError when N <= 0

gaoflow:
## Summary

`chebyfit(f, interval, N)` with `N=0` (or any negative integer) crashes
with an uninformative internal exception instead of a clear error:

```python
>>> from mpmath import chebyfit, sin
>>> chebyfit(sin, [-1, 1], 0)
IndexError: list index out of range
```

The `IndexError` originates at `d[0] = -c[0]/2` in `approximation.py`
because when `N=0` the list comprehension `[chebcoeff(...) for k in range(N)]`
produces an empty list, so index 0 is out of range.  With `N=-1` the
failure is a `TypeError` buried inside `chebcoeff`.

## Fix

Add a `ValueError` guard at the start of `chebyfit` so the failure is
immediate and informative: `ValueError: chebyfit requires N >= 1`.

A new test `test_chebyfit_nonpositive_N` covers `N=0` and `N=-3`.

---

This pull request was prepared with the assistance of AI, under my direction and review.

### DIFF
--- a/mpmath/tests/test_calculus.py
+++ b/mpmath/tests/test_calculus.py
@@ -26,8 +26,6 @@ def test_chebyfit():
 def test_chebyfit_nonpositive_N():
     with pytest.raises(ValueError):
         chebyfit(sin, [-1, 1], 0)
-    with pytest.raises(ValueError):
-        chebyfit(sin, [-1, 1], -3)
 
 def test_limits():
     assert limit(lambda x: (x-sin(x))/x**3, 0).ae(mpf(1)/6)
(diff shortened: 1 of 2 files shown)


skirpichev (approved):

mpmath/tests/test_calculus.py
```diff
@@ -23,6 +23,12 @@ def test_chebyfit():
         x = 2 + i/5.
         assert abs(polyval(p, x) - f(x)) < err
 
+def test_chebyfit_nonpositive_N():
+    with pytest.raises(ValueError):
+        chebyfit(sin, [-1, 1], 0)
+    with pytest.raises(ValueError):
+        chebyfit(sin, [-1, 1], -3)
```
skirpichev:
```suggestion
```

PR #1117, opened 2026-06-30
Add kleinjinv and support g and omega in taufrom

HeskethGD:
- Make `_inverse_kleinj` into public `kleinjinv` which takes `J` and returns `tau`
- Add support for `g2, g3` inputs to `kleinj` removing the redundant helper `‎_kleinj_from_g2g3`
- Add support for `g2, g3` and `omega1, omega2` inputs to `taufrom` removing the redundant helper `‎_tau_from_g`
- Update tests and docs where appropriate.

AI use declaration: Codex GPT5.5 was used to assist with mpmath style. All lines of code were read by me and iteratively adjusted.

### DIFF
--- a/docs/functions/elliptic.rst
+++ b/docs/functions/elliptic.rst
@@ -63,3 +63,4 @@ Modular functions
 
 .. autofunction:: mpmath.eta
 .. autofunction:: mpmath.kleinj
+.. autofunction:: mpmath.kleinjinv
--- a/mpmath/functions/elliptic.py
+++ b/mpmath/functions/elliptic.py
@@ -474,6 +474,79 @@ def kleinj(ctx, tau=None, **kwargs):
     return P/Q
 
 
+@defun_wrapped
+def kleinjinv(ctx, J):
+    r"""
+    Evaluates a branch of the inverse Klein j-invariant.
+
+    Given a value `J`, returns a half-period ratio `\tau` in the upper
+    half-plane such that ``kleinj(tau)`` equals `J`, up to numerical error.
+    Since ``kleinj`` is invariant under modular transformations, the inverse
+    is multivalued; this function returns one representative.
+    The branch is determined by principal square/cube roots in the formula
+    below; the returned value is one modularly equivalent preimage and is not
+    canonicalized to a fundamental domain.
+
+    The implementation uses the classical inverse construction via the
+    modular lambda function, described as Method 1 in the Wikipedia article
+    at https://en.wikipedia.org/wiki/J-invariant#Inverse_functions.
+
+    It first converts from the normalized invariant `J` to `j = 1728 J`,
+    then solves
+
+    .. math ::
+
+        j = 256 (1 - x)^3 / x^2, \quad x = \lambda (1 - \lambda),
+
+    for `lambda`.  The half-period ratio is then obtained from
+
+    .. math ::
+
+        \tau = i K(1 - \lambda) / K(\lambda),
+
+    where `K` is the complete elliptic integral.  Using
+
+    .. math ::
+
+        K(m) = \pi / (2 \operatorname{AGM}(1, \sqrt{1 - m})),
+
+    this is evaluated by the arithmetic-geometric mean formula
+
+    .. math ::
+
+        \tau =
+        i \operatorname{AGM}(1, \sqrt{1 - \lambda}) /
+        \operatorname{AGM}(1, \sqrt{\lambda}).
+
+    Different root choices give different modularly equivalent branches.
+
+    **Examples**
+
+        >>> from mpmath import mp, kleinj, kleinjinv, chop
+        >>> mp.dps = 25
+        >>> mp.pretty = True
+        >>> tau = 0.625+0.75j
+        >>> chop(kleinj(kleinjinv(kleinj(tau))) - kleinj(tau))
+        0.0
+        >>> kleinjinv(1)
+        (0.0 + 1.0j)
+
+    """
+    J = ctx.convert(J)
+    if J == 0:
+        return ctx.mpc(-ctx.mpf(1)/2, ctx.sqrt(3)/2)
+    _j = 1728 * J
+    sqrt_arg = 3*(1728*_j**2 - _j**3)
+    exponent = ctx.mpf(1) / ctx.mpf(3)
+    t = (-_j**3 + 2304*_j**2 - 884736*_j +
+         12288*ctx.sqrt(sqrt_arg))**exponent
+    x = ctx.mpf(1)/768*t + (1 - _j/768) - (1536*_j - _j**2) / (768*t)
+
+    lbd = (1 + ctx.sqrt(1 - 4*x)) / 2
+    tau = ctx.j * ctx.agm(1, ctx.sqrt(1-lbd)) / ctx.agm(1, ctx.sqrt(lbd))
+    return tau
+
+
 def RF_calc(ctx, x, y, z, r):
     if y == z: return RC_calc(ctx, x, y, r)
     if x == z: return RC_calc(ctx, y, x, r)
@@ -1518,23 +1591,6 @@ def _eisenstein_G4_G6(ctx, tau):
     G6 = 2 * ctx.zeta(6) * E6
     return G4, G6
 
-def _inverse_kleinj(ctx, J):
-    """
-    Compute tau from Klein's J-invariant using the inverse j-function.
-    See: https://en.wikipedia.org/wiki/J-invariant
-    """
-    J = ctx.convert(J)
-    _j = 1728 * J
-    sqrt_arg = 3*(1728*_j**2 - _j**3)
-    exponent = ctx.mpf(1) / ctx.mpf(3)
-    t = (-_j**3 + 2304*_j**2 - 884736*_j +
-         12288*ctx.sqrt(sqrt_arg))**exponent
-    x = ctx.mpf(1)/768*t + (1 - _j/768) - (1536*_j - _j**2) / (768*t)
-
-    lbd = (1 + ctx.sqrt(1 - 4*x)) / 2
-    tau = ctx.j * ctx.agm(1, ctx.sqrt(1-lbd)) / ctx.agm(1, ctx.sqrt(lbd))
-    return tau
-
 def _kleinj_from_g2g3(ctx, g2, g3):
     """
     Klein's absolute invariant J from g2, g3.
@@ -1552,7 +1608,7 @@ def _tau_from_g(ctx, g2, g3):
     g2 = ctx.convert(g2)
     g3 = ctx.convert(g3)
     J = _kleinj_from_g2g3(ctx, g2, g3)
-    tau = _inverse_kleinj(ctx, J)
+    tau = ctx.kleinjinv(J)
     return tau
 
 def _weierstrass_omega_tau(ctx, funcname, g2=None, g3=None, tau=None,
@@ -1636,7 +1692,7 @@ def weierhalfperiods(ctx, g2, g3):
         >>> chop(g2), chop(g3)
         (60.0, 140.0)
         >>> chop(omega2/omega1)
-        (0.5 + 0.209032224450873j)
+        (-0.5 + 0.209032224450873j)
 
     """
     with ctx.extraprec(10):
@@ -1935,7 +1991,7 @@ def weierpinv(ctx, p, g2=None, g3=None, tau=None, omega1=None, omega2=None,
     - `g2, g3`: elliptic invariants
     - `tau` or `omega1, omega2`: alternative parameterizations
     - `weierp_prime` (optional): derivative value used to choose the sign of
-       the inverse
+      the inverse
 
     **Examples**
 
--- a/mpmath/functions/elliptic.py
+++ b/mpmath/functions/elliptic.py
@@ -10,10 +10,11 @@ a function which satisfies
     f(z + 2 \omega_1) = f(z + 2 \omega_2) = f(z)
 
 for some half-periods `\omega_1, \omega_2` with
-`\mathrm{Im}[\omega_1 / \omega_2] > 0`. The canonical elliptic
-functions are the Jacobi elliptic functions. More broadly, this section
-includes  quasi-doubly periodic functions (such as the Jacobi theta
-functions) and other functions useful in the study of elliptic functions.
+`\mathrm{Im}[\omega_2 / \omega_1] > 0`. The classical examples include
+the Jacobi elliptic functions and the Weierstrass elliptic functions.
+More broadly, this section includes quasi-doubly periodic functions
+(such as the Jacobi theta functions) and other functions useful in the
+study of elliptic functions.
 
 Many different conventions for the arguments of
 elliptic functions are in use. It is even standard to use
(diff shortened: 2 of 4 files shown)


docs/functions/elliptic.rst
```diff
@@ -63,3 +63,4 @@ Modular functions
 
 .. autofunction:: mpmath.eta
 .. autofunction:: mpmath.kleinj
+.. autofunction:: mpmath.kleinjinv
```
skirpichev, line 66: This looks as a separate change wrt the taufrom().  Probably, it's better to open a new pr and revert this here.  Or there are reasons to keep both changes here?

I think we should have also some arguments beyond "we can expose a private functions" for addition of the kleinjinv.  I don't see this API in the FLINT or Mathematica.

mpmath/functions/elliptic.py
```diff
@@ -185,10 +185,12 @@ def qbarfrom(ctx, q=None, m=None, k=None, tau=None, qbar=None):
         return ctx.expjpi(2*tau)
 
 @defun_wrapped
-def taufrom(ctx, q=None, m=None, k=None, tau=None, qbar=None):
+def taufrom(ctx, q=None, m=None, k=None, tau=None, qbar=None,
+            g2=None, g3=None, omega1=None, omega2=None):
```
skirpichev, line 219: Shouldn't we then extend all foofrom() functions to have API like that?

I think that the module introduction must be extended with new parameters.

mpmath/functions/elliptic.py
```diff
@@ -473,6 +503,41 @@ def kleinj(ctx, tau=None, **kwargs):
     Q = 54*(t2*t3*t4)**8
     return P/Q
 
+@defun_wrapped
+def kleinjinv(ctx, J):
+    r"""
+    Evaluates a branch of the inverse Klein j-invariant.
```
skirpichev, line 558: Looks too vague.  Which branch?

skirpichev (commented):
I assume this is intentionally a draft.  Anyway, few comments.

docs/functions/elliptic.rst
HeskethGD: > This looks as a separate change wrt the taufrom(). Probably, it's better to open a new pr and revert this here. Or there are reasons to keep both changes here?

While it would not be too difficult to do two PRs, the `taufrom` change calls the `kleinjinv` so it would be good to have a decision on the latter to avoid duplicative work.

> I think we should have also some arguments beyond "we can expose a private functions" for addition of the kleinjinv. I don't see this API in the FLINT or Mathematica.

That's fair. I think the case for and against is something like:
AGAINST:
- Neither FLINK or Mathematica provide it which could indicate low demand.
- Additional code change

FOR:
- Neither FLINK or Mathematica provide it 
- we already have the code in place, we are using it ourselves to calculate `tau` from `J` under the hood and others may wish to also do that
- it is not just any private helper function we are exposing, it is the inverse of a function we already provide `kleinj`
- the J invariant wikipedia page has an entire section dedicated to the inverse and the several exact formulas that provide it
- researchers are writing papers about it - https://arxiv.org/pdf/2202.08189
image

mpmath/functions/elliptic.py
HeskethGD: Yeah, still working on that doc string have an edit to push.

mpmath/functions/elliptic.py
HeskethGD: > Shouldn't we then extend all foofrom() functions to have API like that?

Probably.

> I think that the module introduction must be extended with new parameters.

Yep.

HeskethGD: Done

mpmath/functions/elliptic.py
HeskethGD: I commented on this. Cannot be more specific re a branch than the docstring says but it is transparent as to the process.

docs/functions/elliptic.rst
HeskethGD: If `klienjinv` wont be made public, I'd probably just change to `_klienjinv` and make private in this PR rather than undo so as to keep the better name and improved docs.

mpmath/functions/elliptic.py
```diff
@@ -120,10 +125,72 @@ def nome(ctx, m):
...
 
+def _validate_weierstrass_pair_args(funcname, g2, g3, omega1, omega2):
+    if (g2 is None) != (g3 is None):
+        raise ValueError("%s: must provide both g2 and g3" % funcname)
+    if (omega1 is None) != (omega2 is None):
+        raise ValueError("%s: must provide both omega1 and omega2" % funcname)
+
 @defun_wrapped
-def qfrom(ctx, q=None, m=None, k=None, tau=None, qbar=None):
+def taufrom(ctx, q=None, m=None, k=None, tau=None, qbar=None,
```
skirpichev: Why this was moved around?

mpmath/functions/elliptic.py
```diff
@@ -379,7 +447,7 @@ def f(*args, **kwargs):
     return +v
 
 @defun_wrapped
-def kleinj(ctx, tau=None, **kwargs):
+def kleinj(ctx, tau=None, g2=None, g3=None, **kwargs):
```
skirpichev: I think this could go to a separate PR.

And we should drop kwargs.  Arguments should be specified explicitly.  And we must decide which signature we prefer:
1. qfrom-like, as in ellipfun()
2. signature with just one specific parametrization, e.g. tau.  Like for jtheta().

I think that ellipfun-like API should be limited to parameter conversion functions.

mpmath/functions/elliptic.py
```diff
@@ -1572,8 +1674,9 @@ def _weierstrass_omega_tau(ctx, funcname, g2=None, g3=None, tau=None,
     if omega1 is not None:
         omega1 = ctx.convert(omega1)
         omega2 = ctx.convert(omega2)
-        tau = omega2 / omega1
-        if ctx.im(tau) <= 0:
+        try:
+            tau = ctx.taufrom(omega1=omega1, omega2=omega2)
+        except ValueError:
             raise ValueError("%s: omega ratio must be in upper half-plane" %
```
skirpichev: I think now it's redundant, no?

mpmath/functions/elliptic.py
```diff
@@ -50,8 +54,9 @@
 
     \bar{q} = q^2 = e^{2 i \pi \tau}
 
-For convenience, mpmath provides functions to convert
-between the various parameters (:func:`~mpmath.qfrom`, :func:`~mpmath.mfrom`,
+For convenience, mpmath provides functions to convert between the various
+parameters, including half-periods and Weierstrass invariants where
+applicable (:func:`~mpmath.qfrom`, :func:`~mpmath.mfrom`,
```
skirpichev, line 59: Probably, you should mention weierinvariants/halfperiods here as well.

mpmath/functions/elliptic.py
HeskethGD: Only because the latter functions were calling `taufrom` now and I thought it would be more readable in that order and to have `kleinjinv` next to `kleinj`. Admittedly, it wasn't really necessary, let me know if you prefer that I revert the order and I will change it back.

mpmath/functions/elliptic.py
HeskethGD: > And we must decide which signature we prefer:
> 
> 1. qfrom-like, as in ellipfun()
> 2. signature with just one specific parametrization, e.g. tau. Like for jtheta().
> 
> I think that ellipfun-like API should be limited to parameter conversion functions.

In this particular instance `kleinj` is being calculated quite differently by either `tau` or `g2,g3` and there is a simple polynomial for `kleinj` from `g2,g3`. It would be unnecessarily inefficient to calculate `kleinj` from `tau` when used in that path. 

> I think this could go to a separate PR.

The `taufrom` change is calling `kleinj` by `g2,g3` so that seems like duplicating changes on the same code if that is split into a separate pr. In fact, I'm not even sure how you would implement the `taufrom` by `g2,g3` path if `kleinj` only takes `tau`, we'd be back to the separate helper that calculates `kleinj` by `g2,g3` independently.

> And we should drop kwargs. Arguments should be specified explicitly. 

The kwargs chnage looks like a bigger change that is somewhat separate to this main aim so perhaps would be better in a new PR.

mpmath/functions/elliptic.py
HeskethGD: Pretty much, the only reason it was kept was to display a message that referenced that function the other choice was to bubble up from `taufrom` which could have been just as valid a choice. Happy to change it.

mpmath/functions/elliptic.py
HeskethGD: I was planning another PR that will optimise the way `omega1, omega2` are calculated as the current code has some legacy opinions on ordering to match a particular order of roots which I do not think is now needed and simplifying it could offer significant speed up and more readable code. I would then look to document the way `g2, g3` and `omega1, omega2` are calculated after that change so I would defer further modification here until that PR.

mpmath/functions/elliptic.py
HeskethGD: Actually, re kwargs, on closer inspection I see that the historical `kleinj` use of kwargs is the only public use in this file and its role was only to forward them to `qfrom`. For that reason, I have modified the signature to take the args explicitly, keeping `tau` first to avoid breaking `kleinj(tau)` calls, and prefering to calculate from `g2, g3` from the polynomial when they are available. There are now no public instances of `kwargs` in this file.

mpmath/functions/elliptic.py
HeskethGD: I have removed the error wrapper.

mpmath/functions/elliptic.py
skirpichev: Yes, lets reduce diff a bit.

mpmath/functions/elliptic.py
```diff
@@ -465,14 +539,82 @@ def kleinj(ctx, tau=None, **kwargs):
         1264538.909475140509320227
 
     """
-    q = ctx.qfrom(tau=tau, **kwargs)
+    _validate_weierstrass_pair_args("kleinj", g2, g3, omega1, omega2)
+    if g2 is not None:
+        g2 = ctx.convert(g2)
+        g3 = ctx.convert(g3)
+        return g2**3 / (g2**3 - 27*g3**2)
+    q = ctx.qfrom(q=q, m=m, k=k, tau=tau, qbar=qbar,
```
skirpichev, line 545: But qfrom supports g2/g3, why reinvent handling of arguments here?

HeskethGD: For `g2,g3`, qfrom calls taufrom which calls this by `g2,g3`, previously there was a separate helper that duplicated it. See this thread for related comments: https://github.com/mpmath/mpmath/pull/1117#discussion_r3525025673

mpmath/functions/elliptic.py
HeskethGD: Done. Diff reduced.

mpmath/functions/elliptic.py
```diff
@@ -218,12 +253,27 @@ def taufrom(ctx, q=None, m=None, k=None, tau=None, qbar=None):
     if qbar is not None:
         qbar = ctx.convert(qbar)
         return ctx.log(qbar) / (2*ctx.pi*ctx.j)
+    if g2 is not None:
+        g2 = ctx.convert(g2)
+        g3 = ctx.convert(g3)
```
skirpichev:
```suggestion
```

mpmath/functions/elliptic.py
```diff
@@ -1636,7 +1738,7 @@ def weierhalfperiods(ctx, g2, g3):
         >>> chop(g2), chop(g3)
         (60.0, 140.0)
         >>> chop(omega2/omega1)
-        (0.5 + 0.209032224450873j)
+        (-0.5 + 0.209032224450873j)
```
skirpichev, line 1750: Why this was changed?

skirpichev (commented):
Mostly LGTM, just one question.

mpmath/functions/elliptic.py
HeskethGD: This is a subtle one but it was as follows: 
- The issue was noticed when automated tests observed an error in the docstring value after a change.
- The old value in the docstring with +0.5 is modularly equivalent to the new value with -0.5 however...
- When we were using the old helper `_kleinj_from_g2g3` we wrote a formula that looked like `1 / (1 - 27*g3**2/g2**3)` and this path returned `0.5 + 0.209032224450873j`
- When we folded it into the `kleinj` we followed the formula that was already in the docstring `g_2^3 / (g_2^3 - 27 g_3^2)` which is mathematically identical, but not necessarily bit-for-bit identical at finite precision. For `(g2, g3) = (60, 140)`, inside the extraprec(10) context used by weierhalfperiods, there are differences on the order of 1e-19 in the resulting `tau`.
- This then feeds into the `weierhalfperiods` which calls `_roots_from_omega` which has sorting logic on roots. For `(g2, g3) = (60, 140)` we get one real and two complex conjugate roots with the "same" real part. But if the real parts fluctuate they can be returned in slightly different order. This then leads to `weierhalfperiods` returning half periods corresponding to a different cell and thus a different but modularly equivalent `tau`.
- Consequently the value was updated in the docstring having established that it was modularly equivalent and not a fundamental bug.

As an aside, I intend to follow up with a new PR that will review the possibly redundant root sorting logic, the purpose of which was to attempt to make `omega1` the real period for integrable systems use but it is not necessarily the best option here. My intention was to test for speed up and return half periods that match against Wolfram Engine with numeric tests to confirm.

skirpichev: > The old value in the docstring with +0.5 is modularly equivalent to the new value with -0.5 however...

As new functions not included yet into a stable release, I think it's fine to break compatibility.

> When we folded it into the kleinj we followed the formula that was already in the docstring g_2^3 / (g_2^3 - 27 g_3^2)

Thanks for confirmation, this was also my guess.

skirpichev (approved):
LGTM

I could merge this as is, but would prefer if you rebase PR to two commits (introduction of the kleinjinv and changes in *from helpers).  Let me know if you can do this.

HeskethGD:
> LGTM
> 
> I could merge this as is, but would prefer if you rebase PR to two commits (introduction of the kleinjinv and changes in *from helpers). Let me know if you can do this.

Can I check what this would likely involve so that we are aligned before I make changes. Would it be something like:
```
git branch backup-before-two-commit-rebase

git reset --soft origin/master
git reset

# stage only kleinjinv-related hunks
git add -p ...
git commit -m "Add inverse Klein j-function"

git push --force-with-lease
# wait for Actions

# stage remaining helper changes
git add ...
git commit -m "Support Weierstrass data in elliptic parameter helpers"

git push
# wait for Actions
```

skirpichev:
Ideally, yes.  Every commit should pass CI.

Though, if you factor out just changes in code+docs for kleinjinv() (not tests) - I would appreciate this.

HeskethGD:
> LGTM
> 
> I could merge this as is, but would prefer if you rebase PR to two commits (introduction of the kleinjinv and changes in *from helpers). Let me know if you can do this.

This has been done.

skirpichev (approved):

skirpichev:
Thanks.

PR #1121, opened 2026-07-05
ENH: Implement the modified spherical bessel functions spherical_in() and spherical_kn().

WarrenWeckesser:
The implementations of the two new functions follow exactly the same patterns as the existing `spherical_jn()` and `spherical_yn()`.

FWIW, SciPy has these two function in its `scipy.special` module.


mpmath/tests/test_functions2.py
```diff
@@ -90,6 +90,25 @@ def test_bessel():
     assert besselk(0,j).ae(-0.13863371520405399968-1.20196971531720649914j)
     assert (besselk(3, 10**10) * mpf(10)**4342944824).ae(1.1628981033356187851)
     assert besselk(1,inf) == 0
+
+    # Wolfram Alpha 'BesselI(1/2, 1) * sqrt(Pi/2)':
+    #   1.1752011936438014568823818505956008151557179813340958702295654130...
```
skirpichev: I'm not sure if this is reproducible.  Could you use Mathematica or Wolfram Engine to compute reference values?

WarrenWeckesser: Done.  I used Wolfram Engine 15.

mpmath/tests/test_functions2.py
```diff
@@ -90,6 +90,29 @@ def test_bessel():
     assert besselk(0,j).ae(-0.13863371520405399968-1.20196971531720649914j)
     assert (besselk(3, 10**10) * mpf(10)**4342944824).ae(1.1628981033356187851)
     assert besselk(1,inf) == 0
+
+    # Wolfram Engine 15:
+    # 'N[BesselI[1/2, 1] * Sqrt[Pi/2], 30]' gives
+    # 1.17520119364380145688238185060
```
skirpichev: I think you should describe here *how* the reference value was computed, don't repeat it.

WarrenWeckesser: As far as I can tell, Wolfram doesn't have implementations of the modified spherical Bessel functions, so to get the reference values, I'm using the same expression in wolframscript that I use here in Python code for mpmath.  That means I'm the author of both formulas, and we're using wolfram just to check that these two implementations give the same result. (The advantage of using Wolfram instead of some other library is that we can easily get a high precision result.) This is not as strong as if Wolfram had its own implementation of these functions--maybe I implemented the same incorrect formula in both places!--but it's better than nothing, and hopefully the wise reviewer would catch such a mistake :smile:.

So that's "how" the reference values are computed: evaluate that expression with Wolfram Engine. I'm not sure what you would like to change, other than perhaps removing the redundant numerical result from the comments.

I was more-or-less following the pattern established in the existing test, but with comments to explain where the reference values come from.  I would be happy to break these tests out of `test_bessel()` into one or more separate tests, use higher precision, use `pytest.mark.parametrize(...)` where appropriate, etc.  Let me know if you are interested in a more involved change like that.

WarrenWeckesser: For what it's worth...

I've been experimenting with wolframscript since my last push to this PR.  This short script

```
PrintNumber[z_, n_Integer?Positive, label_String : ""] := Print[label, ToString[NumberForm[N[z, n], n]]];

SphericalIn[n_, z_] := BesselI[n + 1/2, z] * Sqrt[Pi / (2*z)]
SphericalKn[n_, z_] := BesselK[n + 1/2, z] * Sqrt[Pi / (2*z)]

digits = 24
PrintNumber[SphericalIn[0, 1], digits, "SphericalIn[0, 1]         = "]
PrintNumber[SphericalIn[6, -3/2 + 2I], digits, "SphericalIn[6, -3/2 + 2I] = "]
PrintNumber[SphericalKn[0, 1], digits, "SphericalKn[0, 1]         = "]
PrintNumber[SphericalKn[6, -3/2 + 2I], digits, "SphericalKn[6, -3/2 + 2I] = "]
```

prints

```
SphericalIn[0, 1]         = 1.17520119364380145688238
SphericalIn[6, -3/2 + 2I] = 0.00148388231096733251173089 + 0.00084586141172470688312022 I
SphericalKn[0, 1]         = 0.577863674895460858955047
SphericalKn[6, -3/2 + 2I] = -25.4279100776794722746615 - 13.3888853002501422390151 I
```

skirpichev: > As far as I can tell, Wolfram doesn't have implementations of the modified spherical Bessel functions, so to get the reference values, I'm using the same expression in wolframscript that I use here in Python code for mpmath.

That's fine.  Just describe how these values were computed.  No need to repeat them in comments.

WarrenWeckesser: OK, updated.

skirpichev (approved):

WarrenWeckesser:
Thanks @skirpichev.

PR #1132, opened 2026-07-16
Fix last-digit rounding of the "f" format type

### LINKED ISSUE #1131: Incorrectly rounded formatting with "f" format type
Examples could be produced e.g. by following test:
```pycon
>>> import gmpy2, mpmath
... for _a in range(1000000):
...     x = mpmath.rand()
...     x_h = f"{x:a}"
...     gx = gmpy2.mpfr(x_h)
...     for gm in ['N', 'U', 'D']:
...         for dps in [15, 21, 30, 46, 100]:
...             fmt = f".{dps}{gm}f"
...             m_res = format(x, fmt)
...             g_res = format(gx, fmt)
...             if m_res != g_res:
...                 print(f"{gm} {x:a} {fmt}")
...
'N' 0x1.52e639e794efdp-1 '.21Nf'
'U' 0x1.42696b91b17cap-1 '.21Uf'
'Y' 0x1.42696b91b17cap-1 '.21Yf'
>>> 
```

Not occurs with "e" type (i.e. using ``fmt = f".{dps}{gm}e"`` or ``m_res = mpmath.nstr(x, dps+1, strip_zeros=False, min_fixed=float('inf'), max_fixed=float('-inf'), rnd=mpmath.mp._MPFR_rounding_map[gm])``).

More examples:
```
'N' 0x1.52e639e794efdp-1 '.21Nf'
'U' 0x1.42696b91b17cap-1 '.21Uf'
'Y' 0x1.42696b91b17cap-1 '.21Yf'
'U' 0x1.605b39ff191e1p-1 '.15Uf'
'Y' 0x1.605b39ff191e1p-1 '.15Yf'
'U' 0x1.72a03c916e4c8p-1 '.30Uf'
'N' 0x1.ec33380df762cp-1 '.21Nf'
'U' 0x1.7a4c433af45ap-4 '.21Uf'
'U' 0x1.51c5f179176a4p-2 '.21Uf'
'N' 0x1.6f81e644335f2p-2 '.21Nf'
'U' 0x1.3c38bbc793bf4p-3 '.30Uf'
'Y' 0x1.3c38bbc793bf4p-3 '.30Yf'
'U' 0x1.fef1f6e96e4c8p-1 '.30Uf'
'Y' 0x1.fef1f6e96e4c8p-1 '.30Yf'
'U' 0x1.c6b77b2a23d2p-3 '.21Uf'
'Y' 0x1.c6b77b2a23d2p-3 '.21Yf'
'U' 0x1.4ea9e44a5b932p-1 '.30Uf'
'Y' 0x1.4ea9e44a5b932p-1 '.30Yf'
'N' 0x1.b4c6f1e047ffdp-1 '.15Nf'
'U' 0x1.e804179a90e09p-1 '.15Uf'
'Y' 0x1.e804179a90e09p-1 '.15Yf'
'U' 0x1.253d77b16e4c8p-1 '.30Uf'
'Y' 0x1.253d77b16e4c8p-1 '.30Yf'
'U' 0x1.0679aa06b5fd2p-2 '.15Uf'
'Y' 0x1.0679aa06b5fd2p-2 '.15Yf'
'U' 0x1.7d6acf035d7e1p-1 '.21Uf'
'Y' 0x1.7d6acf035d7e1p-1 '.21Yf'
'U' 0x1.83ce371bf577p-1 '.15Uf'
'Y' 0x1.83ce371bf577p-1 '.15Yf'
'N' 0x1.179d1e92ea602p-1 '.15Nf'
'N' 0x1.6041f3e2de70dp-1 '.15Nf'
'N' 0x1.d909854abc6acp-2 '.21Nf'
'N' 0x1.98c43a712dc99p-1 '.30Nf'
'U' 0x1.7904d5c711986p-2 '.15Uf'
'U' 0x1.d2a4000f28c2fp-1 '.15Uf'
'U' 0x1.4d56aa96931dp-2 '.21Uf'
'N' 0x1.b2675a9e0663fp-1 '.15Nf'
'U' 0x1.47431590e918p-1 '.15Uf'
'U' 0x1.60da46ef3e1f9p-1 '.21Uf'
'U' 0x1.9cfd0eb8754eap-2 '.15Uf'
```

gaoflow:
The `f` type rounds the last digit incorrectly under directed/nearest rounding, while the `e` path and MPFR round the same value correctly:

```pycon
>>> from mpmath import mpf
>>> x = mpf(float.fromhex('0x1.605b39ff191e1p-1'))   # 0.68819600333204900000083...
>>> format(x, '.15Uf')     # f, toward +inf
'0.688196003332049'
>>> format(x, '.14Ue')     # e, same value
'6.88196003332050e-01'
```

| input | spec | `f` (before) | `e` / MPFR |
|---|---|---|---|
| `0x1.605b39ff191e1p-1` | `.15Uf` | `…049` | `…050` |
| `0x1.52e639e794efdp-1` | `.21Nf` | `…541160` | `…541161` |
| `0x1.7a4c433af45ap-4`  | `.21Uf` | `…733410` | `…733411` |

`format_fixed()` extracts only `dps + O(1)` guard digits and lets `round_digits()` decide. When the true remainder lies past those digits (every extracted guard digit is `0`), the value looks exact at the cutoff and directed/nearest rounding truncates instead of rounding up. `format_scientific()` escapes this only by requesting a wider window, so raising the guard constant would just move the boundary.

Since the value is dyadic, whether the extracted digits are exact is decidable directly (`exp + fractional_digits < 0`). I pass that as a `sticky` flag into `round_digits()` and round on it, so `f` agrees with `e` and MPFR by construction rather than by a guard-digit margin.

Testing turned up a second defect in the same function: a value whose leading digit falls past the last requested place took an early return that printed only zeros and ignored the rounding mode, e.g. `format(mpf(0.0004), '.1Uf')` gave `0.0` where `e`/MPFR give `0.1`. Rounded away from zero in that branch too.

Checked against a gmpy2 differential over a wide range of doubles, precisions and all five rounding modes (no mismatches); the existing suite still passes. Regression tests added in `test_format.py`.

Fixes #1131.

### DIFF
--- a/mpmath/libmp/libmpf.py
+++ b/mpmath/libmp/libmpf.py
@@ -1086,12 +1086,12 @@ def to_digits_exp(s, dps, base=10):
     return sign, digits, exponent
 
 def round_digits(sign, digits, dps, base, rnd=round_down, fixed=False,
-                 sticky=False):
+                 inexact=False):
     """
     Returns the rounded digits, and the number of places the decimal point was
     shifted.
 
-    ``sticky`` marks that the value has a nonzero remainder past the last
+    ``inexact`` marks that the value has a nonzero remainder past the last
     supplied digit, so directed and nearest rounding round it up even when
     every extracted guard digit is zero.
     """
@@ -1121,7 +1121,7 @@ def round_digits(sign, digits, dps, base, rnd=round_down, fixed=False,
         # The first digit after dps is a 5 and we should determine whether we
         # round it up or down.
         if digits[dps] == rnd_digs[0]:
-            tie_down = not sticky
+            tie_down = not inexact
 
             # If the digit we round to is even, we may round down if all the
             # following digits are 0.
@@ -1136,7 +1136,7 @@ def round_digits(sign, digits, dps, base, rnd=round_down, fixed=False,
     elif rnd == round_up:
         # If any digit following a 0 is different from zero, we round up.
         if digits[dps] == '0':
-            tie_up = sticky
+            tie_up = inexact
             for i in range(dps+1, len(digits)):
                 if digits[i] != '0':
                     tie_up = True
@@ -1491,43 +1491,30 @@ def format_fixed(s, dps, rnd=round_down):
             s, max(dps+exponent+4, int(s[3]/blog2_10)), base)
     # to_digits_exp truncates; flag a nonzero remainder past the last digit so
     # rounding is not fooled by a short zero tail (#1131).
-    sticky = s[2] + len(digits) - 1 - exponent < 0
+    inexact = s[2] + len(digits) - 1 - exponent < 0
     orig_dps = dps
     dps += exponent + 1
 
-    # The value's leading digit is past the last requested place, so it rounds
-    # to 0 -- or to a single unit there when rounding away from zero (#1131).
     if dps < 0:
-        if rnd == round_ceiling:
-            roundup = not s[0]
-        elif rnd == round_floor:
-            roundup = bool(s[0])
-        else:
-            roundup = rnd == round_up
-        digits = '1'.rjust(orig_dps + 1, '0') if roundup else '0'*(orig_dps + 1)
-        int_part, frac_part = digits[0], digits[1:]
+        # The number we want to print is lower in magnitude that the
+        # requested precision.
+        digits = '0'*(-dps) + digits
+        exponent -= dps
+        dps = 0
 
+    digits, exp_add = round_digits(s[0], digits, dps, base, rnd, True, inexact)
+    exponent += exp_add
+
+    # Here we prepend the corresponding 0s to the digits string, according
+    # to the value of exponent
+    split = 1
+    if exponent < 0:
+        digits = "0"*(-exponent) + digits
     else:
-        digits, exp_add = round_digits(s[0], digits, dps, base, rnd, True,
-                                       sticky)
-        exponent += exp_add
+        split += exponent
 
-        # Here we prepend the corresponding 0s to the digits string, according
-        # to the value of exponent
-        if exponent < 0:
-            digits = ("0"*(-exponent)) + digits
-            split = 1
-        else:
-            split = exponent + 1
-        int_part = digits[:split]
-
-        # Finally, assemble the digits including the decimal point
-        if orig_dps == 0:
-            return int_part, ''
-
-        frac_part = digits[split:]
-
-    return int_part, frac_part
+    # Finally, assemble the digits including the decimal point
+    return digits[:split], digits[split:]
 
 
 def format_scientific(s, dps, rnd=round_down):
--- a/mpmath/libmp/libmpf.py
+++ b/mpmath/libmp/libmpf.py
@@ -1085,10 +1085,15 @@ def to_digits_exp(s, dps, base=10):
     exponent += len(digits) - fixdps - 1
     return sign, digits, exponent
 
-def round_digits(sign, digits, dps, base, rnd=round_down, fixed=False):
+def round_digits(sign, digits, dps, base, rnd=round_down, fixed=False,
+                 sticky=False):
     """
     Returns the rounded digits, and the number of places the decimal point was
     shifted.
+
+    ``sticky`` marks that the value has a nonzero remainder past the last
+    supplied digit, so directed and nearest rounding round it up even when
+    every extracted guard digit is zero.
     """
 
     assert len(digits) > dps
@@ -1116,7 +1121,7 @@ def round_digits(sign, digits, dps, base, rnd=round_down, fixed=False):
         # The first digit after dps is a 5 and we should determine whether we
         # round it up or down.
         if digits[dps] == rnd_digs[0]:
-            tie_down = True
+            tie_down = not sticky
 
             # If the digit we round to is even, we may round down if all the
             # following digits are 0.
@@ -1131,6 +1136,7 @@ def round_digits(sign, digits, dps, base, rnd=round_down, fixed=False):
     elif rnd == round_up:
         # If any digit following a 0 is different from zero, we round up.
         if digits[dps] == '0':
+            tie_up = sticky
             for i in range(dps+1, len(digits)):
                 if digits[i] != '0':
                     tie_up = True
@@ -1483,17 +1489,27 @@ def format_fixed(s, dps, rnd=round_down):
     # exponent by +- 1)
     _, digits, exponent = to_digits_exp(
             s, max(dps+exponent+4, int(s[3]/blog2_10)), base)
+    # to_digits_exp truncates; flag a nonzero remainder past the last digit so
+    # rounding is not fooled by a short zero tail (#1131).
+    sticky = s[2] + len(digits) - 1 - exponent < 0
     orig_dps = dps
     dps += exponent + 1
 
-    # The number we want to print is lower in magnitude that the requested
-    # precision. We should only print 0s.
+    # The value's leading digit is past the last requested place, so it rounds
+    # to 0 -- or to a single unit there when rounding away from zero (#1131).
     if dps < 0:
-        int_part = '0'
-        frac_part = orig_dps*'0'
+        if rnd == round_ceiling:
+            roundup = not s[0]
+        elif rnd == round_floor:
+            roundup = bool(s[0])
+        else:
+            roundup = rnd == round_up
+        digits = '1'.rjust(orig_dps + 1, '0') if roundup else '0'*(orig_dps + 1)
+        int_part, frac_part = digits[0], digits[1:]
 
     else:
-        digits, exp_add = round_digits(s[0], digits, dps, base, rnd, True)
+        digits, exp_add = round_digits(s[0], digits, dps, base, rnd, True,
+                                       sticky)
         exponent += exp_add
 
         # Here we prepend the corresponding 0s to the digits string, according
--- a/mpmath/tests/test_format.py
+++ b/mpmath/tests/test_format.py
@@ -916,16 +916,16 @@ def test_hexadecimal_fmt():
 
 @given(st.floats(allow_nan=False, allow_infinity=False,
                  allow_subnormal=False),
-       st.integers(min_value=1, max_value=40))
-def test_fixed_with_gmpy2_bulk(x, p):
+       st.integers(min_value=1, max_value=40),
+       st.sampled_from(list('UDNYZ')))
+def test_fixed_with_gmpy2_bulk(x, dps, mode):
     gmpy2 = pytest.importorskip('gmpy2')
     if not x and math.copysign(1, x) == -1:
         return  # skip negative zero
+    fmt = f'.{dps}{mode}f'
     gx = gmpy2.mpfr(x)
     mx = mp.mpf(x)
-    for rnd in 'NUDYZ':
-        fmt = '.%d%sf' % (p, rnd)
-        assert format(mx, fmt) == format(gx, fmt)
+    assert format(mx, fmt) == format(gx, fmt)
 
 
 def test_issue_1131():
--- a/mpmath/tests/test_format.py
+++ b/mpmath/tests/test_format.py
@@ -930,17 +930,17 @@ def test_fixed_with_gmpy2_bulk(x, p):
 
 def test_issue_1131():
     # 'f' formatting must round the last digit like the 'e' path and MPFR do.
-    # ~0.000464, below 0.1 unit in the last place at .1f:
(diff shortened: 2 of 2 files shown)


mpmath/libmp/libmpf.py
```diff
@@ -1483,17 +1489,27 @@ def format_fixed(s, dps, rnd=round_down):
...
-        int_part = '0'
-        frac_part = orig_dps*'0'
+        if rnd == round_ceiling:
+            roundup = not s[0]
+        elif rnd == round_floor:
+            roundup = bool(s[0])
+        else:
+            roundup = rnd == round_up
+        digits = '1'.rjust(orig_dps + 1, '0') if roundup else '0'*(orig_dps + 1)
+        int_part, frac_part = digits[0], digits[1:]
```
skirpichev: No, this is a wrong place for that.  It rather belongs to the round_digits().

mpmath/libmp/libmpf.py
```diff
@@ -1483,17 +1489,27 @@ def format_fixed(s, dps, rnd=round_down):
     # exponent by +- 1)
     _, digits, exponent = to_digits_exp(
             s, max(dps+exponent+4, int(s[3]/blog2_10)), base)
+    # to_digits_exp truncates; flag a nonzero remainder past the last digit so
+    # rounding is not fooled by a short zero tail (#1131).
+    sticky = s[2] + len(digits) - 1 - exponent < 0
```
skirpichev: Before introducing a new flag, did you try to increase dps value in the to_digits_exp() call above?

mpmath/tests/test_format.py
```diff
@@ -912,3 +912,50 @@ def test_hexadecimal_fmt():
...
+    ]
+    for x, fmt, expected in cases:
+        assert format(mp.mpf(x), fmt) == expected, (x.hex(), fmt)
+
+    # exact dyadic half-way ties still round to even; carries still propagate
+    assert format(mp.mpf('0.125'), '.2Nf') == '0.12'
+    assert format(mp.mpf('0.375'), '.2Nf') == '0.38'
+    assert format(mp.mpf('2.5'), '.0Nf') == '2'
+    assert format(mp.mpf('3.5'), '.0Nf') == '4'
+    assert format(mp.mpf('0.6999999999'), '.4Uf') == '0.7000'
```
skirpichev, line 962: In which sense that is an exact dyadic?

mpmath/tests/test_format.py
```diff
@@ -912,3 +912,50 @@ def test_hexadecimal_fmt():
...
+    mx = mp.mpf(x)
+    for rnd in 'NUDYZ':
+        fmt = '.%d%sf' % (p, rnd)
+        assert format(mx, fmt) == format(gx, fmt)
+
+
+def test_issue_1131():
+    # 'f' formatting must round the last digit like the 'e' path and MPFR do.
+    # ~0.000464, below 0.1 unit in the last place at .1f:
+    tiny = float.fromhex('0x1.e6a84962cb8p-12')
```
skirpichev: This is rather a random number, why you prefer hexadecimal form over literal value?

skirpichev (commented):
Note, that issue was assigned.  I did pr: https://github.com/mpmath/mpmath/pull/1133

Please argue why this one is better.

binggao1230:
First — sorry for jumping on the issue right after you filed it; I didn't notice you had assigned #1131 to yourself.

To the question: #1133 widens the extraction window by a fixed margin, and no fixed margin is sufficient at arbitrary precision, because the zero run after the requested place can grow like ~0.3*(prec - dps). Constructed case (remainder = 2^-170, i.e. 51 zero digits after place 30 and the first nonzero digit at place 82, while the window extracts ~70 digits):

```pycon
>>> from mpmath import mp
>>> mp.prec = 210
>>> m = 803469023414207548062917070783370784802306362831722016595097
>>> x = mp.mpf(m) / 2**200    # m = 5**-30 mod 2**170, lifted to 200 bits
>>> format(x, '.30Uf')
```

#1133 returns `'0.500000000799478495690452934926'`; this PR and gmpy2 at 210-bit precision return `'...927'`, which is correct: frac(x*10^30) = 2^-170 > 0, so rounding toward +inf must bump the last place. The construction scales — for any margin G it fails again once prec - dps > ~3.3*G. Doubles can't reach that boundary at dps <= 100, which is why the float-based gmpy2 fuzz doesn't catch it.

Extracting enough digits to be provably safe means going until the dyadic expansion terminates, about 0.3*|exp| digits — unbounded work for tiny exponents. The exponent already tells you whether the extracted digits are exact; that is all the sticky flag is.

That said, your restructuring of format_fixed (zero-prefixing the dps < 0 case into round_digits) is cleaner than my special case. Simplest path is probably to add the sticky bit to #1133 and close this one — fine by me. Or I can rework this PR on top of your structure, whichever you prefer.

mpmath/libmp/libmpf.py
binggao1230: I did — any fixed increase moves the boundary rather than removing it: the zero run after the cutoff grows like ~0.3*(prec - dps), so it fails again at higher precision. Concrete case in the main thread.

mpmath/libmp/libmpf.py
binggao1230: Fair — your zero-prefix version in #1133 handles this more uniformly. Happy to adopt that shape.

mpmath/tests/test_format.py
binggao1230: Sloppy wording on my part: meant the tie is exact — 0.125 terminates right at the '5', so it exercises round-half-even rather than the sticky path. 'exact tie' would be clearer.

mpmath/tests/test_format.py
binggao1230: Only to keep the bit pattern explicit; decimal literals round-trip doubles exactly, so happy to switch them.

mpmath/libmp/libmpf.py
skirpichev: Or adapt round_digits().

mpmath/tests/test_format.py
skirpichev: Please fix either code or commentary.

skirpichev:
> https://github.com/mpmath/mpmath/pull/1133 widens the extraction window by a fixed margin, and no fixed margin is sufficient at arbitrary precision

Yes, unfortunately this depends on the exponent of the mpf.

> Simplest path is probably to add the sticky bit to https://github.com/mpmath/mpmath/pull/1133 and close this one — fine by me.

The way you *set* it looks like a heuristic.

Edit: I also don't like name of the flag.  How about something like ``exact``, having clear meaning that all digits are exact for decimal expansion of the given mpf?  If not, assuming that digits come from to_digits_exp(), we know that they are rounded toward zero.

mpmath/libmp/libmpf.py
skirpichev: I doubt this will work.

We have two cases.  Assuming `s = (sign, m, e, bc)`,
1. `e >= 0`: to_digits_exp() exact when `len(digits) >= exponent + 1`
2. `e < 0`: it's exact when `len(digits) >= exponent + 1 - e`

The #1133 requests appropriate number of digits, max of above cases.  But you don't, in the (2) case: `bc/blog2_10` is a wrong bound for `exponent - e`.  So, if inequality in (2) is false - you can't say, that digits were truncated.  Expansion can be exact, you just asked too few digits to be sure.

binggao1230: The inference doesn't need the request to cover the full expansion — that's what the flag replaces. mpf tuples are normalized, so `m` is odd, and for `e < 0` the exact expansion terminates with a 5 at position `10^e` exactly: `exponent + 1 - e` is the exact significant length, not an upper bound. So `len(digits) < exponent + 1 - e` means the window stopped short of the terminating digit, and the dropped tail is necessarily nonzero — "asked too few digits" and "truncated with nonzero remainder" coincide here.

For `e >= 0` the window can't stop short at all: `numeral()` never drops integer digits, so `len(digits) >= exponent + 1` always holds and the flag stays False.

`bc/blog2_10` isn't meant to bound `exponent - e`; it's only a floor on the window so short-`dps` requests still get their guard digits.

Rechecked against an exact `Fraction` oracle: 40k extractions, `bc <= 260`, `e in [-300, 40]`, deliberately undersized dps requests (1–40) — zero disagreements between the flag and the exact remainder, on top of the earlier 120k-extraction and 7.9M-format sweeps.

mpmath/tests/test_format.py
binggao1230: Switched to decimal literals in d4413be.

mpmath/tests/test_format.py
binggao1230: Reworded in d4413be — the terminating-5 cases and the carry case are now labeled separately.

mpmath/libmp/libmpf.py
skirpichev: > mpf tuples are normalized, so m is odd

Ah, good catch.  Now it's obvious, that `exponent + 1 + max(0, -e)` is the exact bound.  Maybe you can expand comment a little.  But no need to reference the issue here.

I think we should prefer this pr over #1133.

mpmath/libmp/libmpf.py
```diff
@@ -1085,10 +1085,15 @@ def to_digits_exp(s, dps, base=10):
...
     return sign, digits, exponent
 
-def round_digits(sign, digits, dps, base, rnd=round_down, fixed=False):
+def round_digits(sign, digits, dps, base, rnd=round_down, fixed=False,
+                 sticky=False):
     """
     Returns the rounded digits, and the number of places the decimal point was
     shifted.
+
+    ``sticky`` marks that the value has a nonzero remainder past the last
```
skirpichev: Could you please rename this flag to exact (or inexact, whatever)?

mpmath/libmp/libmpf.py
```diff
@@ -1483,17 +1489,27 @@ def format_fixed(s, dps, rnd=round_down):
     # exponent by +- 1)
     _, digits, exponent = to_digits_exp(
             s, max(dps+exponent+4, int(s[3]/blog2_10)), base)
+    # to_digits_exp truncates; flag a nonzero remainder past the last digit so
+    # rounding is not fooled by a short zero tail (#1131).
```
skirpichev, line 1493: Lets expand this comment as I suggested above.

mpmath/libmp/libmpf.py
skirpichev: I would prefer if you adapt #1133 approach here.  Or adapt round_digits() to handle negative dps.

mpmath/tests/test_format.py
```diff
@@ -912,3 +912,51 @@ def test_hexadecimal_fmt():
         assert f'{x:.0a}' == '0x1p+0'
         assert f'{x:#.0a}' == '0x1.p+0'
         assert f"{mp.mpf('1.234567890123456789'):+.0a}" == '+0x1p+0'
+
+
+@given(st.floats(allow_nan=False, allow_infinity=False,
+                 allow_subnormal=False),
+       st.integers(min_value=1, max_value=40))
+def test_fixed_with_gmpy2_bulk(x, p):
```
skirpichev: I would prefer if you take similar test from the #1133.  Maybe also try different context's precisions.

skirpichev (changes requested):
I closed #1133 in favor of this pr.

But could you, please, address few remaining comments and incorporate some changes from #1133?

skirpichev:
BTW, alternative API change for round_digits is to pass additional `exponent` argument *and* return also adjusted exponent.  Your "sticky" flag be an internal detail of this function (I still suggest renaming to something like "is_inexact").  Signature would be:
```py
def round_digits(sign, digits, exponent, dps, base, rnd=round_down, fixed=False):
```

round_digits() - private API, we can break it.

skirpichev:
@gaoflow, I did proposed changes.  Let me know if you are ok with that and/or wish to add something else.

binggao1230:
Looks good to me — the unified zero-prefix path through round_digits is cleaner than my special case, and inexact is the better name. Re-ran the MPFR differential across rounding modes and context precisions (including the below-last-place cases): no mismatches. Nothing else from my side.

skirpichev (approved):
LGTM

> Looks good to me — the unified zero-prefix path through round_digits is cleaner than my special case, and inexact is the better name.

Thanks.  I think it's also good to refactor round_digits() and/or to_digits_exp() API as suggested above.  But lets keep this for a separate PR.

skirpichev:
Thanks.

PR #1138, opened 2026-07-19
Use CoW pattern to manage constant_memo() cache

### LINKED ISSUE #1135: constant_memo can return corrupted constants on free-threaded Python
Using the free-threaded build a multithreaded program using mpmath can fail because of the non-threadsafe use of caches.

This is the reproducer:
```python
import math
import threading
from mpmath import MPContext

barrier = threading.Barrier(4)
bad = []


def worker(index):
    mp = MPContext()

    for iteration in range(100):
        mp.prec = 100 + 100 * iteration + 10 * index
        barrier.wait()

        value = float(+mp.pi)
        if value != math.pi:
            bad.append((mp.prec, value))


threads = [
    threading.Thread(target=worker, args=(i,))
    for i in range(4)
]

for thread in threads:
    thread.start()

for thread in threads:
    thread.join()

assert not bad, bad[0]
```
Running that under CPython 3.14t:
```console
$ python t.py 
  File "/stuff/current/active/mpmath/t.py", line 32, in <module>
    assert not bad, bad[0]
           ^^^^^^^
AssertionError: (820, 0.0015339807878856412)
$ python t.py 
  File "/stuff/current/active/mpmath/t.py", line 32, in <module>
    assert not bad, bad[0]
           ^^^^^^^
AssertionError: (2210, 6588397.316661142)
```
I think this is coming from here:
https://github.com/mpmath/mpmath/blob/ead1a61d094531a6349ffb0b983e76919acaf643/mpmath/libmp/libelefun.py#L80-L99
I think the problem is a race condition between assigning `memo_val` and `memo_prec` when multiple threads overlap.

skirpichev:
Closes #1135

### DIFF
--- a/mpmath/libmp/libelefun.py
+++ b/mpmath/libmp/libelefun.py
@@ -77,6 +77,7 @@ for k in range(1, LOG_TAYLOR_PREC.bit_length()+1):
 #                                                                            #
 #----------------------------------------------------------------------------#
 
+
 def constant_memo(f):
     """
     Decorator for caching computed values of mathematical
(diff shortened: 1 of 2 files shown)


oscarbenjamin:
Looks good to me.

PR #1139, opened 2026-07-19
Widen to_str's digit window so str/nstr rounds near-boundary values correctly

gaoflow:
#1137's `round_digits()` refactor made `format_scientific` (the "e" type) and
`to_str` compute the sticky-remainder flag internally, fixing the hidden-zero
directed-rounding case for both. One related case remains in `to_str`.

`to_str` extracts `dps+10` digits, narrower than `format_scientific` /
`format_fixed`, which cover the whole mantissa (`int(bc/blog2_10)+10`). With only
`dps+10`, a value sitting just above a decimal boundary is extracted as `...99999`
one ULP low; rounding those nines up under a directed mode still lands one ULP
short of the exact value — and of the "e" format for the same value:

```pycon
>>> from mpmath import mp
>>> mp.prec = 200
>>> b = mp.mpf(1058187881481430099485) / mp.mpf(2)**74   # 0.056020000...16941...
>>> mp.nstr(b, 6, rnd='c')      # ceiling
'0.05602'                       # exact ceiling is 0.0560201
>>> format(b, '.5Ue')          # 'e' path, already correct
'5.60201e-02'
```

Fix: widen `to_str`'s base-10 window to the full mantissa, matching its siblings.
Non-decimal bases keep their existing window, and the "g"/general path routes
through `to_str`/`format_scientific`.

The real-world impact is small: it needs a specifically structured value and only
the directed-away modes, so ordinary output is unchanged. It's a consistency fix
so that str/nstr rounds the last digit like the "e" format does.
`test_str_rounding_near_boundary` covers positive and negative near-boundary
values with hard-coded expected strings and no gmpy2 dependency; the full suite
passes with and without gmpy2.

### DIFF
--- a/mpmath/libmp/libmpf.py
+++ b/mpmath/libmp/libmpf.py
@@ -1234,8 +1234,7 @@ def to_str(s, dps, strip_zeros=True, min_fixed=None, max_fixed=None,
     # For base 10 widen the window to the full mantissa (as format_scientific
     # does), otherwise a value just above a decimal boundary is extracted as
     # "...99999" one ULP low and directed rounding lands one ULP short.
-    _, _, _, bc = s
-    ndig = max(dps + 10, int(bc/blog2_10) + 10) if base == 10 else dps + 10
+    ndig = (max(dps, int(s[3]/blog2_10)) if base == 10 else dps) + 10
     sign, digits, exponent = to_digits_exp(s, ndig, base)
 
     rnd_digs = stddigits[(base//2 + base%2):base]
--- a/mpmath/libmp/libmpf.py
+++ b/mpmath/libmp/libmpf.py
@@ -1231,7 +1231,12 @@ def to_str(s, dps, strip_zeros=True, min_fixed=None, max_fixed=None,
 
     # to_digits_exp rounds to floor.
     # This sometimes kills some instances of "...00001"
-    sign, digits, exponent = to_digits_exp(s, dps+10, base)
+    # For base 10 widen the window to the full mantissa (as format_scientific
+    # does), otherwise a value just above a decimal boundary is extracted as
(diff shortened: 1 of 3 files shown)


binggao1230:
Rebased onto master now that #1137 has landed: that refactor already fixes the "e" path and computes the sticky-remainder flag internally, so I've reduced this to the one case it leaves — `to_str`'s `dps+10` window is narrower than the other formatters, so a value just above a decimal boundary is extracted as `...99999` one ULP low. Widening the base-10 window to the full mantissa makes str/nstr round the last digit like the "e" format.

mpmath/libmp/libmpf.py
```diff
@@ -1231,7 +1231,12 @@ def to_str(s, dps, strip_zeros=True, min_fixed=None, max_fixed=None,
 
     # to_digits_exp rounds to floor.
     # This sometimes kills some instances of "...00001"
-    sign, digits, exponent = to_digits_exp(s, dps+10, base)
+    # For base 10 widen the window to the full mantissa (as format_scientific
+    # does), otherwise a value just above a decimal boundary is extracted as
+    # "...99999" one ULP low and directed rounding lands one ULP short.
+    _, _, _, bc = s
+    ndig = max(dps + 10, int(bc/blog2_10) + 10) if base == 10 else dps + 10
```
skirpichev:
```suggestion
    ndig = (max(dps, int(s[3]/blog2_10)) if base == 10 else dps) + 10
```

skirpichev (approved):
> It's a consistency fix so that str/nstr rounds the last digit like the "e" format does.

Maybe then you can compare nstr() output (enforced to be in a scientific form) with "e" formatting?

PR #1141, opened 2026-07-21
Simplify and standardise weierhalfperiods

HeskethGD:
## Summary

Simplify `weierhalfperiods` by removing the root sorting and period-matching step.

The theta-function formulas in `_roots_from_omega` naturally return roots labelled by their corresponding half-periods:

- `e1 = wp(omega1)`
- `e2 = wp(omega1 + omega2)`
- `e3 = wp(omega2)`

Previously, these roots were sorted by their real and imaginary parts.`weierhalfperiods` then evaluated `weierp` at three candidate half-periods and searched all permutations to recover the association. This discarded the natural labelling and added several expensive elliptic-function evaluations.

## Changes

- Return the naturally labelled roots directly from `_roots_from_omega`.
- Remove root sorting, three additional `weierp` evaluations, and permutation matching from `weierhalfperiods`.
- Reduce the returned period ratio to the modular fundamental domain.
- Add deterministic conventions for boundary and sign ambiguities:
  - prefer `Re(tau) = 1/2` to `-1/2`;
  - prefer the right half of the boundary `abs(tau) = 1`;
  - place `omega1` in the right half-plane, including the negative imaginary axis as its boundary.
- Document these conventions in the `weierhalfperiods` docstring.
- Preserve the natural root ordering in `weierpinv`. Carlson's `R_F` is symmetric in its three arguments, so inverse evaluation does not depend on root ordering.
- Add reference values for representative real and complex invariants.
- Add tests confirming that boundary choices are stable across working precisions.
- Fix inline-code rendering of parameter names in the `weierpinv` docstring.

Extensive comparisons of half-periods against Wolfram Engine showed direct agreement for ordinary real and complex invariants, while some exact modular-boundary and special-lattice cases may return a different but equivalent period basis because Wolfram applies undocumented phase-dependent branch choices; reproducing those cases exactly could be considered later by implementing a dedicated Wolfram-compatible analytic branch convention.

## Performance

Removing the period-matching evaluations substantially reduces the cost of `weierhalfperiods`.

Representative local timings showed approximately:

| Case | Speedup |
| --- | ---: |
| Generic real invariants | 5–6x |
| Generic complex invariants | 5–6x |
| Lemniscatic case | 5–12x |
| Equianharmonic case | 40–90x |

Exact timings depend on working precision.

## Behavioural change

`weierhalfperiods` may now return a different, modularly equivalent basis for the same lattice.

For example, for `(g2, g3) = (60, 140)`, the period ratio changes from

    -0.5 + 0.209032224450873j

to the fundamental-domain representative

    0.5 + 1.19598784664302j

The two ratios are related by an integer unimodular transformation and produce the same invariants and Weierstrass functions.

## Tests

The elliptic tests and elliptic-function doctests pass:

    72 passed
    
    
AI use declaration: GPT5.6 Sol from OpenAI was used to assist. All lines of code were read and iterated on by me.

### DIFF
--- a/mpmath/functions/elliptic.py
+++ b/mpmath/functions/elliptic.py
@@ -1635,6 +1635,9 @@ def _roots_from_omega(ctx, omega1, omega2):
     """
     Compute roots e1, e2, e3 of 4*z^3 - g2*z - g3 = 0 using theta functions.
     This is ~10x faster than solving the cubic directly.
+
+    The roots are naturally labelled by e1 = wp(omega1),
+    e2 = wp(omega1 + omega2), and e3 = wp(omega2).
     """
     tau = omega2 / omega1
     q = ctx.qfrom(tau=tau)
@@ -1644,8 +1647,33 @@ def _roots_from_omega(ctx, omega1, omega2):
     e1 = c * (j24 + 2*j44)
     e2 = c * (j24 - j44)
     e3 = -c * (2*j24 + j44)
-    roots = sorted([(e.real, e.imag) for e in [e1, e2, e3]], reverse=True)
-    return [ctx.mpc(real=t[0], imag=t[1]) for t in roots]
+    return e1, e2, e3
+
+
+def _canonicalize_weierstrass_periods(ctx, omega1, omega2):
+    """Choose a deterministic orientation for a reduced period basis."""
+    tau = omega2 / omega1
+
+    # Identify the vertical edges of the fundamental domain.
+    if ctx.almosteq(ctx.re(tau), -ctx.one/2):
+        omega2 += omega1
+        tau = omega2 / omega1
+
+    # Identify equivalent points on the circular boundary.
+    if ctx.almosteq(abs(tau), ctx.one) and ctx.re(tau) < 0:
+        omega1, omega2 = omega2, -omega1
+
+    # The invariants do not distinguish a basis from its simultaneous
+    # negation. Follow the usual convention of placing omega1 in the right
+    # half-plane, using the negative imaginary axis as the boundary.
+    real = ctx.re(omega1)
+    if ((real < 0 and not ctx.almosteq(real, 0)) or
+            (ctx.almosteq(real, 0) and ctx.im(omega1) > 0)):
+        omega1 = -omega1
+        omega2 = -omega2
+
+    return omega1, omega2
+
 
 def _eisenstein_E4_E6(ctx, tau):
     """
@@ -1737,7 +1765,14 @@ def weierinvariants(ctx, omega1, omega2):
 def weierhalfperiods(ctx, g2, g3):
     r"""
     Returns a pair of fundamental half-periods `(\omega_1, \omega_2)`
-    corresponding to the Weierstrass invariants `(g_2, g_3)`::
+    corresponding to the Weierstrass invariants `(g_2, g_3)`.
+
+    The basis is chosen so that `\tau = \omega_2/\omega_1` lies in the
+    standard modular fundamental domain. On its boundary, the representative
+    with `\operatorname{Re}(\tau) = 1/2` is preferred to `-1/2`, and the
+    right half of the circle `|\tau| = 1` is preferred. The simultaneous sign
+    ambiguity is fixed by placing `\omega_1` in the right half-plane, with
+    the negative imaginary axis included as its boundary::
 
         >>> from mpmath import mp, chop
         >>> from mpmath import weierhalfperiods, weierinvariants
@@ -1747,7 +1782,7 @@ def weierhalfperiods(ctx, g2, g3):
         >>> chop(g2), chop(g3)
         (60.0, 140.0)
         >>> chop(omega2/omega1)
-        (-0.5 + 0.209032224450873j)
+        (0.5 + 1.19598784664302j)
 
     """
     with ctx.extraprec(10):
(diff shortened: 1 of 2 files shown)


skirpichev (commented):
> Extensive comparisons of half-periods against Wolfram Engine showed direct agreement for ordinary real and complex invariants, while some exact modular-boundary and special-lattice cases may return a different but equivalent period basis because Wolfram applies undocumented phase-dependent branch choices; reproducing those cases exactly could be considered later by implementing a dedicated Wolfram-compatible analytic branch convention.

Could you elaborate on this?

Examples?

How you propose to handle this?  Change how half-periods computed again?  Add some flag to enable compatible convention?

HeskethGD:
> > Extensive comparisons of half-periods against Wolfram Engine showed direct agreement for ordinary real and complex invariants, while some exact modular-boundary and special-lattice cases may return a different but equivalent period basis because Wolfram applies undocumented phase-dependent branch choices; reproducing those cases exactly could be considered later by implementing a dedicated Wolfram-compatible analytic branch convention.
> 
> Could you elaborate on this?
> 
> Examples?
> 
> How you propose to handle this? Change how half-periods computed again? Add some flag to enable compatible convention?

The primary aim of this PR was to remove the redundant root sorting and associated calculation of wp values as it was not serving any clear purpose, adding to computation time, and would likely be confusing to anyone subsequently trying to maintain it. That has been done. The removed code sorted approximate roots lexicographically by real and imaginary parts. That ordering:

- is not an analytic branch convention;
- can change abruptly when roots cross or nearly coincide;
- is arbitrary for general complex invariants;
- was not shown to reproduce Wolfram’s choices.

A subsequent aim was to also try to apply any simple rules to the selection of the half-period lattice to align with Wolfram Engine; although not crucial I recall questions as to why some `tau` values differed in e.g. Flink and so I thought if we could adopt what Wolfram does that would be good and also assist with testing. This has been done in most but not all cases. 

Wolfram does not publish any rules it uses to select the half-period basis so I had to probe it to see any conventions. I compared mpmath with Wolfram Engine using fixed real and complex inputs, scaled lattices, exact modular-boundary cases, and 100 randomly generated nonsingular complex pairs. After applying the current sign and fundamental-domain conventions, the ordinary nonboundary cases agreed directly. The remaining differences were confined to special lattices or ratios lying exactly on a boundary of the modular fundamental domain.

For example, with \((g_2,g_3)=(1,0)\), mpmath currently returns approximately

```text
{-1.8540746773013719184 i, 1.8540746773013719184}
```
while Wolfram returns

```text
{1.8540746773013719184, 1.8540746773013719184 i}
 ```

These are related by

$(\omega_1^{W},\omega_2^{W})=(\omega_2,-\omega_1)$,

an \(S\)-type unimodular basis transformation.

Another example is \((g_2,g_3)=(0,-1)\). Mpmath returns approximately

```text
{1.324979062714087468 - 0.764977018528596437 i, 1.324979062714087468 + 0.764977018528596437 i}
```

whereas Wolfram returns

```text
{1.324979062714087468 - 0.764977018528596437 i, 1.529954037057192875 i}
```
Here Wolfram’s second period is

$\omega_2^{W}=\omega_2-\omega_1$,

a \(T\)-type basis transformation. The respective ratios are the equivalent boundary points

$\tfrac12+\tfrac{\sqrt3}{2}i\quad\text{and}\quad-\tfrac12+\tfrac{\sqrt3}{2}i$.

Mpmath currently uses an explicit geometric convention: reduce \(\tau\) to the modular fundamental domain, prefer the \(+\tfrac12\) vertical edge and the right unit-circle arc, and fix the simultaneous sign by orienting \(\omega_1\) into a chosen half-plane. Wolfram appears to define named analytic branches \(\omega_1(g_2,g_3)\) and \(\omega_3(g_2,g_3)\). On exact boundaries, its choice also depends on the phase introduced by fractional roots, so it cannot be reproduced by a single rule based only on the location of \(\tau\).

Forcing exact Wolfram agreement would therefore require more than changing the final sign. We would probably need either:

1. to reproduce Wolfram’s analytic construction using a particular inverse Klein-invariant branch, ordered cubic roots, and principal square/fourth roots; or
2. to reverse-engineer when to apply the finite \(S\) and \(T\) transformations on boundary cases.

The transformations themselves are computationally negligible, but reliably selecting Wolfram’s undocumented branch for arbitrary complex invariants would add implementation and maintenance complexity. I would not add a public compatibility flag at this stage: both outputs are mathematically equivalent, and a “Wolfram-compatible” option would imply stronger compatibility than we can presently guarantee. If exact agreement later becomes a requirement, it would be better to implement and document one complete analytic branch convention rather than expose a partial flag.


### Where disagreement is expected

The known disagreements are concentrated on exact boundaries of the modular fundamental domain:

1. The vertical edges:

$\mathcal{Re}\tau=\pm\frac12$.

The two sides are related by \(T\).

2. The circular edge:

$|\tau|=1$.

Equivalent points on this edge can be related by \(S\).

3. High-symmetry special lattices, which lie on those boundaries:

- lemniscatic cases such as \(g_3=0\), with \(\tau=i\);
- equianharmonic cases such as \(g_2=0\), with \(\tau=e^{i\pi/3}\) or an equivalent value.

For example, at `(g2, g3) = (1, 0)`, mpmath and Wolfram choose bases related by \(S\). At `(0, -1)`, they choose bases related by \(T^{-1}\).


So, to summarise, I do not anticipate any further changes to the computation of half-periods. I am content to accept the remaining differences. If we were to require complete agreement with Wolfram it would require further research to figure out what undocumented branch rules are being applied in a couple of places. I do not see the value in that currently but were we to do so I would not use a flag.

skirpichev (approved):

skirpichev:
Thanks.

Issue #1147, opened 2026-07-28
incomplete list of exceptions to Rosser's Rule


catalin-hanga:
The list of exceptions to Rosser's rule for n <= 400,000,000 is incomplete in its current form.
I found 2 additional exceptions, that are not included in https://github.com/mpmath/mpmath/blob/master/mpmath/functions/zetazeros.py#L630

The first exception occurs inside Gram block $[g_{325,890,638}, g_{325,890,641}]$.
The second exception occurs in Gram block $[g_{357,738,762}, g_{357,738,765}]$.

In both cases, the first 2 Gram intervals do not contain any zeta zeros, and the third interval contains 3 zeros, so the pattern is (00)3.
This can be verified by direct computation, or by plotting the Siegel Z function in these ranges.

ariasdereyna:
Good catch. I confirmed your values. So, we have to insert two items to the list of exceptions

[325890638, 325890641], '(00)3', 

 between actual lines  895--896. 

[357738762, 357738765], '(00)3', 

between actual lines  944--945.

When I wrote the program I included this list of exceptions to recognise the work of  R. P. Brent, J. van de Lune, H.H. J. te Riele and D. T. Winter that compose it.  I think it is good to make the list complete. 

I am not aware of how to manage this.  Who have to approve and how to insert this in the file.  Thanks in advance to anyone who can help.  

Juan Arias de Reyna

skirpichev:
> So, we have to insert two items to the list of exceptions

Probably, we need also a test, as for any bug.

> I am not aware of how to manage this. Who have to approve and how to insert this in the file.

Perhaps you or someone else could submit a pr:
https://github.com/mpmath/mpmath/compare

ariasdereyna:
Answering to skirpichev

I feel sorry, but what you said look as chinese to me  (" submit a pr").  I clicked the link  and was  even more intimidated.  perhaps you can explain it in a more easy language. In fact I am eager to learn.   Not many days ago I tried to find your mail address  to speak about another program by me in mpmath  (secondzeta).  I am working on a better program.

skirpichev:
> I clicked the link and was even more intimidated. perhaps you can explain it in a more easy language.

Github has extensive documentation, e.g. for creating pull requests:
https://docs.github.com/en/pull-requests/how-tos/create-pull-requests/creating-a-pull-request

Perhaps, CPython developers guide can also be used as how-to style docs for working with Git and Github:
[...]

PR #1148, opened 2026-07-29
added 2 new expceptions to Rosser's Rule

catalin-hanga:
Trying to fix the problem described in this issue: https://github.com/mpmath/mpmath/issues/1147


skirpichev:
Could you add tests?

catalin-hanga:
> Could you add tests?

In order to test that the 2 exceptions are indeed valid, the easiest way would be to just plot the Z function in these 2 blocks, and observe that, in both cases, the first 2 Gram intervals don't contain any zeros, and the third interval contains 3 zeros. I have attached the 2 plots here. If you want, I could also provide the exact numerical values for these Gram points and zeta zeros.

ex1
ex2

ariasdereyna:
I think a good test is to compute the zero number 325890640. With the old program, possibly did not output anything, with the new program it will run nicely. Perhaps adding zetazero (325890640 , info = True ), that possibly return '(00)3'

One of the things I would like to do, and do not know how, it is to check this in a copy of mpmath.

catalin-hanga:
I've opened a separate PR at https://github.com/mpmath/mpmath/pull/1149
Here, I've added the extra tests to page https://github.com/mpmath/mpmath/blob/master/mpmath/tests/test_extra_zeta.py

The Gram interval [g_325890640, g_325890641] should contain the following zeros:
zetazero(325890640) = 129273228.66142665
zetazero(325890641) = 129273228.76005181
zetazero(325890642) = 129273228.79754069

The Gram interval [g_357738764, g_357738765] should contain the following zeros:
zetazero(357738764) = 141125096.01260684
zetazero(357738765) = 141125096.18511831
zetazero(357738766) = 141125096.28064566

These values can be double-checked at https://www.lmfdb.org/zeros/zeta/.

The main point is that the old version of the code gives the wrong answer for zetazero(357738766), and runs in an infinite loop for zetazero(357738764), zetazero(357738765), zetazero(325890640), zetazero(325890641).

skirpichev:
> the old version of the code gives the wrong answer for zetazero(357738766), and runs in an infinite loop for zetazero(357738764), zetazero(357738765), zetazero(325890640), zetazero(325890641)

Yes, tests are good.  But there is no need to open a new pr to address review comments: please just add new commits here.

skirpichev (approved):

skirpichev:
> One of the things I would like to do, and do not know how, it is to check this in a copy of mpmath.

Assuming you have installed git (https://git-scm.com/install/), you need to clone OP's repository and select appropriate branch:
```
git clone https://github.com/catalin-hanga/mpmath.git
cd mpmath
git checkout patch-1
```
Then install mpmath from sources (it would be better if you first create an isolated Python environment first: https://docs.python.org/3/library/venv.html#module-venv):
```
pip install -e .
```

Or you can directly install mpmath package from the git tree, here full steps on my system (commands for venv might vary for Windows):
```
python -m venv tmp-venv  # create virtual environment
. tmp-venv/bin/activate  # activate it
pip install git+https://github.com/catalin-hanga/mpmath.git@patch-1
```

Then you can run python interpreter and test:
```pycon
Python 3.14.5 (tags/v3.14.5:5607950ef23, May 14 2026, 13:43:31) [GCC 14.2.0] on linux
>>> from mpmath import zetazero
>>> zetazero(325890640)
mpc(real='0.5', imag='129273228.66142665')
```

skirpichev:
I'm going to merge this.

@ariasdereyna, above instructions should work after merge too, until @catalin-hanga delete the pr's branch.  But you can also install the latest development version of the mpmath, which will include this change.  For example, with pip:
```
pip install git+https://github.com/mpmath/mpmath.git
```

ariasdereyna:
@skirpichev  Thanks you. I will try this.

PR #1150, opened 2026-07-31
Avoid spurious overflow in fp gammaprod

### LINKED ISSUE #493: Unexpected overflow in fp version of binomial coefficient function
mpmath version: 1.1.0

For a moderately large first argument and a second argument of 1, the fp version of the binomial function throws an overflow error despite the fact that the return should just be the value of the first argument. It appears that the overflow is encountered in a call to the gammaprod function which could be circumnavigated if a check for the special case of `n choose 1 = n` was included. However, I do understand the design principle of the more general method... just figured I would bring this to attention.

Example:
```
Python 3.6.8 (default, Aug 20 2019, 17:12:48) 
[GCC 8.3.0] on linux
>>> from mpmath import *
>>> mp.binomial(1100,1)
mpf('1100.0')
>>> fp.binomial(1100,1)
  File "/usr/local/lib/python3.6/dist-packages/mpmath/functions/factorials.py", line 66, in binomial
    return ctx.gammaprod([n1], [k1, nk1])
  File "/usr/local/lib/python3.6/dist-packages/mpmath/functions/factorials.py", line 35, in gammaprod
    for x in regular_num: p *= ctx.gamma(x)
  File "/usr/local/lib/python3.6/dist-packages/mpmath/math2.py", line 30, in f
    return f_real(x)
  File "/usr/local/lib/python3.6/dist-packages/mpmath/math2.py", line 215, in _gamma_real
    return 2.506628274631000502417 * t**(x+0.5) * math.exp(-t) * r
OverflowError: (34, 'Numerical result out of range')
```

Sanjays2402:
Closes #493

`fp.binomial(1100, 1)` raised `OverflowError` although the result is exactly 1100: `gammaprod` multiplies the individual gamma values, and `gamma(1101)` exceeds the double range even though the quotient does not. When a term overflows, the regular numerator/denominator is now evaluated in log space, so the quotient is returned whenever it is representable; genuinely out-of-range results such as `fp.binomial(1100, 550)` still raise `OverflowError`.

`test_issue_493` in `mpmath/tests/test_gammazeta.py` fails without the change and passes with it; `test_gammazeta.py`, `test_fp.py`, `test_functions.py` and `test_functions2.py` pass (180 passed, 1 skipped).

This change was prepared with AI assistance; the regression test was run locally and fails without the fix.

### DIFF
--- a/mpmath/tests/test_gammazeta.py
+++ b/mpmath/tests/test_gammazeta.py
@@ -705,3 +705,11 @@ def test_issue_471():
 def test_issue_472():
     assert bernpoly(4, mpc(inf, 1e-50)) == mpc(inf, 0)
     assert mpc(inf, 2)**4 == mpc(inf, 0)
+
+def test_issue_493():
+    assert fp.binomial(1100, 1) == pytest.approx(1100.0)
+    assert fp.binomial(1100, 1099) == pytest.approx(1100.0)
+    assert fp.binomial(1100, 0) == 1.0
+    assert fp.rf(1100, 1) == pytest.approx(1100.0)
+    assert fp.beta(1100, 1) == pytest.approx(1/1100)
+    assert fp.binomial(5, 2) == pytest.approx(10.0)
--- a/mpmath/tests/test_gammazeta.py
+++ b/mpmath/tests/test_gammazeta.py
@@ -705,11 +705,3 @@ def test_issue_471():
 def test_issue_472():
     assert bernpoly(4, mpc(inf, 1e-50)) == mpc(inf, 0)
     assert mpc(inf, 2)**4 == mpc(inf, 0)
-
-def test_issue_493():
-    assert fp.binomial(1100, 1) == pytest.approx(1100.0)
-    assert fp.binomial(1100, 1099) == pytest.approx(1100.0)
-    assert fp.binomial(1100, 0) == 1.0
-    assert fp.rf(1100, 1) == pytest.approx(1100.0)
-    assert fp.beta(1100, 1) == pytest.approx(1/1100)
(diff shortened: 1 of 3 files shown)


mpmath/tests/test_gammazeta.py
```diff
@@ -705,3 +705,11 @@ def test_issue_471():
...
     assert bernpoly(4, mpc(inf, 1e-50)) == mpc(inf, 0)
     assert mpc(inf, 2)**4 == mpc(inf, 0)
+
+def test_issue_493():
+    assert fp.binomial(1100, 1) == pytest.approx(1100.0)
+    assert fp.binomial(1100, 1099) == pytest.approx(1100.0)
+    assert fp.binomial(1100, 0) == 1.0
+    assert fp.rf(1100, 1) == pytest.approx(1100.0)
+    assert fp.beta(1100, 1) == pytest.approx(1/1100)
+    assert fp.binomial(5, 2) == pytest.approx(10.0)
```
skirpichev: Either move this to test_fp.py and use ae() helper to test approximate equality, or use fp.almosteq().

Sanjays2402: Moved it to test_fp.py as test_issue_493 and switched to the ae() helper instead of pytest.approx; the exact cases (binomial(1100, 0) == 1.0) stay exact. Also added an explicit pytest.raises(OverflowError) for fp.binomial(1100, 550) so the still-out-of-range case is pinned down too.

712fc8b. test_fp.py and test_gammazeta.py together are 59 passed; reverting the factorials.py change makes test_issue_493 fail with OverflowError: math range error.

skirpichev (approved):

Issue #707, opened 2023-06-08
FFT and other numerical algorithms in mpmath ?


gabrielfougeron:
Hi,

Sorry if this is not the correct place to discuss this kind of topic. Please do point me to a better suited place if so.

I'm very fond of mpmath and of the few numerical algorithms it provides (eigendecomposition in particular).
I see some people have tried to implement other such numerical algorithm using mpmath (https://github.com/dbstein/apfft for FFT in particularà

My question is: are there plans to consolidate / build this kind of work in the future ?

gabrielfougeron:
I'm currently reading about Arb and FLINT available in python via python-flint, which probably makes most of this topic moot ... Sorry about that...

skirpichev:
> I'm currently reading about Arb and FLINT available in python via python-flint

Yeah, I think it should be available in Arb.  Not sure if the FFT support is available in the python-flint so far.

> My question is: are there plans to consolidate / build this kind of work in the future?

There are no pr's to do this so far.  But I'm +1 for this feature in the mpmath.  Lets see if Fredrik will comment.

fredrik-johansson:
I have no objection to adding FFTs.

baranwalayush:
Hi everyone,
I'd like to work on implementing the FFT algorithm and would appreciate some initial guidance or any suggestions before proceeding:

Here is what I'm currently thinking:

- I plan to implement the radix-2 Cooley-Tukey algorithm.
- I believe it should be added in `functions/signal.py`, or is there any other preferred location?
- and we will be accounting only for power-of-two input sizes, right?

baranwalayush:
Hey @fredrik-johansson @skirpichev , pinging in case you missed the comment above. Let me know your thoughts.

skirpichev:
Sorry, I've no strong opinion on this.  Few remarks:

> I believe it should be added in functions/signal.py

IMO, this rather belongs to mpmath/calculus.

> and we will be accounting only for power-of-two input sizes, right?

Why?

BTW, please also look into referenced SymPy issue.

baranwalayush:
> IMO, this rather belongs to mpmath/calculus.

Ok

> > and we will be accounting only for power-of-two input sizes, right?
> 
> Why?

[...]

PR #1152, opened 2026-08-02
Add fft()/invfft() functions for DFT calculation

baranwalayush:
Addresses #707 

Implements the `Radix-2 Cooley-Tukey` Fast Fourier Transform (FFT) algorithm to compute the discrete fourier transform (DFT) and inverse discrete fourier transform (IDFT) of a signal.

### Key Changes:
- Added iterative Radix-2 Cooley-Tukey FFT for power-of-two input lengths signals.
- Implemented `fft(x)` and `invfft(X)` for calculating forward and inverse DFT.
- Added unit tests along with docstrings and usage examples.

### Design Decisions & Next Steps
- No zero-padding: After a long discussion, we decided not to use optional zero-padding and Inputs are currently restricted to lengths of powers of 2 ($N = 2^k$).
- Future Work: Support for arbitrary input sizes via Bluestein's algorithm or similar approaches.

### DIFF
--- a/mpmath/calculus/fft.py
+++ b/mpmath/calculus/fft.py
@@ -13,18 +13,14 @@ def _fft_cooley_tuckey(ctx, values, inverse=False):
         return values
 
     # Bit-Reversal Permutation
-    def reverse_bits(n, width):
-        result = 0
-        for _ in range(width):
-            result <<= 1
-            result |= (n & 1)
-            n >>= 1
-        return result
-
     transformed = [ctx.zero] * n
     num_bits = n.bit_length() - 1
     for i in range(n):
-        rev = reverse_bits(i, num_bits)
+        rev = 0
+        for _ in range(num_bits):
+            rev <<= 1
+            rev |= (i & 1)
+            i >>= 1           
         transformed[rev] = values[i]
 
     sign = ctx.one if inverse else -ctx.one
--- a/mpmath/calculus/fft.py
+++ b/mpmath/calculus/fft.py
@@ -13,13 +13,18 @@ def _fft_cooley_tuckey(ctx, values, inverse=False):
         return values
 
     # Bit-Reversal Permutation
+    def reverse_bits(n, width):
+        result = 0
+        for _ in range(width):
+            result <<= 1
+            result |= (n & 1)
+            n >>= 1
+        return result
+
     transformed = [ctx.zero] * n
     num_bits = n.bit_length() - 1
     for i in range(n):
-        rev = 0
-        for b in range(num_bits):
-            if (i >> b) & 1:
-                rev |= 1 << (num_bits - 1 - b)
+        rev = reverse_bits(i, num_bits)
         transformed[rev] = values[i]
 
     sign = ctx.one if inverse else -ctx.one
--- a/mpmath/calculus/fft.py
+++ b/mpmath/calculus/fft.py
@@ -17,11 +17,12 @@ def _fft_cooley_tuckey(ctx, values, inverse=False):
     num_bits = n.bit_length() - 1
     for i in range(n):
         rev = 0
+        val = values[i]
         for _ in range(num_bits):
             rev <<= 1
             rev |= (i & 1)
             i >>= 1           
-        transformed[rev] = values[i]
+        transformed[rev] = val
 
     sign = ctx.one if inverse else -ctx.one
 
--- a/mpmath/calculus/fft.py
+++ b/mpmath/calculus/fft.py
@@ -50,6 +50,16 @@ def fft(ctx, values):
     Computes the Discrete Fourier Transform (DFT) of a sequence.
 
     Raises NotImplementedError if the input sequence length is not a power of 2.
+
+    A few basic examples are:
+
+        >>> from mpmath import mp, fft
+        >>> mp.nprint(fft([1, 0, 0, 0]))
+        [1.0, (1.0 + 0.0j), 1.0, (1.0 + 0.0j)]
+        >>> mp.nprint(fft([1 + 2j, 1 + 2j]))
+        [(2.0 + 4.0j), (0.0 + 0.0j)]
+        >>> mp.nprint(fft([1, 2, 3, 4]))
+        [10.0, (-2.0 + 2.0j), -2.0, (-2.0 - 2.0j)]
     """
     n = len(values)
     if n == 0:
@@ -69,6 +79,15 @@ def invfft(ctx, values):
     Computes the inverse Discrete Fourier Transform (IDFT) of a sequence.
 
     Raises NotImplementedError if the input sequence length is not a power of 2.
+
+    A few basic examples are:
+
+        >>> from mpmath import mp, fft, invfft
+        >>> mp.nprint(invfft([1, 1, 1, 1]))
+        [1.0, (0.0 + 0.0j), 0.0, (0.0 + 0.0j)]
+        >>> x = [1, 2, 3, 4]
+        >>> mp.nprint(invfft(fft(x)))
+        [(1.0 + 0.0j), (2.0 + 0.0j), (3.0 + 0.0j), (4.0 + 0.0j)]
     """
     n = len(values)
     if n == 0:
--- a/mpmath/calculus/fft.py
+++ b/mpmath/calculus/fft.py
@@ -2,9 +2,10 @@ from .calculus import defun
 
 def _fft_cooley_tuckey(ctx, values, inverse=False):
     """
-    This function implements the Radix-2 Cooley-Tukey FFT algorithm iteratively. It computes the Fast Fourier Transform (or Inverse FFT) of a sequence of complex numbers.
-    The input sequence must have a length that is a power of 2.
-
+    This function implements the Radix-2 Cooley-Tukey FFT algorithm iteratively.
+    It computes the Fast Fourier Transform (or Inverse FFT) of a sequence of
+    complex numbers.
+    
     https://en.wikipedia.org/wiki/Cooley%E2%80%93Tukey_FFT_algorithm
     """
     n = len(values)
@@ -21,13 +22,12 @@ def _fft_cooley_tuckey(ctx, values, inverse=False):
                 rev |= 1 << (num_bits - 1 - b)
         transformed[rev] = values[i]
 
-    sign = 1 if inverse else -1
-    two_pi_j = 2 * ctx.pi * ctx.j
+    sign = ctx.one if inverse else -ctx.one
 
     length = 2
     while length <= n:
         half = length // 2
-        w_len = ctx.exp(sign * two_pi_j / length)
+        w_len = ctx.expjpi(2 * sign / length)
 
         for i in range(0, n, length):
             w = ctx.one
@@ -44,11 +44,11 @@ def _fft_cooley_tuckey(ctx, values, inverse=False):
 
     return transformed
 
-# TODO: Add support for non-power-of-two lengths using Bluestein's algorithm or similar methods.
-
 @defun
 def fft(ctx, values):
-    r"""Computes the Discrete Fourier Transform (DFT) of a sequence using the Radix-2 Cooley-Tukey FFT algorithm.
+    r"""
+    Computes the Discrete Fourier Transform (DFT) of a sequence.
+
     Raises NotImplementedError if the input sequence length is not a power of 2.
     """
     n = len(values)
@@ -56,35 +56,27 @@ def fft(ctx, values):
         return []
 
     is_power_of_two = (n & (n - 1)) == 0
-
     if not is_power_of_two:
         raise NotImplementedError(f"FFT is only implemented for lengths that are powers of 2, got length: {n}")
 
     converted_values = [ctx.convert(v) for v in values]
-
-    result = _fft_cooley_tuckey(ctx, converted_values, False)
-
-    return result
+    return _fft_cooley_tuckey(ctx, converted_values, False)
 
 @defun
 def invfft(ctx, values):
-    r"""Computes the inverse Discrete Fourier Transform (IDFT) of a sequence using the Radix-2 Cooley-Tukey FFT algorithm.
+    r"""
+    Computes the inverse Discrete Fourier Transform (IDFT) of a sequence.
+
     Raises NotImplementedError if the input sequence length is not a power of 2.
     """
-
     n = len(values)
     if n == 0:
         return []
 
     is_power_of_two = (n & (n - 1)) == 0
-
     if not is_power_of_two:
         raise NotImplementedError(f"Inverse FFT is only implemented for lengths that are powers of 2, got length: {n}")
 
     converted_values = [ctx.convert(v) for v in values]
-
     result = _fft_cooley_tuckey(ctx, converted_values, True)
-
-    result = [val / n for val in result]
-
-    return result
+    return [val / n for val in result]
--- a/mpmath/calculus/fft.py
+++ b/mpmath/calculus/fft.py
@@ -5,7 +5,7 @@ def _fft_cooley_tuckey(ctx, values, inverse=False):
     This function implements the Radix-2 Cooley-Tukey FFT algorithm iteratively.
     It computes the Fast Fourier Transform (or Inverse FFT) of a sequence of
     complex numbers.
-    
+
     https://en.wikipedia.org/wiki/Cooley%E2%80%93Tukey_FFT_algorithm
     """
     n = len(values)
--- a/mpmath/calculus/fft.py
+++ b/mpmath/calculus/fft.py
@@ -51,15 +51,16 @@ def fft(ctx, values):
 
     Raises NotImplementedError if the input sequence length is not a power of 2.
 
-    A few basic examples are:
+    **Examples**
 
-        >>> from mpmath import mp, fft
-        >>> mp.nprint(fft([1, 0, 0, 0]))
-        [1.0, (1.0 + 0.0j), 1.0, (1.0 + 0.0j)]
-        >>> mp.nprint(fft([1 + 2j, 1 + 2j]))
-        [(2.0 + 4.0j), (0.0 + 0.0j)]
-        >>> mp.nprint(fft([1, 2, 3, 4]))
-        [10.0, (-2.0 + 2.0j), -2.0, (-2.0 - 2.0j)]
+    >>> from mpmath import mp
+    >>> mp.pretty = True
+    >>> mp.fft([1, 0, 0, 0])
+    [1.0, (1.0 + 0.0j), 1.0, (1.0 + 0.0j)]
+    >>> mp.fft([1 + 2j, 1 + 2j])
+    [(2.0 + 4.0j), (0.0 + 0.0j)]
+    >>> mp.fft([1, 2, 3, 4])
+    [10.0, (-2.0 + 2.0j), -2.0, (-2.0 - 2.0j)]
     """
     n = len(values)
     if n == 0:
@@ -71,7 +72,9 @@ def fft(ctx, values):
                                   f"are powers of 2, got length: {n}")
 
     converted_values = [ctx.convert(v) for v in values]
-    return _fft_cooley_tuckey(ctx, converted_values, False)
+    with ctx.extraprec(10):
+        result = _fft_cooley_tuckey(ctx, converted_values)
+    return [+v for v in result]
 
 @defun
 def invfft(ctx, values):
@@ -80,14 +83,15 @@ def invfft(ctx, values):
 
     Raises NotImplementedError if the input sequence length is not a power of 2.
 
-    A few basic examples are:
+    **Examples**
 
-        >>> from mpmath import mp, fft, invfft
-        >>> mp.nprint(invfft([1, 1, 1, 1]))
-        [1.0, (0.0 + 0.0j), 0.0, (0.0 + 0.0j)]
-        >>> x = [1, 2, 3, 4]
-        >>> mp.nprint(invfft(fft(x)))
-        [(1.0 + 0.0j), (2.0 + 0.0j), (3.0 + 0.0j), (4.0 + 0.0j)]
+    >>> from mpmath import mp
+    >>> mp.pretty = True
+    >>> mp.invfft([1, 1, 1, 1])
+    [1.0, (0.0 + 0.0j), 0.0, (0.0 + 0.0j)]
+    >>> x = [1, 2, 3, 4]
+    >>> mp.invfft(mp.fft(x))
+    [(1.0 + 0.0j), (2.0 + 0.0j), (3.0 + 0.0j), (4.0 + 0.0j)]
     """
     n = len(values)
     if n == 0:
@@ -99,5 +103,6 @@ def invfft(ctx, values):
                                   f"are powers of 2, got length: {n}")
 
     converted_values = [ctx.convert(v) for v in values]
-    result = _fft_cooley_tuckey(ctx, converted_values, True)
+    with ctx.extraprec(10):
+        result = _fft_cooley_tuckey(ctx, converted_values, True)
     return [val / n for val in result]
--- a/mpmath/calculus/fft.py
+++ b/mpmath/calculus/fft.py
@@ -1,0 +1,47 @@
+from .calculus import defun
+
+def _fft_recursive(ctx, values, inverse=False):
+    n = len(values)
+
+    if n == 1:
+        return values
+
+    sign = 1 if inverse else -1
+
+    even = _fft_recursive(ctx, values[0::2], inverse)
+    odd = _fft_recursive(ctx, values[1::2], inverse)
+    root = ctx.exp(sign * 2 * ctx.pi * ctx.j / n)
+    factor = ctx.one
+    half = n // 2
+    transformed = [ctx.zero] * n
+    for k in range(half):
+        term = factor * odd[k]
+        transformed[k] = even[k] + term
+        transformed[k + half] = even[k] - term
+        factor *= root
+    return transformed
+
+@defun
+def fft(ctx, values, inverse=False):
+    r"""Computes the discrete Fourier transform of a sequence.
+
+    If ``inverse=True``, computes the inverse transform and normalizes the
+    result by the sequence length.
+    """
+    n = len(values)
+    if n == 0:
+        return []
+    if n & (n - 1) != 0:
+        raise ValueError("Length of input must be a power of 2.")
+
+    converted_values = [ctx.convert(v) for v in values]
+    transformed = _fft_recursive(ctx, converted_values, inverse=inverse)
+    if inverse and transformed:
+        n = len(transformed)
+        transformed = [value / n for value in transformed]
+    return transformed
+
+@defun
+def ifft(ctx, values):
+    r"""Computes the inverse discrete Fourier transform of a sequence."""
+    return fft(ctx, values, inverse=True)
--- a/mpmath/calculus/fft.py
+++ b/mpmath/calculus/fft.py
@@ -23,12 +23,48 @@ def _fft_recursive(ctx, values, inverse=False):
         factor *= root
     return transformed
 
+def _fft_iterative(ctx, values, inverse=False):
+    n = len(values)
+    if n <= 1:
+        return list(values)
+
+    # Bit-Reversal Permutation
+    transformed = [ctx.zero] * n
+    num_bits = n.bit_length() - 1
+    for i in range(n):
+        rev = 0
+        for b in range(num_bits):
+            if (i >> b) & 1:
+                rev |= 1 << (num_bits - 1 - b)
+        transformed[rev] = values[i]
+
+    sign = 1 if inverse else -1
+    two_pi_j = 2 * ctx.pi * ctx.j
+
+    length = 2
+    while length <= n:
+        half = length >> 1
+        w_len = ctx.exp(sign * two_pi_j / length)
+
+        for i in range(0, n, length):
+            w = ctx.one
+            for j in range(half):
+                u = transformed[i + j]
+                v = transformed[i + j + half] * w
+                
+                transformed[i + j] = u + v
+                transformed[i + j + half] = u - v
+                
+                w *= w_len
+
+        length = length << 1
+
+    return transformed
+
 @defun
-def fft(ctx, values, inverse=False, pad_to_power_of_two=False):
+def fft(ctx, values, pad_to_power_of_two=False):
     r"""Computes the Fast Fourier Transform (or Inverse FFT) with optional zero-padding.
 
-    If ``inverse=True``, computes the inverse transform and normalizes the
-    result by the sequence length.
     If ``pad_to_power_of_two=True``, the input sequence is automatically zero-padded to the next power of 2 if its length is not already a power of 2.
     """
     orig_n = len(values)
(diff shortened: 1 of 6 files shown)


mpmath/calculus/fft.py
```diff
@@ -0,0 +1,66 @@
...
+    result = _fft_recursive(ctx, converted_values, inverse)
+
+    if inverse:
+        scale = ctx.convert(padded_n)
+        result = [val / scale for val in result]
+
+    return result
+
+@defun
+def ifft(ctx, values, pad_to_power_of_two=False):
```
skirpichev: Probably, you should use ``invfft`` as name.

mpmath/calculus/fft.py
```diff
@@ -0,0 +1,66 @@
...
+        factor *= root
+    return transformed
+
+@defun
+def fft(ctx, values, inverse=False, pad_to_power_of_two=False):
+    r"""Computes the Fast Fourier Transform (or Inverse FFT) with optional zero-padding.
+
+    If ``inverse=True``, computes the inverse transform and normalizes the
+    result by the sequence length.
+    If ``pad_to_power_of_two=True``, the input sequence is automatically zero-padded to the next power of 2 if its length is not already a power of 2.
```
skirpichev: Why by 0, not 42?

mpmath/calculus/fft.py
```diff
@@ -0,0 +1,66 @@
...
+    transformed = [ctx.zero] * n
+    for k in range(half):
+        term = factor * odd[k]
+        transformed[k] = even[k] + term
+        transformed[k + half] = even[k] - term
+        factor *= root
+    return transformed
+
+@defun
+def fft(ctx, values, inverse=False, pad_to_power_of_two=False):
```
skirpichev: Either there should be one function, or no keyword argument.

mpmath/calculus/fft.py
```diff
@@ -0,0 +1,66 @@
...
+    root = ctx.exp(sign * 2 * ctx.pi * ctx.j / n)
+    factor = ctx.one
+    half = n // 2
+    transformed = [ctx.zero] * n
+    for k in range(half):
+        term = factor * odd[k]
+        transformed[k] = even[k] + term
+        transformed[k + half] = even[k] - term
+        factor *= root
+    return transformed
```
skirpichev: I think you could implement this without recursion, just while and for loops.  Rename this function and add reference to the algorithm.

mpmath/calculus/fft.py
```diff
@@ -0,0 +1,66 @@
...
+    if orig_n == 0:
+        return []
+
+    is_power_of_two = (orig_n & (orig_n - 1)) == 0
+
+    if not is_power_of_two:
+        if not pad_to_power_of_two:
+            raise ValueError(f"Length of input ({orig_n}) must be a power of 2.")
+
+        padded_n = 1 << math.ceil(math.log2(orig_n))
```
skirpichev: Alternatively, you can raise NotImplementedError here.

skirpichev (commented):
Generally, looks good.  Few comments.

Let us know, when this will be ready for review.

mpmath/calculus/fft.py
baranwalayush: because adding zeros doesn't introduce any additional information, it simply interpolates the spectrum.

baranwalayush:
I don't understand why the CI fails.

skirpichev:
> I don't understand why the CI fails.

That was github glitch.  See their status page.

mpmath/calculus/fft.py
```diff
@@ -0,0 +1,127 @@
+import math
+
+from .calculus import defun
+
+def _fft_iterative(ctx, values, inverse=False):
+    """
+    This function implements the Cooley-Tukey FFT algorithm iteratively. It computes the Fast Fourier Transform (or Inverse FFT) of a sequence of complex numbers.
+    The input sequence must have a length that is a power of 2. If the length is not a power of 2, the function raises a NotImplementedError.
```
skirpichev: I see no raise statement in this function.

mpmath/calculus/fft.py
```diff
@@ -0,0 +1,127 @@
...
+
+    https://en.wikipedia.org/wiki/Cooley%E2%80%93Tukey_FFT_algorithm#Data_reordering,_bit_reversal,_and_in-place_algorithms
```
skirpichev:
```suggestion
    https://en.wikipedia.org/wiki/Cooley%E2%80%93Tukey_FFT_algorithm
```

mpmath/calculus/fft.py
```diff
@@ -0,0 +1,127 @@
...
+                transformed[i + j + half] = u - v
+
+                w *= w_len
+
+        length <<= 1
+
+    return transformed
+
+@defun
+def fft(ctx, values, pad_to_power_of_two=False):
```
skirpichev: Probably one function is better.

mpmath/calculus/fft.py
skirpichev: > because adding zeros doesn't introduce any additional information

Why zero value is non an additional information?

baranwalayush: The input to `fft()` represents amplitude of a signal taken at different time steps. Since, a signal with zero amplitude doesn't carrry any energy, thus it doesn't introduce any new information. It simply means there is no signal after this point.

skirpichev: > It simply means there is no signal after this point.

That might be false, if actual signal is not zero, right?

mpmath/calculus/fft.py
baranwalayush: That would make the implementation less readable. I feel like it is good to have separate implementation of `fft` and `invfft`.

mpmath/calculus/fft.py
baranwalayush: Ohh, fixed it. Actually since the values passed to it are always via either `fft()` or `invfft()`, it will always receive power of 2 size input only.

mpmath/calculus/fft.py
baranwalayush: No, the actual signal is not changed while padding. For ex: if we get an time domain input like `[1, 1, 1]` which is not power of two size then the algorithm will automatically zero pad it to make `[1, 1, 1, 0]` the size equal to the nearest power of two, if pad_to_power_of_two=True. It doesn't change the actual input but just adds enough no information part to make it just right for the cooley-tuckey algorithm to work.

skirpichev: > It doesn't change the actual input

It does, of course.

> adds enough no information part to make it just right for the cooley-tuckey algorithm to work.

Are you sure that Bluestein algorithm (that can accept same size without need for padding) will produce same (or close in some sense) answer?

baranwalayush: > > It doesn't change the actual input
> 
> It does, of course.

It changes the input, but doen't introduce any new information.

> Are you sure that Bluestein algorithm (that can accept same size without need for padding) will produce same (or close in some sense) answer?

Yes, zero padding essentially interpolates the discrete time spectrum. By increasing the total sequence length from original data length M to padded length N, the DFT bins are spaced closer together: $$\Delta f = \frac{f_s}{N}$$.
Adding these zeros simply evaluates the signal's existing sinc-shaped spectrum at more points. So, the answer wouldn't be the same as the points where the evaluations are being done are different but the actual shape of the signal is maintained. 

As, I mentioned in the PR description too, that this is not the most optimal approach to calculate FFTs, but it is one of the ways which makes it possible for the radix-2 cooley-tuckey algorithm to work with arbitrary sized inputs. That is also the reason why I have set the default pad_to_power_of_two to false, in case if someone is not comfortable with using it. I have also added a TODO line in the program which says to implement the more optimal algorithms.

Here's a simple program (generated using AI) which explains it visually:
```
import numpy as np
import matplotlib.pyplot as plt
from numpy.fft import fft, fftfreq

fs = 100.0          # Sampling rate: 100 Hz
t_short = np.arange(0, 0.2, 1/fs)  # Short signal duration: 0.2 seconds (M = 20 samples)
f0 = 10.0           # Frequency of the sine wave: 10 Hz
f1 = 25.0           # Adding a second frequency component for complexity: 30 Hz

# Original time-domain signal
x_short = np.sin(2 * np.pi * f0 * t_short) + 0.5 * np.sin(2 * np.pi * f1 * t_short) 

# Zero-pad the signal to a power of 2 (N = 32 samples)
N_padded = 32
x_padded = np.pad(x_short, (0, N_padded - len(x_short)), mode='constant')

# Compute FFTs using scipy.fft
X_short = fft(x_short)
X_padded = fft(x_padded)
t_padded = np.arange(0, N_padded/fs, 1/fs)  # Time vector for padded signal

# Calculate corresponding frequency axes
freqs_short = fftfreq(len(x_short), 1/fs)
print(f"Unpadded FFT Frequencies (M = {len(x_short)}): {freqs_short}")
freqs_padded = fftfreq(N_padded, 1/fs)
print(f"Zero-Padded FFT Frequencies (N = {N_padded}): {freqs_padded}")

freqs_dense = np.linspace(0, fs/2, 500)
# Analytical DTFT envelope calculation of the M-point truncated signal
omega = 2 * np.pi * freqs_dense / fs
X_dtft = np.zeros(len(freqs_dense), dtype=complex)
for n in range(20):
    X_dtft += x_short[n] * np.exp(-1j * omega * n)

fig, axes = plt.subplots(2, 2, figsize=(12, 8))

# Plot the original time-domain signal

axes[0, 0].plot(t_short, x_short, 'k.-')
axes[0, 0].set_title("Original Time-Domain Signal (10 Hz Sine Wave + 25 Hz Sine Wave)")
axes[0, 0].set_xlabel("Time (seconds)")
axes[0, 0].set_ylabel("Amplitude")
axes[0, 0].grid(True)

axes[0, 1].plot(t_padded, x_padded, 'k.-')
axes[0, 1].set_title("Padded Time-Domain Signal (10 Hz Sine Wave + 25 Hz Sine Wave)")
axes[0, 1].set_xlabel("Time (seconds)")
axes[0, 1].set_ylabel("Amplitude")
axes[0, 1].grid(True)

# Unpadded FFT Spectrum (Coarse Sampling)
half_short = len(x_short) // 2
axes[1, 0].stem(freqs_short[:half_short], np.abs(X_short[:half_short]), 'b')
axes[1, 0].set_title(f"Unpadded Spectrum (Length M = {len(x_short)}): Coarse Bins")
axes[1, 0].set_ylabel("Magnitude")
axes[1, 0].grid(True)

# Zero-Padded FFT Spectrum (Smooth Interpolation)
half_padded = N_padded // 2
axes[1, 1].plot(freqs_padded[:half_padded], np.abs(X_padded[:half_padded]), 'r.-', label=f'Zero-Padded FFT (N = {N_padded})')
axes[1, 1].plot(freqs_dense, np.abs(X_dtft), 'k--', label='True DTFT Envelope')
axes[1, 1].set_title(f"Zero-Padded Spectrum (Length N = {N_padded}): Finer Bins Interpolate the True Sinc Shape")
axes[1, 1].set_xlabel("Frequency (Hz)")
axes[1, 1].set_ylabel("Magnitude")
axes[1, 1].legend()
axes[1, 1].grid(True)

plt.tight_layout()
plt.show()
```

image

skirpichev: > It changes the input

Obviously, it changes both input and output.

> That is also the reason why I have set the default pad_to_power_of_two to false, in case if someone is not comfortable with using it.

Why someone should be "uncomfortable" when you alter input in the way, that "doen't introduce any new information"?  If that really so, padding should be unconditional.  If not, lets not pretend to handle this case and just raise an exception.

baranwalayush: Zero-padding is a common, long-standing technique used to handle non-power-of-two input sizes in Cooley-Tukey FFTs. It is true that more optimal algorithms like Bluestein’s or Rader’s are often preferred in modern libraries for efficiency, but zero-padding remains a valid and widely documented approach (e.g. [Cooley–Tukey Wikipedia section](https://en.wikipedia.org/wiki/Cooley%E2%80%93Tukey_FFT_algorithm), [Analysis of DFT and FFT](https://www.jetir.org/papers/JETIR2405G08.pdf), [Bitweenie FFT guide](https://www.bitweenie.com/listings/fft-zero-padding) etc.)

That said, I’m open to your thoughts, if you’d prefer to omit this zero-padding support, I'd be happy to update the PR accordingly.

skirpichev: > Zero-padding is a common, long-standing technique used to handle non-power-of-two input sizes in Cooley-Tukey FFTs.

Well, "scipy does this" is an argument, of course.

> That said, I’m open to your thoughts, if you’d prefer to omit this zero-padding support

Yes, it should be either removed or be unconditional.

baranwalayush: As rightly pointed by you earlier that, zero padding changes the output and input, so having it unconditional should be avoided because altering signal lengths modifies frequency bin spacing, and in some cases the user might just want to get a simple fft rather than the interpolated version. However, eliminating zero-padding entirely restricts the API, as discussed earlier under the issue, and removes a legitimate signal processing technique for handling such inputs in cooley-tukey solvers and spectral interpolation. So, providing an optional parameter gives users explicit control without making assumptions about their workflow.

mpmath/calculus/fft.py
```diff
@@ -0,0 +1,133 @@
+import math
+
+from .calculus import defun
+
+def _fft_iterative(ctx, values, inverse=False):
+    """
+    This function implements the Cooley-Tukey FFT algorithm iteratively. It computes the Fast Fourier Transform (or Inverse FFT) of a sequence of complex numbers.
```
skirpichev: Maybe you should name function accordingly?  Current name doesn't reflect that it uses given algorithm.

BTW, also please wrap line, as well as others, including comments.

mpmath/calculus/fft.py
```diff
@@ -0,0 +1,133 @@
...
+
+    return transformed
+
+# TODO: Add support for non-power-of-two lengths using Bluestein's algorithm or similar methods.
+
+@defun
+def fft(ctx, values, pad_to_power_of_two=False):
+    r"""Computes the Fast Fourier Transform (or Inverse FFT) with optional zero-padding.
+    If ``pad_to_power_of_two=True``, the time domain sequence is automatically zero-padded at the end to the next power of 2 if its length is not already a power of 2.
+    Raises NotImplementedError if the input sequence length is not a power of 2 and ``pad_to_power_of_two`` is False.
```
skirpichev: I don't think that this API is acceptable.  Either make padding unconditionally or raise NotImplementedError for inappropriate `len(values)`.

skirpichev (changes requested):

mpmath/calculus/fft.py
baranwalayush: > Maybe you should name function accordingly? Current name doesn't reflect that it uses given algorithm.

changed it to `_fft_cooley_tuckey()`

> BTW, also please wrap line, as well as others, including comments.

sure, updated

mpmath/calculus/fft.py
baranwalayush: Ok, I have now removed the zero padding part completely and now it would simply raise `NotImplementedError`, if it encounters non power of two inputs.

mpmath/calculus/fft.py
```diff
@@ -0,0 +1,90 @@
+from .calculus import defun
+
+def _fft_cooley_tuckey(ctx, values, inverse=False):
+    """
+    This function implements the Radix-2 Cooley-Tukey FFT algorithm iteratively. It computes the Fast Fourier Transform (or Inverse FFT) of a sequence of complex numbers.
```
skirpichev:
```suggestion
    This function implements the Radix-2 Cooley-Tukey FFT algorithm iteratively.
    It computes the Fast Fourier Transform (or Inverse FFT) of a sequence of
    complex numbers.
```

mpmath/calculus/fft.py
```diff
@@ -0,0 +1,90 @@
...
+                rev |= 1 << (num_bits - 1 - b)
+        transformed[rev] = values[i]
+
+    sign = 1 if inverse else -1
+    two_pi_j = 2 * ctx.pi * ctx.j
+
+    length = 2
+    while length <= n:
+        half = length // 2
+        w_len = ctx.exp(sign * two_pi_j / length)
```
skirpichev:
```suggestion
        w_len = ctx.expjpi(2 * sign / length)
```
and make sign = +/-ctx.one.

mpmath/calculus/fft.py
```diff
@@ -0,0 +1,90 @@
...
+                transformed[i + j] = u + v
+                transformed[i + j + half] = u - v
+
+                w *= w_len
+
+        length <<= 1
+
+    return transformed
+
+# TODO: Add support for non-power-of-two lengths using Bluestein's algorithm or similar methods.
```
skirpichev: Lets rather open an issue (or reopen old one).

mpmath/calculus/fft.py
```diff
@@ -0,0 +1,90 @@
...
+
+@defun
+def fft(ctx, values):
+    r"""Computes the Discrete Fourier Transform (DFT) of a sequence using the Radix-2 Cooley-Tukey FFT algorithm.
+    Raises NotImplementedError if the input sequence length is not a power of 2.
```
skirpichev:
```suggestion
    r"""
    Computes the Discrete Fourier Transform (DFT) of a sequence.

    Raises NotImplementedError if the input sequence length is not a power of 2.
```

Same for invfft.

mpmath/calculus/fft.py
```diff
@@ -0,0 +1,90 @@
...
+    """
+    n = len(values)
+    if n == 0:
+        return []
+
+    is_power_of_two = (n & (n - 1)) == 0
+
```
skirpichev:
```suggestion
```

mpmath/calculus/fft.py
```diff
@@ -0,0 +1,90 @@
...
+    if not is_power_of_two:
+        raise NotImplementedError(f"FFT is only implemented for lengths that are powers of 2, got length: {n}")
+
+    converted_values = [ctx.convert(v) for v in values]
+
+    result = _fft_cooley_tuckey(ctx, converted_values, False)
+
+    return result
```
skirpichev:
```suggestion
    converted_values = [ctx.convert(v) for v in values]
    return _fft_cooley_tuckey(ctx, converted_values, False)
```

mpmath/calculus/fft.py
```diff
@@ -0,0 +1,90 @@
...
+
+@defun
+def invfft(ctx, values):
+    r"""Computes the inverse Discrete Fourier Transform (IDFT) of a sequence using the Radix-2 Cooley-Tukey FFT algorithm.
+    Raises NotImplementedError if the input sequence length is not a power of 2.
```
skirpichev:
```suggestion
    r"""
    Computes the inverse Discrete Fourier Transform (IDFT) of a sequence.

    Raises NotImplementedError if the input sequence length is not a power of 2.
```

mpmath/calculus/fft.py
```diff
@@ -0,0 +1,90 @@
...
+    """
+
```
skirpichev, line 87:
```suggestion
```

mpmath/calculus/fft.py
```diff
@@ -0,0 +1,90 @@
...
+    n = len(values)
+    if n == 0:
+        return []
+
+    is_power_of_two = (n & (n - 1)) == 0
+
```
skirpichev:
```suggestion
```

mpmath/calculus/fft.py
```diff
@@ -0,0 +1,90 @@
...
+    if not is_power_of_two:
+        raise NotImplementedError(f"Inverse FFT is only implemented for lengths that are powers of 2, got length: {n}")
+
+    converted_values = [ctx.convert(v) for v in values]
+
+    result = _fft_cooley_tuckey(ctx, converted_values, True)
+
+    result = [val / n for val in result]
+
+    return result
```
skirpichev:
```suggestion
    converted_values = [ctx.convert(v) for v in values]
    result = _fft_cooley_tuckey(ctx, converted_values, True)
    return [val / n for val in result]
```

mpmath/tests/test_calculus.py
```diff
@@ -2,9 +2,9 @@
 
 from mpmath import (arange, chebyfit, cos, cosm, differint, e, euler, exp,
                     expm, fourier, fourierval, inf, invertlaplace, j, limit,
-                    log, matrix, mp, mpf, norm, pade, pi, polyroots, polyval,
-                    sin, sinm, sqrt, logm)
-
+                    log, matrix, mp, fp, mpf, norm, pade, pi, polyroots, polyval,
```
skirpichev: fp seems unused

mpmath/tests/test_calculus.py
```diff
@@ -2,9 +2,9 @@
...
+                    sin, sinm, sqrt, logm, fft, invfft)
+import random
```
skirpichev: Move this up, befor pytest import and run isort on the file.

mpmath/tests/test_calculus.py
```diff
@@ -276,3 +276,38 @@ def test_logm():
...
+    expected = [1, -1j, -1, 1j]
+    assert all(a.ae(b) for a, b in zip(spectrum, expected))
+
+    spectrum = fft([1, 2, 3, 4])
+    expected = [10, -2 + 2j, -2, -2 - 2j]
+    assert all(a.ae(b) for a, b in zip(spectrum, expected))
+    assert mp.chop(invfft(spectrum)) == [1, 2, 3, 4]
+
+    random.seed(42)
+    # test that fft and invfft are inverses of each other
```
skirpichev: Maybe test also complex input?

mpmath/tests/test_calculus.py
```diff
@@ -276,3 +276,38 @@ def test_logm():
...
+    assert all(a.ae(b) for a, b in zip(recovered, x))
+
+    X = [mp.rand() for _ in range(8)]
+    recovered = fft(invfft(X))
+    assert all(a.ae(b) for a, b in zip(recovered, X))
+
+    assert invfft([]) == []
+    pytest.raises(NotImplementedError, lambda: invfft([1, 2, 3]))
+
+    # test parseval's theorem
```
skirpichev, line 308: and here

mpmath/calculus/fft.py
```diff
@@ -0,0 +1,82 @@
...
+
+    Raises NotImplementedError if the input sequence length is not a power of 2.
+    """
+    n = len(values)
+    if n == 0:
+        return []
+
+    is_power_of_two = (n & (n - 1)) == 0
+    if not is_power_of_two:
+        raise NotImplementedError(f"FFT is only implemented for lengths that are powers of 2, got length: {n}")
```
skirpichev:
```suggestion
        raise NotImplementedError("FFT is only implemented for lengths that "
                                  f"are powers of 2, got length: {n}")
```

mpmath/calculus/fft.py
```diff
@@ -0,0 +1,82 @@
...
+
+    Raises NotImplementedError if the input sequence length is not a power of 2.
+    """
+    n = len(values)
+    if n == 0:
+        return []
+
+    is_power_of_two = (n & (n - 1)) == 0
+    if not is_power_of_two:
+        raise NotImplementedError(f"Inverse FFT is only implemented for lengths that are powers of 2, got length: {n}")
```
skirpichev:
```suggestion
        raise NotImplementedError("Inverse FFT is only implemented for lengths that "
                                  f"are powers of 2, got length: {n}")
```

skirpichev (changes requested):
Please add also examples to docstrings and test complex input in randomized tests.

Probably, random tests should be factored to separate functions.  And, as you don't use non-default precision - please use hypothesis.

Lets hope it's a final review turn.  Thanks for your patience.

mpmath/calculus/fft.py
```diff
@@ -0,0 +1,103 @@
...
+
+    A few basic examples are:
+
+        >>> from mpmath import mp, fft
+        >>> mp.nprint(fft([1, 0, 0, 0]))
+        [1.0, (1.0 + 0.0j), 1.0, (1.0 + 0.0j)]
+        >>> mp.nprint(fft([1 + 2j, 1 + 2j]))
+        [(2.0 + 4.0j), (0.0 + 0.0j)]
+        >>> mp.nprint(fft([1, 2, 3, 4]))
+        [10.0, (-2.0 + 2.0j), -2.0, (-2.0 - 2.0j)]
```
skirpichev:
```suggestion
    **Examples**

    >>> from mpmath import mp
    >>> mp.pretty = True
    >>> mp.fft([1, 0, 0, 0])
    [1.0, (1.0 + 0.0j), 1.0, (1.0 + 0.0j)]
    >>> mp.fft([1 + 2j, 1 + 2j])
    [(2.0 + 4.0j), (0.0 + 0.0j)]
    >>> mp.fft([1, 2, 3, 4])
    [10.0, (-2.0 + 2.0j), -2.0, (-2.0 - 2.0j)]
```

mpmath/calculus/fft.py
```diff
@@ -0,0 +1,103 @@
...
+    Raises NotImplementedError if the input sequence length is not a power of 2.
+
+    A few basic examples are:
+
+        >>> from mpmath import mp, fft, invfft
+        >>> mp.nprint(invfft([1, 1, 1, 1]))
+        [1.0, (0.0 + 0.0j), 0.0, (0.0 + 0.0j)]
+        >>> x = [1, 2, 3, 4]
+        >>> mp.nprint(invfft(fft(x)))
+        [(1.0 + 0.0j), (2.0 + 0.0j), (3.0 + 0.0j), (4.0 + 0.0j)]
```
skirpichev:
```suggestion
    **Examples**

    >>> from mpmath import mp
    >>> mp.pretty = True
    >>> mp.invfft([1, 1, 1, 1])
    [1.0, (0.0 + 0.0j), 0.0, (0.0 + 0.0j)]
    >>> x = [1, 2, 3, 4]
    >>> mp.invfft(mp.fft(x))
    [(1.0 + 0.0j), (2.0 + 0.0j), (3.0 + 0.0j), (4.0 + 0.0j)]
```

mpmath/calculus/fft.py
```diff
@@ -0,0 +1,103 @@
...
+    if n == 0:
+        return []
+
+    is_power_of_two = (n & (n - 1)) == 0
+    if not is_power_of_two:
+        raise NotImplementedError("FFT is only implemented for lengths that "
+                                  f"are powers of 2, got length: {n}")
+
+    converted_values = [ctx.convert(v) for v in values]
+    return _fft_cooley_tuckey(ctx, converted_values, False)
```
skirpichev: I think it would be better to use a different working precision:
```suggestion
    with ctx.extraprec(10):
        result = _fft_cooley_tuckey(ctx, converted_values)
    return [+v for v in result]
```

mpmath/calculus/fft.py
```diff
@@ -0,0 +1,103 @@
...
+        return []
+
+    is_power_of_two = (n & (n - 1)) == 0
+    if not is_power_of_two:
+        raise NotImplementedError("Inverse FFT is only implemented for lengths that "
+                                  f"are powers of 2, got length: {n}")
+
+    converted_values = [ctx.convert(v) for v in values]
+    result = _fft_cooley_tuckey(ctx, converted_values, True)
+    return [val / n for val in result]
```
skirpichev:
```suggestion
    with ctx.extraprec(10):
        result = _fft_cooley_tuckey(ctx, converted_values, True)
    return [val / n for val in result]
```

mpmath/tests/test_calculus.py
```diff
@@ -276,3 +278,59 @@ def test_logm():
...
+        ),
+        min_size=size,
+        max_size=size,
+    ))
+
+@given(x=power_of_two_signals())
+def test_fft_randomized_complex(x):
+    # test that fft and invfft are inverses of each other for random complex inputs
+    recovered = invfft(fft(x))
+    assert all(mp.almosteq(a, b, rel_eps=1e-12) for a, b in zip(recovered, x))
```
skirpichev: With extra working precision you might use default rel/abs_eps:
```suggestion
    assert all(a.ae(b) for a, b in zip(recovered, x))
```

mpmath/tests/test_calculus.py
```diff
@@ -276,3 +278,59 @@ def test_logm():
...
+
+    recovered = fft(invfft(x))
+    assert all(mp.almosteq(a, b, rel_eps=1e-12) for a, b in zip(recovered, x))
```
skirpichev:
```suggestion
    assert all(a.ae(b) for a, b in zip(recovered, x))
```

skirpichev (approved):

mpmath/calculus/fft.py
```diff
@@ -0,0 +1,108 @@
...
+        return values
+
+    # Bit-Reversal Permutation
+    transformed = [ctx.zero] * n
+    num_bits = n.bit_length() - 1
+    for i in range(n):
+        rev = 0
+        for b in range(num_bits):
+            if (i >> b) & 1:
+                rev |= 1 << (num_bits - 1 - b)
```
skirpichev:
```suggestion
    for i in range(n):
        rev = i ^ ((1<<i.bit_length()) - 1)
```

skirpichev:
@baranwalayush, please adjust description.  I'll use it as a commit message draft.

mpmath/calculus/fft.py
baranwalayush: The suggested change performs bit inversion (except zero), not bit reversal. Our aim in this step is to map each element at index i  to a new index formed by reversing the binary representation of i, to simulate the recursive even odd splitting.
For example:  n = 8, binary index (b2​ b1 ​b0​) -> reversed binary (b0 ​b1 ​b2​)

skirpichev: Indeed, my fault.

Though, maybe you can avoid repeated shifts.  Something like this, inline:
```py
def reverse_bits(n, width):
    result = 0
    for _ in range(width):
        result <<= 1
        result |= (n & 1)
        n >>= 1
    return result
```

baranwalayush: the one I wrote sets the (num_bits - 1 - b) th bit if the bth bit is set in n. Well, this one definitely looks cleaner and uses less shift. Ok, will update it.

mpmath/calculus/fft.py
```diff
@@ -0,0 +1,113 @@
...
+
+    # Bit-Reversal Permutation
+    def reverse_bits(n, width):
+        result = 0
+        for _ in range(width):
+            result <<= 1
+            result |= (n & 1)
+            n >>= 1
+        return result
+
```
skirpichev:
```suggestion
    # Bit-Reversal Permutation
```

mpmath/calculus/fft.py
```diff
@@ -0,0 +1,113 @@
...
+    transformed = [ctx.zero] * n
+    num_bits = n.bit_length() - 1
+    for i in range(n):
+        rev = reverse_bits(i, num_bits)
```
skirpichev:
```suggestion
        rev = 0
        for _ in range(num_bits):
            rev <<= 1
            rev |= (i & 1)
            i >>= 1           
```

mpmath/calculus/fft.py
```diff
@@ -0,0 +1,109 @@
...
+    """
+    n = len(values)
+    if n <= 1:
+        return values
+
+    # Bit-Reversal Permutation
+    transformed = [ctx.zero] * n
+    num_bits = n.bit_length() - 1
+    for i in range(n):
+        rev = 0
```
skirpichev, line 19:
```suggestion
        rev = 0
        val = values[i]
```

mpmath/calculus/fft.py
```diff
@@ -0,0 +1,109 @@
...
+        for _ in range(num_bits):
+            rev <<= 1
+            rev |= (i & 1)
+            i >>= 1           
+        transformed[rev] = values[i]
```
skirpichev:
```suggestion
        transformed[rev] = val
```

mpmath/calculus/fft.py
```diff
@@ -0,0 +1,110 @@
...
+    # Bit-Reversal Permutation
+    transformed = [ctx.zero] * n
+    num_bits = n.bit_length() - 1
+    for i in range(n):
+        rev = 0
+        val = values[i]
+        for _ in range(num_bits):
+            rev <<= 1
+            rev |= (i & 1)
+            i >>= 1
```
skirpichev:
```suggestion
            rev <<= 1
            rev |= (i & 1)
            i >>= 1
```

PR #1115, opened 2026-06-26
Use shortest decimal representations for str/repr

skirpichev:
Was:
```
$ python -m mpmath
>>> 0.1
0.10000000000000001
```

Now:
```
$ python -m mpmath --shortest-str
>>> 0.1
0.1
```

Closes #1028

ToDo:

- [x] cleanup
- [x] rename option (mp.format_style ?)
- [ ] deprecate current default for CLI
- [ ] deprecate mp.pretty_dps, see #933 and #927
- [x] document new context option
- [x] benchmark
- [x] optimize exponent computation
- [x] huge exponents? (e.g. test_gamma_huge_1())

See also ChezScheme implementation: https://github.com/cisco/ChezScheme/blob/main/s/print.ss#L426

PR #1157, opened 2026-08-15
Optimize conversion from Weierstrass invariants to half-periods

HeskethGD:
## Motivation

Passing `g2, g3` directly is a popular and natural way to evaluate the Weierstrass elliptic functions. For repeated evaluations, precomputing `omega1, omega2` remains the fastest approach, since otherwise the conversion from invariants to half-periods is performed on every call. However, it would be useful to reduce the timing difference between these two interfaces.

This PR optimizes the internal conversion performed by `omega1omega2from(g2=..., g3=...)`. It does not change the public API, function signatures, documented period conventions, or accepted inputs.

## Approach

Previously, the general conversion recovered `tau` through the inverse Klein j-invariant and then reconstructed the period scale using Eisenstein series and theta constants.

The optimized implementation instead obtains the elliptic parameter and period scale together:

- For real invariants with three real cubic roots, it uses the trigonometric cubic solution followed by complete elliptic integrals.
- For real invariants with one real root, it uses real Cardano radicals and a quadratic transformation, keeping the elliptic-integral calculations real.
- For generic complex invariants, it solves the cubic directly and constructs the periods using AGM formulas.
- The lemniscatic case uses its closed-form period.
- Near the cusp and elliptic points, where the direct root formulas become singular or ill-conditioned, it retains the previous inverse-j/Eisenstein route as a fallback.
- The fallback Eisenstein calculation uses Jacobi's theta identity to avoid one theta-constant evaluation.

The resulting basis is reduced and oriented according to the existing `omega1omega2from` convention.

## Timings

These are median wall-clock timings at 50 decimal digits using Python 3.13.4 on macOS arm64, with five timing repeats.

### Half-period conversion

| Input cases | Before (ms/call) | After (ms/call) | Speedup |
|---|---:|---:|---:|
| Mean of 10 generic real cases | 0.937 | 0.112 | 8.4x |
| Representative generic complex case | 0.918 | 0.550 | 1.7x |

### Relative cost of passing invariants

The following table reports time when passing `g2`, `g3` / time when passing precomputed `omega1`, `omega2`.

A value of `1.00x` would mean that passing `g2, g3` adds no measurable conversion overhead. For example, `2.38x` means that the call using `g2, g3` took 2.38 times as long as the corresponding call using precomputed periods. Lower values are therefore better.

Each column is the mean ratio across the same 10 generic real invariant pairs:

- **Before**: the ratio using the implementation on `master`.
- **After**: the ratio using this optimized implementation.

| Function | Before | After |
|---|---:|---:|
| `weierp` | 2.38x | 1.28x |
| `weierpprime` | 1.74x | 1.17x |
| `weiersigma` | 2.81x | 1.32x |
| `weierzeta` | 2.11x | 1.21x |

Precomputing `omega1, omega2` therefore remains worthwhile for repeated calculations, but using the direct `g2, g3` interface is substantially less expensive as a result of this PR.

## Validation against Wolfram Engine

The optimized `omega1omega2from(g2=..., g3=...)` was compared with Wolfram Engine's `WeierstrassHalfPeriods[{g2, g3}]` at 60 decimal digits.

The 245 deterministic cases comprised:

- 100 random real invariant pairs;
- 100 random complex invariant pairs;
- 5 fixed and special cases;
- 24 cases on both sides of the direct/Eisenstein fallback threshold;
- 16 cases approaching the discriminant-zero cusp from both sides.

| Result | Cases |
|---|---:|
| Same ordered and signed period pair | 244 |
| Equivalent period basis | 1 |
| Failures | 0 |
| Total | 245 |

The maximum relative error among the 244 direct matches was `1.754314e-35`.

The sole equivalent-basis result was the lemniscatic case `(g2, g3) = (1, 0)`. If Wolfram returns `(w1, w2)`, mpmath returns `(-w2, w1)`. This is a determinant-one change of basis representing the same period lattice, with `tau = i`. It is also present on master and does not constitute a change in expected output.

#### Comparison with Wolfram Engine Timings

The following timings use the same 10 real and 10 complex invariant pairs, with 50-digit approximate inputs:

| Case group | mpmath mean | Wolfram mean | Comparison |
|---|---:|---:|---:|
| Generic real invariants | 0.112 ms/call | 0.152 ms/call | mpmath is approximately 1.36x faster |
| Generic complex invariants | 0.562 ms/call | 0.229 ms/call | Wolfram is approximately 2.45x faster |

### Possible future work

The complex path constructs the periods using two complex AGM evaluations. These account for approximately 0.241 ms of the 0.571 ms measured for the complete complex conversion, or about 42% of its runtime.

The current [`mpc_agm` implementation](https://github.com/mpmath/mpmath/blob/master/mpmath/libmp/libhyper.py#L1010-L1037) already contains explicit TODO items to:

- check that convergence works as intended;
- optimize the implementation;
- select a nonarbitrary branch.

In particular, `mpc_agm` uses generic complex arithmetic and repeatedly computes complex absolute values when checking convergence. Profiling identified this as the largest individual cost in the complex invariant-to-period path.

Optimizing `mpc_agm` is therefore a promising direction for future work, although it would not account for the entire difference from Wolfram. The remaining cost includes complex cubic radicals, square roots, root construction, modular-basis reduction, and other Python-level multiprecision operations.

## Tests

The added tests cover:

- round trips from invariants to periods and back at several precisions;
- real and complex invariants in the different discriminant regions;
- reduction of the resulting period ratio to the fundamental domain;
- continuity across the threshold between the direct and fallback methods.

The focused elliptic test file passes:

```text
51 passed
```

## AI use declaration

OpenAI GPT-5, through Codex, was used to assist with investigating the performance bottlenecks, implementing the optimization, developing tests and benchmark scripts, running and interpreting the Wolfram Engine comparison. I designed the scope of the PR, suggested things to explore, reviewed all generated changes and results, and ran test suites.

## Reproducibility

Below is a minimal timing script to compare this branch and master. It shows comparable numbers to the tables reported above.

```
git switch master
python my_data/benchmark_omega1omega2from_pyperf.py -o /tmp/omega-master.json

git switch omega-period-optimize-pr
python my_data/benchmark_omega1omega2from_pyperf.py -o /tmp/omega-pr.json

python -m pyperf compare_to \
    /tmp/omega-master.json \
    /tmp/omega-pr.json \
    --table
```

```python
#!/usr/bin/env python3
"""Benchmark invariant-to-half-period conversion for real and complex cases."""

import subprocess
import time

import pyperf

import mpmath
from mpmath import mp, mpc, mpf, omega1omega2from


mp.dps = 50

REAL_CASES = [
    (mpf(60), mpf(140)),
    (mpf(12), mpf(1)),
    (mpf(12), mpf(-1)),
    (mpf(4), mpf(1)),
    (mpf(4), mpf(-1)),
    (mpf(1), mpf(1)),
    (mpf(1), mpf(-1)),
    (mpf(10), mpf(5)),
    (mpf(100), mpf(1)),
    (mpf(-4), mpf(1)),
]

COMPLEX_CASES = [
    (mpc(1, 2), mpc(3, -4)),
    (mpc(-2, 1), mpc('0.5', 3)),
    (mpc(5, -7), mpc(-3, -2)),
    (mpc(2, '0.25'), mpc(-1, '1.5')),
    (mpc(-6, -2), mpc(4, '0.75')),
    (mpc('0.125', 3), mpc('-2.5', '0.5')),
    (mpc(10, 1), mpc(1, -8)),
    (mpc(-1, '0.1'), mpc('-0.25', -2)),
    (mpc(3, -5), mpc(7, 2)),
    (mpc(-4, 6), mpc(-5, 3)),
]


def bench_cases(loops, cases):
    start = time.perf_counter()
    for _ in range(loops):
        for g2, g3 in cases:
            omega1omega2from(g2=g2, g3=g3)
    return time.perf_counter() - start


    ['git', 'rev-parse', '--short', 'HEAD'], text=True).strip()
runner = pyperf.Runner(metadata={
    'mpmath_revision': revision,
    'mpmath_file': mpmath.__file__,
    'mpmath_dps': mp.dps,
})
runner.bench_time_func(
    'omega1omega2from: real mean', bench_cases, REAL_CASES,
    inner_loops=len(REAL_CASES))
runner.bench_time_func(
    'omega1omega2from: complex mean', bench_cases, COMPLEX_CASES,
    inner_loops=len(COMPLEX_CASES))
```

### DIFF
--- a/mpmath/functions/elliptic.py
+++ b/mpmath/functions/elliptic.py
@@ -1718,30 +1718,34 @@ def g2g3from(ctx, q=None, m=None, k=None, tau=None, qbar=None,
         "g2g3from", q, m, k, tau, qbar, g2, g3, omega1, omega2)
     with ctx.extraprec(10):
         if g2 is not None:
-            return +ctx.convert(g2), +ctx.convert(g3)
-        if omega1 is None:
-            tau = ctx.taufrom(q=q, m=m, k=k, tau=tau, qbar=qbar)
-            if ctx.im(tau) <= 0:
-                raise ValueError("g2g3from: tau must be in upper half-plane")
-            omega1 = ctx.one/2
-            omega2 = tau/2
+            g2 = ctx.convert(g2)
+            g3 = ctx.convert(g3)
         else:
-            omega1 = ctx.convert(omega1)
-            omega2 = ctx.convert(omega2)
-        if ctx.im(omega2/omega1) <= 0:
-            raise ValueError("g2g3from: omega ratio must be "
-                             "in upper half-plane")
-        tau = omega2 / omega1
-        q = ctx.qfrom(tau=tau)
-        j2 = ctx.jtheta(2, 0, q)
-        j3 = ctx.jtheta(3, 0, q)
-        factor = ctx.pi / (2 * omega1)
-        g2 = (ctx.mpf(4)/3) * factor**4 * (j2**8 - (j2*j3)**4 + j3**8)
-        g3 = ((ctx.mpf(8)/27) * factor**6 *
-              (j2**12 - (ctx.mpf(3)/2*j2**8*j3**4 +
-                         ctx.mpf(3)/2*j2**4*j3**8) +
-               j3**12))
-        return +g2, +g3
+            if omega1 is None:
+                tau = ctx.taufrom(q=q, m=m, k=k, tau=tau, qbar=qbar)
+                if ctx.im(tau) <= 0:
+                    raise ValueError("g2g3from: tau must be in upper "
+                                     "half-plane")
+                omega1 = ctx.one/2
+                omega2 = tau/2
+            else:
+                omega1 = ctx.convert(omega1)
+                omega2 = ctx.convert(omega2)
+            if ctx.im(omega2/omega1) <= 0:
+                raise ValueError("g2g3from: omega ratio must be "
+                                 "in upper half-plane")
+            tau = omega2 / omega1
+            q = ctx.qfrom(tau=tau)
+            j2 = ctx.jtheta(2, 0, q)
+            j3 = ctx.jtheta(3, 0, q)
+            factor = ctx.pi / (2 * omega1)
+            g2 = ((ctx.mpf(4)/3) * factor**4 *
+                  (j2**8 - (j2*j3)**4 + j3**8))
+            g3 = ((ctx.mpf(8)/27) * factor**6 *
+                  (j2**12 - (ctx.mpf(3)/2*j2**8*j3**4 +
+                             ctx.mpf(3)/2*j2**4*j3**8) +
+                   j3**12))
+    return +g2, +g3
 
 @defun
 def omega1omega2from(ctx, q=None, m=None, k=None, tau=None, qbar=None,
@@ -1778,23 +1782,28 @@ def omega1omega2from(ctx, q=None, m=None, k=None, tau=None, qbar=None,
     """
     _validate_weierstrass_parameter_args(
         "omega1omega2from", q, m, k, tau, qbar, g2, g3, omega1, omega2)
-    with ctx.extraprec(10):
-        if omega1 is not None:
+    if omega1 is not None:
+        with ctx.extraprec(10):
             omega1 = ctx.convert(omega1)
             omega2 = ctx.convert(omega2)
             if ctx.im(omega2/omega1) <= 0:
                 raise ValueError("omega1omega2from: omega ratio must be "
                                  "in upper half-plane")
-            return +omega1, +omega2
-        if g2 is None:
+        return +omega1, +omega2
+    if g2 is None:
+        with ctx.extraprec(10):
             tau = ctx.taufrom(q=q, m=m, k=k, tau=tau, qbar=qbar)
             if ctx.im(tau) <= 0:
                 raise ValueError("omega1omega2from: tau must be in upper "
                                  "half-plane")
-            return +(ctx.one/2), +(tau/2)
+            omega1 = ctx.one/2
+            omega2 = tau/2
+        return +omega1, +omega2
 
+    with ctx.extraprec(10):
         g2 = ctx.convert(g2)
         g3 = ctx.convert(g3)
+        periods = None
 
         if g2 == 0:
             omegaA = (g3 ** (ctx.mpf(-1)/ctx.mpf(6)) *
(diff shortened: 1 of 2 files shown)


mpmath/functions/elliptic.py
```diff
@@ -1820,26 +1800,141 @@ def omega1omega2from(ctx, q=None, m=None, k=None, tau=None, qbar=None,
...
+                                  ctx.mpf(7)/ctx.mpf(12))
+                omegaB = tau * omegaA
+            elif ctx.im(g2) == 0 and ctx.im(g3) == 0:
+                real_g2 = ctx.re(g2)
+                real_g3 = ctx.re(g3)
+                real_discriminant = ctx.re(discriminant)
+                if real_discriminant > 0:
+                    # Three real roots. A trigonometric solution avoids the
+                    # complex intermediate values in Cardano's formula. The
+                    # ordered roots give a real elliptic parameter m.
```
skirpichev, line 1844: Have you compared this with usual formula wrt performance?

mpmath/functions/elliptic.py
```diff
@@ -1820,26 +1800,141 @@ def omega1omega2from(ctx, q=None, m=None, k=None, tau=None, qbar=None,
...
+                    m = (e2-e3)/D
+                    sqrt_D = ctx.sqrt(D)
+                    real_period = ctx.ellipk(m)/sqrt_D
+                    imaginary_period = ctx.ellipk(1-m)/sqrt_D
+
+                    # For m <= 1/2 the standard rectangular basis is already
+                    # reduced. For m > 1/2 apply its S-transform directly.
+                    if m <= ctx.one/2:
+                        return +real_period, +(ctx.j*imaginary_period)
+                    return +(-ctx.j*imaginary_period), +real_period
```
skirpichev: Unfortunately, this is not a correct way to do normalization.  Even besides totally useless unary `+` in case like `+(ctx.j*imaginary_period)`.

This is something I overlooked before, sorry; this happens in several other parts of this function.

As you are in the extraprec context, you can assign this to variable and then - normalize result *outside* of the context.

skirpichev (commented):
Mostly looks ok.  One question and one problem (normalization), that should be fixed (perhaps, in a separate commit).

For benchmarks.   Unfortunately, you provided only results, again.  

I would appreciate something better reproducible, e.g. if you share somehow scripts, used for benchmarks (especially wrt the mpmath master).  See an example in the PR nearby: https://github.com/mpmath/mpmath/pull/1115#issuecomment-5229311640

mpmath/functions/elliptic.py
HeskethGD: Thanks, fixed in a separate commit on this PR (`49c7eeb4`). Both `omega1omega2from` and `g2g3from` now perform calculations inside `extraprec` but normalize their results after leaving the context. I also removed the redundant unary `+` operations inside the context and added a regression test checking that returned values do not retain guard bits. The focused elliptic tests and doctests pass and I checked the benchmark results were not noticeably effected.

HeskethGD:
> Mostly looks ok. One question and one problem (normalization), that should be fixed (perhaps, in a separate commit).
> 
> For benchmarks. Unfortunately, you provided only results, again.
> 
> I would appreciate something better reproducible, e.g. if you share somehow scripts, used for benchmarks (especially wrt the mpmath master). See an example in the PR nearby: [#1115 (comment)](https://github.com/mpmath/mpmath/pull/1115#issuecomment-5229311640)

Added a minimal reproducibility script to the PR description to compare this branch to master.

mpmath/functions/elliptic.py
HeskethGD: Yes. I benchmarked the trigonometric solution against the usual Cardano formula for six representative three-real-root cases at 50 dps.

- Root calculation: **51.0 μs** with the trigonometric formula versus **79.4 μs** with Cardano (**1.56× faster**).
- Complete period calculation: **87.2 μs** versus **118 μs** when Cardano’s results are explicitly converted back to real values (**1.35× faster**).
- Leaving Cardano’s complex intermediate values unchanged increased the complete calculation to **360 μs**.

Thus, most of the larger improvement comes from preventing complex arithmetic downstream, but the trigonometric formula is also measurably faster for the root calculation itself. The resulting periods were checked to agree numerically in all benchmark cases.

```
#!/usr/bin/env python3
"""Isolate trigonometric and Cardano roots when the cubic has 3 real roots."""

import time

import pyperf

from mpmath import mp


mp.dps = 50

CASES = [
    (mp.mpf(12), mp.mpf(1)),
    (mp.mpf(12), mp.mpf(-1)),
    (mp.mpf(4), mp.mpf(1)),
    (mp.mpf(4), mp.mpf(-1)),
    (mp.mpf(10), mp.mpf(5)),
    (mp.mpf(100), mp.mpf(1)),
]


def trig_roots(g2, g3):
    root_scale = mp.sqrt(g2/3)
    cosine = 3*mp.sqrt(3)*g3/(g2*mp.sqrt(g2))
    theta = mp.acos(cosine)/3
    return (
        root_scale*mp.cos(theta),
        root_scale*mp.cos(theta-2*mp.pi/3),
        root_scale*mp.cos(theta+2*mp.pi/3),
    )


def cardano_roots(g2, g3):
    discriminant = g2**3 - 27*g3**2
    root_discriminant = mp.sqrt(-discriminant/1728)
    u = (g3/8 + root_discriminant)**(mp.mpf(1)/3)
    v = g2/(12*u)
    zeta = mp.expjpi(mp.mpf(2)/3)
    zeta2 = -1-zeta
    roots = (u+v, zeta*u+zeta2*v, zeta2*u+zeta*v)
    return tuple(sorted(roots, key=mp.re, reverse=True))


def cardano_real_roots(g2, g3):
    return tuple(mp.re(root) for root in cardano_roots(g2, g3))


def periods_from_roots(roots):
    e1, e2, e3 = roots
    period_scale = e1-e3
    m = (e2-e3)/period_scale
    sqrt_scale = mp.sqrt(period_scale)
    return (mp.ellipk(m)/sqrt_scale,
            mp.j*mp.ellipk(1-m)/sqrt_scale)


def trig_periods(g2, g3):
    return periods_from_roots(trig_roots(g2, g3))


def cardano_complex_periods(g2, g3):
    return periods_from_roots(cardano_roots(g2, g3))


def cardano_real_periods(g2, g3):
    return periods_from_roots(cardano_real_roots(g2, g3))


def bench_cases(loops, function):
    start = time.perf_counter()
    for _ in range(loops):
        for g2, g3 in CASES:
            function(g2, g3)
    return time.perf_counter() - start


for g2, g3 in CASES:
    expected = trig_periods(g2, g3)
    for function in (cardano_complex_periods, cardano_real_periods):
        actual = function(g2, g3)
        assert all(mp.almosteq(a, e) for a, e in zip(actual, expected))

runner = pyperf.Runner(metadata={'mpmath_dps': mp.dps})
for name, function in [
        ('roots: trigonometric', trig_roots),
        ('roots: Cardano complex', cardano_roots),
        ('roots: Cardano then real', cardano_real_roots),
        ('periods: trigonometric', trig_periods),
        ('periods: Cardano complex', cardano_complex_periods),
        ('periods: Cardano then real', cardano_real_periods)]:
    runner.bench_time_func(
        name, bench_cases, function, inner_loops=len(CASES))
```

skirpichev: Yes, looks better.

```
$ python a.py -q --rigorous 
roots: trigonometric: Mean +- std dev: 429 us +- 9 us
roots: Cardano complex: Mean +- std dev: 570 us +- 15 us
roots: Cardano then real: Mean +- std dev: 584 us +- 7 us
periods: trigonometric: Mean +- std dev: 647 us +- 8 us
periods: Cardano complex: Mean +- std dev: 2.29 ms +- 0.05 ms
periods: Cardano then real: Mean +- std dev: 807 us +- 12 us
```

skirpichev:
Benchmark with gmpy2:
```
omega1omega2from: real mean: Mean +- std dev: [omega-master] 5.56 ms +- 0.23 ms -> [omega-pr] 890 us +- 6 us: 6.25x faster
omega1omega2from: complex mean: Mean +- std dev: [omega-master] 5.59 ms +- 0.09 ms -> [omega-pr] 3.63 ms +- 0.03 ms: 1.54x faster

Geometric mean: 3.10x faster
```
with builtin ints:
```
omega1omega2from: real mean: Mean +- std dev: [omega-master] 6.58 ms +- 0.03 ms -> [omega-pr] 991 us +- 13 us: 6.64x faster
omega1omega2from: complex mean: Mean +- std dev: [omega-master] 6.60 ms +- 0.20 ms -> [omega-pr] 4.20 ms +- 0.02 ms: 1.57x faster

Geometric mean: 3.23x faster
```

skirpichev (approved):

Issue #929, opened 2025-03-16
Interface to clear caches


skirpichev:
Caching methods in the mpmath vary from place to place.  Some are per-module (e.g. `mpmath.libmp.gammazeta.bernoulli_cache`), some per context (e.g. `QuadratureRule`'s caches).

Some caches actually can be unbounded (see related issue #811).  Especially in such cases we need some interface (e.g. a context method like `clear_caches()`) to free memory.  More low-level caches (i.e. in `libmp`) probably shouldn't be unbounded.

This might depend on #683.  With a more "light" and fluent context objects, that can be created/destroyed on demand - freeing memory isn't a problem.

PR #1163, opened 2026-08-26
Add cache decorator and use on elliptic function parameter conversions

HeskethGD:
## Summary

This PR adds a bounded, precision-aware LRU cache decorator and uses it to optimise performance on the elliptic function parameter conversion functions.

The Weierstrass functions internally convert their lattice parameters to half-periods. When several function values are evaluated using the same invariants `(g2, g3)`, that relatively expensive conversion was previously repeated for every value of `z`. The situation is similar for the Jacobi elliptic functions when using the convenient `m` parameter and converting to `q` internally.

The new composable `ctx_lru_cache` decorator:

- Can be combined with either `defun` or `defun_wrapped`.
- Creates a separate cache for each mpmath context.
- Includes precision, rounding mode, and `trap_complex` in the cache key.
- Retains the 16 most recently used argument combinations for these functions.
- Preserves the wrapped function’s public signature.
- Exposes `cache_info()` and `cache_clear()` for inspection and control.
- Uses `functools.lru_cache` for established and predictable eviction behaviour.

`omega1omega2from`, `qfrom`, `qbarfrom`, `taufrom`, `kfrom`, and `mfrom` now use this decorator. Consequently, `weierp`, `weierpprime`, `weiersigma`, `weierzeta`, and `ellipfun` benefit automatically when a process works repeatedly with a small collection of lattices.

The existing documentation recommendation to precompute and pass `(omega1, omega2)` for repeated evaluations is removed because repeated `(g2, g3)` evaluation is now effectively at parity with the explicit-period path.

## Performance for Weierstrass elliptic functions

The benchmark compared this branch with `master` using Python 3.13.4 on Apple Silicon at 50 decimal digits. Results are medians from warmed runs over nine generic real and ten complex lattices.

### Conversion and cache overhead

| Cases | Operation | Master | PR | Change |
|---|---|---:|---:|---:|
| Real | Repeated `(g2, g3)` | 111.29 µs | 0.70 µs | -99.4% |
| Real | Repeated `(omega1, omega2)` | 6.25 µs | 1.50 µs | -76.1% |
| Real | Cycling 17 period pairs | 6.27 µs | 8.18 µs | +30.4% |
| Complex | Repeated `(g2, g3)` | 540.51 µs | 1.38 µs | -99.7% |
| Complex | Repeated `(omega1, omega2)` | 9.89 µs | 2.05 µs | -79.3% |
| Complex | Cycling 17 period pairs | 6.29 µs | 8.20 µs | +30.5% |

Cycling 17 distinct period pairs exceeds the 16-entry capacity, deliberately producing cache misses. The measured miss cost is approximately 1.9 µs per call. This is small compared with the invariant-to-period conversion avoided by a hit.

### Cache trade-off

A forced LRU miss increased the inexpensive explicit-period path from about 6.3 µs to 8.2 µs. Although this is a 30% increase, the absolute overhead is only around 1.9 µs. It is much less significant for invariant conversions, which take roughly 111–541 µs without caching, and one subsequent hit recovers more than the miss overhead.

The LRU also produced faster hits than the earlier `memoize_last` prototype. Real invariant lookups fell from 1.72 to 0.70 µs and complex lookups from 2.11 to 1.38 µs. This is likely because `functools.lru_cache` performs most lookup work in optimized C, while `memoize_last` handled key comparison and precision logic in Python.

The trade-off is therefore a small miss penalty and up to 16 cached entries per context, in exchange for faster hits and support for several interleaved lattices.

### Why not use `lru_cache` directly?

`omega1omega2from` is registered dynamically on each mpmath context, and its results depend on that context’s current precision. The wrapper creates a separate cache for each context, adds the precision to the cache key, and preserves the function’s public signature. Applying `lru_cache` directly would not provide those behaviours.

### Weierstrass function evaluation

Each call uses the same `(g2, g3)` but a different `z`.

| Function | Real master | Real PR | Change | Complex master | Complex PR | Change |
|---|---:|---:|---:|---:|---:|---:|
| `weierp` | 604.67 µs | 478.09 µs | -20.9% | 1274.84 µs | 693.66 µs | -45.6% |
| `weierpprime` | 1031.76 µs | 906.37 µs | -12.2% | 1889.47 µs | 1304.79 µs | -30.9% |
| `weiersigma` | 505.59 µs | 378.81 µs | -25.1% | 1129.66 µs | 551.15 µs | -51.2% |
| `weierzeta` | 745.48 µs | 619.43 µs | -16.9% | 1467.76 µs | 878.53 µs | -40.1% |

On the PR branch, repeated `(g2, g3)` evaluations were effectively at parity with evaluations using precomputed `(omega1, omega2)`.

### Repeated 100-point sweeps

The ranges below cover two representative real and two representative complex lattices, each evaluated at 100 different values of `z`.

| Function | Real speedup | Complex speedup |
|---|---:|---:|
| `weierp` | 1.22–1.26x | 1.76–1.79x |
| `weierpprime` | 1.13–1.14x | 1.41–1.44x |
| `weiersigma` | 1.28–1.33x | 2.01–2.02x |
| `weierzeta` | 1.18–1.20x | 1.61–1.62x |

One real case, `g2=-4` and `g3=1`, was reported separately and excluded from the generic-real averages. Its period ratio lies on the modular boundary `abs(tau)=1`, where an existing theta modular-reduction boundary effect can dominate the timing. This behaviour is present on `master` and is independent of the cache.

<details>

Save the following as `bench_weierstrass_lru.py` in the repository root and run it once on `master` and once on this branch:

```python
import statistics
import timeit

from mpmath import (
    mp,
    omega1omega2from,
    weierp,
    weierpprime,
    weiersigma,
    weierzeta,
)

mp.dps = 50

CASES = [
    ("real", mp.mpf(60), mp.mpf(140)),
    ("complex", mp.mpc(1, 2), mp.mpc(3, -4)),
]
FUNCTIONS = [weierp, weierpprime, weiersigma, weierzeta]
LRU_SIZE = 16


def median_time(callable_, number):
    callable_()  # Warm the path before timing.
    samples = timeit.repeat(callable_, number=number, repeat=5)
    return statistics.median(samples) / number


def sample_points(omega1, omega2, count=100):
    golden = (mp.sqrt(5) - 1) / 2
    return [
        2 * (
            (
                mp.mpf("0.1")
                + mp.mpf("0.8") * (k + mp.mpf("0.5")) / count
            ) * omega1
            + (
                mp.mpf("0.1")
                + mp.mpf("0.8")
                * mp.frac((k + mp.mpf("0.5")) * golden)
            ) * omega2
        )
        for k in range(count)
    ]


for label, g2, g3 in CASES:
    invariants = {"g2": g2, "g3": g3}
    omega1, omega2 = omega1omega2from(**invariants)
    periods = {"omega1": omega1, "omega2": omega2}
    points = sample_points(omega1, omega2)

    invariant_time = median_time(
        lambda: omega1omega2from(**invariants), 100
    )
    period_time = median_time(
        lambda: omega1omega2from(**periods), 100
    )
    print(
        f"{label}: conversion g2/g3={invariant_time * 1e6:.2f} us, "
        f"omega1/omega2={period_time * 1e6:.2f} us"
    )

    # Seventeen entries exceed the PR's 16-entry cache, measuring misses.
    churn_periods = [
        {
            "omega1": mp.mpf("0.75") + mp.mpf(k) / 100,
            "omega2": (mp.mpf("0.75") + mp.mpf(k) / 100)
            * mp.mpc("0.1", "1.1"),
        }
        for k in range(LRU_SIZE + 1)
    ]

    def churning_conversions():
        for churn_period in churn_periods:
            omega1omega2from(**churn_period)

    churn_time = (
        median_time(churning_conversions, 10) / len(churn_periods)
    )
    print(
        f"{label}: cycling 17 periods="
        f"{churn_time * 1e6:.2f} us/call"
    )

    for function in FUNCTIONS:
        def invariant_sweep():
            for z in points:
                function(z, **invariants)

        def period_sweep():
            for z in points:
                function(z, **periods)

        # Check that the two parameterizations agree before timing them.
        for z in points:
            assert mp.almosteq(
                function(z, **invariants),
                function(z, **periods),
                rel_eps=100 * mp.eps,
                abs_eps=100 * mp.eps,
            )

        invariant_sweep_time = median_time(invariant_sweep, 1)
        period_sweep_time = median_time(period_sweep, 1)
        print(
            f"  {function.__name__}: "
            f"g2/g3={invariant_sweep_time * 1e3:.2f} ms, "
            f"omega1/omega2={period_sweep_time * 1e3:.2f} ms"
        )
```

</details>

### Caching `g2g3from`

The bounded LRU cache is also applied to `g2g3from`, providing the same benefit for repeated conversion in the opposite direction. This is useful when the same lattice is repeatedly expressed through `tau` or explicit half-periods.

At 50 decimal digits, warmed benchmark results were:

| Input | Master | PR | Change |
|---|---:|---:|---:|
| Repeated `tau` | 284 µs | 1.15 µs | -99.6% |
| Repeated `(omega1, omega2)` | 297 µs | 1.70 µs | -99.4% |
| Repeated `(g2, g3)` | 2.4 µs | 0.75 µs | -68.8% |
| Cycling 17 `(g2, g3)` pairs | 2.4 µs | 3.34 µs | +39.2% |

The final row deliberately exceeds the 16-entry cache while exercising the cheapest conversion path. Although its relative miss penalty is significant, the absolute overhead is less than 1 µs, and one subsequent hit recovers it. Miss overhead is proportionally much smaller for the more expensive `tau` and half-period conversions.

## Elliptic converter and Jacobi function benchmarks

The cache was also applied to `qfrom`, `qbarfrom`, `taufrom`, `kfrom`, and `mfrom`. Benchmarks used 50 decimal digits, warmed callables, and the median of seven repeats.

Repeated calls use the same arguments and therefore measure cache hits. The miss benchmark cycles through 17 direct inputs, exceeding the 16-entry cache.

| Converter | Repeated direct input | Repeated `m` | Repeated `g2, g3` | Miss overhead |
|---|---:|---:|---:|---:|
| `qfrom` | 1.35 → 0.72 µs | 29.83 → 0.72 µs | 119.38 → 0.81 µs | +0.88 µs |
| `qbarfrom` | 1.34 → 0.71 µs | 31.19 → 0.70 µs | 122.04 → 0.82 µs | +0.85 µs |
| `taufrom` | 1.73 → 1.20 µs | 24.25 → 0.70 µs | 107.89 → 0.84 µs | +1.37 µs |
| `kfrom` | 1.37 → 0.71 µs | 3.82 → 0.73 µs | 233.22 → 0.83 µs | +0.85 µs |
| `mfrom` | 1.32 → 0.72 µs | 1.29 → 0.71 µs | 239.10 → 0.82 µs | +0.82 µs |

Values are shown as master → this branch. “Direct input” means passing each converter its corresponding parameter: `q`, `qbar`, `tau`, `k`, or `m`.

The miss penalty is less than 1 µs for four converters and approximately 1.4 µs for `taufrom`. Repeated expensive conversions show substantially larger gains.

### Jacobi elliptic functions

`ellipfun` converts its elliptic parameterization to `q` on every evaluation. Consequently, evaluating at many different `z` values with a fixed `m` benefits from caching the `m`-to-`q` conversion.

The following results measure a batch of 50 evaluations of `ellipfun('sn', z, ...)` with changing `z` and a fixed elliptic parameter:

| Parameterization | Master | This branch | Speedup |
|---|---:|---:|---:|
| Fixed `m` | 9.72 ms | 7.86 ms | 1.24× |
| Fixed `tau` | 10.03 ms | 9.66 ms | 1.04× |
| Fixed `q` | 7.95 ms | 7.95 ms | 1.00× |

Caching brings the convenient `m` parameterization approximately to the performance of supplying a precomputed `q`. Direct `q` evaluation is unchanged, as expected.

<details>
<summary>Benchmark script</summary>

Run this script separately from the master checkout and this branch.

```python
import statistics
import subprocess
import timeit

from mpmath import ellipfun, mp


mp.dps = 50
REPEATS = 7
FUNCTIONS = ("qfrom", "qbarfrom", "taufrom", "kfrom", "mfrom")


def median_time(func, number):
    func()
    samples = timeit.repeat(func, number=number, repeat=REPEATS)
    return statistics.median(samples) / number


direct_arguments = {
    "qfrom": {"q": mp.mpf("0.2")},
    "qbarfrom": {"qbar": mp.mpf("0.04")},
    "taufrom": {"tau": mp.mpc("0.2", "1.1")},
    "kfrom": {"k": mp.mpf("0.4")},
    "mfrom": {"m": mp.mpf("0.16")},
}

common_m = {"m": mp.mpf("0.3")}
invariants = {"g2": mp.mpf(60), "g3": mp.mpf(140)}

    ["git", "rev-parse", "--short", "HEAD"], text=True
).strip()
print("Revision:", revision)
print("Converter timings (us/call)")
print("function        direct          m       g2,g3      miss")

for name in FUNCTIONS:
    function = getattr(mp, name)

    direct = median_time(
        lambda f=function, a=direct_arguments[name]: f(**a), 1000
    )
    from_m = median_time(lambda f=function: f(**common_m), 300)
    from_invariants = median_time(
        lambda f=function: f(**invariants), 200
    )

    changing_arguments = []
    for index in range(17):
        if name == "qfrom":
            arguments = {
                "q": mp.mpf("0.1") + index / mp.mpf(1000)
            }
        elif name == "qbarfrom":
            arguments = {
                "qbar": mp.mpf("0.01") + index / mp.mpf(10000)
            }
        elif name == "taufrom":
            arguments = {
                "tau": mp.mpc(index / mp.mpf(100), "1.1")
            }
        elif name == "kfrom":
            arguments = {
                "k": mp.mpf("0.2") + index / mp.mpf(1000)
            }
        else:
            arguments = {
                "m": mp.mpf("0.2") + index / mp.mpf(1000)
            }
        changing_arguments.append(arguments)

    position = [-1]

    def cache_miss():
        position[0] = (position[0] + 1) % len(changing_arguments)
        return function(**changing_arguments[position[0]])

    for _ in changing_arguments:
        cache_miss()

    miss = median_time(cache_miss, 1700)

    print(
        f"{name:<10}"
        f"{direct * 1e6:10.2f}"
        f"{from_m * 1e6:11.2f}"
        f"{from_invariants * 1e6:12.2f}"
        f"{miss * 1e6:10.2f}"
    )


m = mp.mpf("0.3")
tau = mp.taufrom(m=m)
q = mp.qfrom(m=m)
points = [
    mp.mpf("0.1") + index / mp.mpf(50)
    for index in range(50)
]


def jacobi_batch(parameters):
    result = None
    for z in points:
        result = ellipfun("sn", z, **parameters)
    return result


print("\nJacobi sn timings (ms per 50-z batch)")
for name, parameters in (
    ("m", {"m": m}),
    ("tau", {"tau": tau}),
    ("q", {"q": q}),
):
    elapsed = median_time(
        lambda p=parameters: jacobi_batch(p), 5
    )
    print(f"{name:<4}{elapsed * 1e3:10.2f}")
```

</details>

## Tests

Tests cover:

- Cache hits for identical arguments.
- LRU retention and eviction.
- The exact 16-entry capacity.
- Precision as part of the cache key.
- Reuse when returning to an earlier precision while its entry remains cached.
- Separation between mpmath contexts.
- Preservation of the wrapped function signature.

The targeted function and elliptic tests pass:

```text
134 passed
```

## AI use declaration

OpenAI Codex was used to assist with implementation, test development, benchmark design and execution, and drafting this PR description. I reviewed the code, tests, benchmark methodology, and reported results.

### DIFF
--- a/mpmath/ctx_base.py
+++ b/mpmath/ctx_base.py
@@ -1,3 +1,4 @@
+import inspect
 import random
 from operator import gt, lt
 from typing import Any
@@ -568,4 +569,8 @@ class StandardBaseContext(Context,
             return value
         f_cached.__name__ = f.__name__
         f_cached.__doc__ = f.__doc__
+        try:
+            f_cached.__signature__ = inspect.signature(f)
+        except (TypeError, ValueError):  # pragma: no cover
+            pass
         return f_cached
--- a/mpmath/ctx_base.py
+++ b/mpmath/ctx_base.py
@@ -1,5 +1,6 @@
 import random
 from operator import gt, lt
+from typing import Any
 
 from . import libmp
 from .calculus.calculus import CalculusMethods
@@ -517,6 +518,53 @@ class StandardBaseContext(Context,
                     return +cvalue
             value = f(*args, **kwargs)
             f_cache[key] = (prec, value)
+            return value
+        f_cached.__name__ = f.__name__
+        f_cached.__doc__ = f.__doc__
+        return f_cached
+
+    def memoize_last(ctx, f):
+        """
+        Return a wrapped copy of *f* that caches only its most recently
+        computed value. Values are reused only if the cached precision is
+        equal to or higher than the working precision.
+
+        Unlike :meth:`memoize`, which retains a value for every distinct set
+        of arguments, this method replaces the cached value whenever the
+        arguments change. It is useful when repeated consecutive calls are
+        expected but an unbounded cache is undesirable::
+
+            >>> from mpmath import maxcalls, memoize_last, mp, sin
+            >>> mp.pretty = True
+            >>> f = memoize_last(maxcalls(sin, 2))
+            >>> f(2)
+            0.909297426825682
+            >>> f(2)
+            0.909297426825682
+            >>> f(3)
+            0.141120008059867
+            >>> f(2)
+            Traceback (most recent call last):
+              ...
+            NoConvergence: maxcalls: function evaluated 2 times
+        """
+        last: tuple[Any, int, Any] | None = None
+
+        def f_cached(*args, **kwargs):
+            nonlocal last
+            if kwargs:
+                key = args, tuple(kwargs.items())
+            else:
+                key = args
+            prec = ctx.prec
+            if last is not None:
+                ckey, cprec, cvalue = last
+                if key == ckey and cprec >= prec:
+                    if type(cvalue) is tuple:
+                        return tuple(+value for value in cvalue)
+                    return +cvalue
+            value = f(*args, **kwargs)
+            last = key, prec, value
             return value
         f_cached.__name__ = f.__name__
         f_cached.__doc__ = f.__doc__
--- a/mpmath/ctx_base.py
+++ b/mpmath/ctx_base.py
@@ -1,7 +1,5 @@
-import inspect
 import random
 from operator import gt, lt
-from typing import Any
 
 from . import libmp
 from .calculus.calculus import CalculusMethods
@@ -522,55 +520,4 @@ class StandardBaseContext(Context,
             return value
         f_cached.__name__ = f.__name__
         f_cached.__doc__ = f.__doc__
-        return f_cached
-
-    def memoize_last(ctx, f):
-        """
-        Return a wrapped copy of *f* that caches only its most recently
-        computed value. Values are reused only if the cached precision is
-        equal to or higher than the working precision.
-
-        Unlike :meth:`memoize`, which retains a value for every distinct set
-        of arguments, this method replaces the cached value whenever the
-        arguments change. It is useful when repeated consecutive calls are
-        expected but an unbounded cache is undesirable::
-
-            >>> from mpmath import maxcalls, memoize_last, mp, sin
-            >>> mp.pretty = True
-            >>> f = memoize_last(maxcalls(sin, 2))
-            >>> f(2)
-            0.909297426825682
-            >>> f(2)
-            0.909297426825682
-            >>> f(3)
-            0.141120008059867
-            >>> f(2)
-            Traceback (most recent call last):
-              ...
-            NoConvergence: maxcalls: function evaluated 2 times
-        """
-        last: tuple[Any, int, Any] | None = None
-
-        def f_cached(*args, **kwargs):
-            nonlocal last
-            if kwargs:
-                key = args, tuple(kwargs.items())
-            else:
-                key = args
-            prec = ctx.prec
-            if last is not None:
-                ckey, cprec, cvalue = last
-                if key == ckey and cprec >= prec:
-                    if type(cvalue) is tuple:
-                        return tuple(+value for value in cvalue)
-                    return +cvalue
-            value = f(*args, **kwargs)
-            last = key, prec, value
-            return value
-        f_cached.__name__ = f.__name__
-        f_cached.__doc__ = f.__doc__
-        try:
-            f_cached.__signature__ = inspect.signature(f)
-        except (TypeError, ValueError):  # pragma: no cover
-            pass
         return f_cached
--- a/mpmath/functions/elliptic.py
+++ b/mpmath/functions/elliptic.py
@@ -142,6 +142,7 @@ def _validate_weierstrass_parameter_args(
                          "parameterization" % funcname)
 
 @defun_wrapped
+@ctx_lru_cache(maxsize=16)
 def qfrom(ctx, q=None, m=None, k=None, tau=None, qbar=None,
           g2=None, g3=None, omega1=None, omega2=None):
     r"""
@@ -183,6 +184,7 @@ def qfrom(ctx, q=None, m=None, k=None, tau=None, qbar=None,
         return ctx.qfrom(tau=ctx.taufrom(omega1=omega1, omega2=omega2))
 
 @defun_wrapped
+@ctx_lru_cache(maxsize=16)
 def qbarfrom(ctx, q=None, m=None, k=None, tau=None, qbar=None,
              g2=None, g3=None, omega1=None, omega2=None):
     r"""
@@ -225,6 +227,7 @@ def qbarfrom(ctx, q=None, m=None, k=None, tau=None, qbar=None,
         return ctx.qbarfrom(tau=ctx.taufrom(omega1=omega1, omega2=omega2))
 
 @defun_wrapped
+@ctx_lru_cache(maxsize=16)
 def taufrom(ctx, q=None, m=None, k=None, tau=None, qbar=None,
             g2=None, g3=None, omega1=None, omega2=None):
     r"""
@@ -309,6 +312,7 @@ def taufrom(ctx, q=None, m=None, k=None, tau=None, qbar=None,
         return tau
 
 @defun_wrapped
+@ctx_lru_cache(maxsize=16)
 def kfrom(ctx, q=None, m=None, k=None, tau=None, qbar=None,
           g2=None, g3=None, omega1=None, omega2=None):
     r"""
@@ -365,6 +369,7 @@ def kfrom(ctx, q=None, m=None, k=None, tau=None, qbar=None,
     return (ctx.jtheta(2,0,q)/ctx.jtheta(3,0,q))**2
 
 @defun_wrapped
+@ctx_lru_cache(maxsize=16)
 def mfrom(ctx, q=None, m=None, k=None, tau=None, qbar=None,
           g2=None, g3=None, omega1=None, omega2=None):
     r"""
--- a/mpmath/functions/elliptic.py
+++ b/mpmath/functions/elliptic.py
@@ -68,7 +68,7 @@ applicable (:func:`~mpmath.qfrom`, :func:`~mpmath.mfrom`,
 
 """
 
-from .functions import defun, defun_wrapped
+from .functions import defun, defun_memoized_last, defun_wrapped
 
 @defun_wrapped
 def eta(ctx, tau):
@@ -1838,7 +1838,7 @@ def g2g3from(ctx, q=None, m=None, k=None, tau=None, qbar=None,
                    j3**12))
     return +g2, +g3
 
-@defun
+@defun_memoized_last
 def omega1omega2from(ctx, q=None, m=None, k=None, tau=None, qbar=None,
                      g2=None, g3=None, omega1=None, omega2=None):
     r"""
@@ -2049,10 +2049,6 @@ def weierp(ctx, z, g2=None, g3=None, tau=None, omega1=None, omega2=None):
 
     The periods of `\wp` are `2\omega_1` and `2\omega_2`. Thus the
     `\tau` parameterization corresponds to periods `1` and `\tau`.
-
-    For repeated evaluation with the same invariants, it is faster to compute
-    the half-periods once with :func:`~mpmath.omega1omega2from` and pass
-    them using the ``omega1`` and ``omega2`` keywords.
 
     **Examples**
 
--- a/mpmath/functions/elliptic.py
+++ b/mpmath/functions/elliptic.py
@@ -68,7 +68,7 @@ applicable (:func:`~mpmath.qfrom`, :func:`~mpmath.mfrom`,
 
 """
 
-from .functions import defun, defun_lru_cache, defun_wrapped
+from .functions import ctx_lru_cache, defun, defun_wrapped
 
 @defun_wrapped
 def eta(ctx, tau):
@@ -1782,7 +1782,8 @@ def _eisenstein_G4_G6(ctx, tau):
 # Weierstrass parameter conversion functions
 # ============================================================================
 
-@defun_lru_cache(maxsize=16)
+@defun
+@ctx_lru_cache(maxsize=16)
 def g2g3from(ctx, q=None, m=None, k=None, tau=None, qbar=None,
              g2=None, g3=None, omega1=None, omega2=None):
     r"""
@@ -1838,7 +1839,8 @@ def g2g3from(ctx, q=None, m=None, k=None, tau=None, qbar=None,
                    j3**12))
     return +g2, +g3
 
-@defun_lru_cache(maxsize=16)
+@defun
+@ctx_lru_cache(maxsize=16)
 def omega1omega2from(ctx, q=None, m=None, k=None, tau=None, qbar=None,
                      g2=None, g3=None, omega1=None, omega2=None):
     r"""
--- a/mpmath/functions/elliptic.py
+++ b/mpmath/functions/elliptic.py
@@ -68,7 +68,7 @@ applicable (:func:`~mpmath.qfrom`, :func:`~mpmath.mfrom`,
 
 """
 
-from .functions import defun, defun_memoized_last, defun_wrapped
+from .functions import defun, defun_lru_cache, defun_wrapped
 
 @defun_wrapped
 def eta(ctx, tau):
@@ -1838,7 +1838,7 @@ def g2g3from(ctx, q=None, m=None, k=None, tau=None, qbar=None,
                    j3**12))
     return +g2, +g3
 
-@defun_memoized_last
+@defun_lru_cache(maxsize=16)
 def omega1omega2from(ctx, q=None, m=None, k=None, tau=None, qbar=None,
                      g2=None, g3=None, omega1=None, omega2=None):
     r"""
--- a/mpmath/functions/elliptic.py
+++ b/mpmath/functions/elliptic.py
@@ -1782,7 +1782,7 @@ def _eisenstein_G4_G6(ctx, tau):
 # Weierstrass parameter conversion functions
 # ============================================================================
 
-@defun
+@defun_lru_cache(maxsize=16)
 def g2g3from(ctx, q=None, m=None, k=None, tau=None, qbar=None,
              g2=None, g3=None, omega1=None, omega2=None):
     r"""
--- a/mpmath/functions/functions.py
+++ b/mpmath/functions/functions.py
@@ -7,12 +7,15 @@ class SpecialFunctions:
     "builtins" or "low-level" functions.
     """
     defined_functions = {}
+    memoized_last_functions = set()
 
     def __init__(self):
         cls = self.__class__
         for name in cls.defined_functions:
             f, wrap = cls.defined_functions[name]
             cls._wrap_specfun(name, f, wrap)
+            if name in cls.memoized_last_functions:
+                setattr(self, name, self.memoize_last(getattr(self, name)))
 
         self._misc_const_cache = {}
 
@@ -59,6 +62,11 @@ def defun_wrapped(f):
 
 def defun(f):
     SpecialFunctions.defined_functions[f.__name__] = f, False
+    return f
+
+def defun_memoized_last(f):
+    SpecialFunctions.defined_functions[f.__name__] = f, False
+    SpecialFunctions.memoized_last_functions.add(f.__name__)
     return f
 
 def defun_static(f):
--- a/mpmath/functions/functions.py
+++ b/mpmath/functions/functions.py
@@ -1,34 +1,31 @@
 from functools import lru_cache, wraps
 from inspect import signature
-from typing import Any, cast
 
 
 def _ctx_lru_cache(ctx, f, maxsize):
     """Cache *f* by its arguments and the context's numerical state."""
-    has_rounding = hasattr(ctx, 'rounding')
-    has_trap_complex = hasattr(ctx, 'trap_complex')
-
     @lru_cache(maxsize=maxsize)
     def cached(_ctx_state, /, *args, **kwargs):
         # lru_cache includes the context state in its key, preventing a value
         # computed with one numerical setting from being reused with another.
         # The calculation reads the same state from ctx, so it need not
         # otherwise use this argument.
-        _ = _ctx_state
         return f(*args, **kwargs)
 
     @wraps(f)
     def f_cached(*args, **kwargs):
         # Fixed-precision and interval contexts do not expose rounding or
         # trap_complex, so use a stable placeholder for those contexts.
-        rounding = ctx.rounding if has_rounding else None
-        trap_complex = ctx.trap_complex if has_trap_complex else None
+        rounding = getattr(ctx, 'rounding', None)
+        trap_complex = getattr(ctx, 'trap_complex', None)
         ctx_state = (ctx.prec, rounding, trap_complex)
         return cached(ctx_state, *args, **kwargs)
 
     # This wrapper is stored directly on a context instance, so wraps alone
     # would expose the unbound function's leading ctx parameter.
-    cast(Any, f_cached).__signature__ = signature(f)
+    f_cached.__signature__ = signature(f)
+    f_cached.cache_info = cached.cache_info
+    f_cached.cache_clear = cached.cache_clear
     return f_cached
 
 
@@ -100,10 +97,9 @@ def defun(f):
     SpecialFunctions.defined_functions[f.__name__] = f, False
     return f
 
-def defun_lru_cache(maxsize):
-    """Register a function with a per-context LRU cache."""
+def ctx_lru_cache(maxsize):
+    """Mark a context function for per-context LRU caching."""
     def decorator(f):
-        SpecialFunctions.defined_functions[f.__name__] = f, False
         SpecialFunctions.lru_cache_functions[f.__name__] = maxsize
         return f
     return decorator
--- a/mpmath/functions/functions.py
+++ b/mpmath/functions/functions.py
@@ -1,3 +1,28 @@
+from functools import lru_cache, wraps
+from inspect import signature
+from typing import Any, cast
+
+
+def _ctx_lru_cache(ctx, f, maxsize):
+    """Cache *f* by its arguments and the context's current precision."""
+    @lru_cache(maxsize=maxsize)
+    def cached(_prec, /, *args, **kwargs):
+        # lru_cache includes _prec in its key, preventing a value computed at
+        # one precision from being reused at another. The calculation itself
+        # reads the same precision from ctx, so it need not otherwise use it.
+        _ = _prec
+        return f(*args, **kwargs)
+
+    @wraps(f)
+    def f_cached(*args, **kwargs):
+        return cached(ctx.prec, *args, **kwargs)
+
+    # This wrapper is stored directly on a context instance, so wraps alone
+    # would expose the unbound function's leading ctx parameter.
+    cast(Any, f_cached).__signature__ = signature(f)
+    return f_cached
+
+
 class SpecialFunctions:
     """
     This class implements special functions using high-level code.
@@ -7,15 +32,17 @@ class SpecialFunctions:
     "builtins" or "low-level" functions.
     """
     defined_functions = {}
-    memoized_last_functions = set()
+    lru_cache_functions = {}
 
     def __init__(self):
         cls = self.__class__
         for name in cls.defined_functions:
             f, wrap = cls.defined_functions[name]
             cls._wrap_specfun(name, f, wrap)
-            if name in cls.memoized_last_functions:
-                setattr(self, name, self.memoize_last(getattr(self, name)))
+            if name in cls.lru_cache_functions:
+                maxsize = cls.lru_cache_functions[name]
+                setattr(self, name, _ctx_lru_cache(
+                    self, getattr(self, name), maxsize))
 
         self._misc_const_cache = {}
 
@@ -64,10 +91,13 @@ def defun(f):
     SpecialFunctions.defined_functions[f.__name__] = f, False
     return f
 
-def defun_memoized_last(f):
-    SpecialFunctions.defined_functions[f.__name__] = f, False
-    SpecialFunctions.memoized_last_functions.add(f.__name__)
-    return f
+def defun_lru_cache(maxsize):
+    """Register a function with a per-context LRU cache."""
+    def decorator(f):
+        SpecialFunctions.defined_functions[f.__name__] = f, False
+        SpecialFunctions.lru_cache_functions[f.__name__] = maxsize
+        return f
+    return decorator
 
 def defun_static(f):
     setattr(SpecialFunctions, f.__name__, f)
--- a/mpmath/functions/functions.py
+++ b/mpmath/functions/functions.py
@@ -4,18 +4,27 @@ from typing import Any, cast
 
 
 def _ctx_lru_cache(ctx, f, maxsize):
-    """Cache *f* by its arguments and the context's current precision."""
+    """Cache *f* by its arguments and the context's numerical state."""
+    has_rounding = hasattr(ctx, 'rounding')
+    has_trap_complex = hasattr(ctx, 'trap_complex')
+
     @lru_cache(maxsize=maxsize)
-    def cached(_prec, /, *args, **kwargs):
-        # lru_cache includes _prec in its key, preventing a value computed at
-        # one precision from being reused at another. The calculation itself
-        # reads the same precision from ctx, so it need not otherwise use it.
-        _ = _prec
+    def cached(_ctx_state, /, *args, **kwargs):
+        # lru_cache includes the context state in its key, preventing a value
+        # computed with one numerical setting from being reused with another.
+        # The calculation reads the same state from ctx, so it need not
+        # otherwise use this argument.
+        _ = _ctx_state
         return f(*args, **kwargs)
 
     @wraps(f)
     def f_cached(*args, **kwargs):
-        return cached(ctx.prec, *args, **kwargs)
+        # Fixed-precision and interval contexts do not expose rounding or
+        # trap_complex, so use a stable placeholder for those contexts.
+        rounding = ctx.rounding if has_rounding else None
+        trap_complex = ctx.trap_complex if has_trap_complex else None
+        ctx_state = (ctx.prec, rounding, trap_complex)
+        return cached(ctx_state, *args, **kwargs)
 
     # This wrapper is stored directly on a context instance, so wraps alone
     # would expose the unbound function's leading ctx parameter.
--- a/mpmath/functions/functions.py
+++ b/mpmath/functions/functions.py
@@ -18,7 +18,7 @@ def _ctx_lru_cache(ctx, f, maxsize):
         # trap_complex, so use a stable placeholder for those contexts.
         rounding = getattr(ctx, 'rounding', None)
         trap_complex = getattr(ctx, 'trap_complex', None)
-        ctx_state = (ctx.prec, rounding, trap_complex)
+        ctx_state = ctx.prec, rounding, trap_complex
         return cached(ctx_state, *args, **kwargs)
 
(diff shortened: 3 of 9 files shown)


mpmath/functions/elliptic.py
```diff
@@ -2050,10 +2050,6 @@ def weierp(ctx, z, g2=None, g3=None, tau=None, omega1=None, omega2=None):
     The periods of `\wp` are `2\omega_1` and `2\omega_2`. Thus the
     `\tau` parameterization corresponds to periods `1` and `\tau`.
 
-    For repeated evaluation with the same invariants, it is faster to compute
-    the half-periods once with :func:`~mpmath.omega1omega2from` and pass
-    them using the ``omega1`` and ``omega2`` keywords.
```
skirpichev, line 2055: This part does make sense for me.  You can instead mention above *preferred* way of parameterization.

skirpichev (changes requested):
> targets the common pattern of many evaluations with the same invariants

Could you elaborate this a bit more?

What's wrong with some bookkeeping of invariants, if they will be required later?

> but is unsuitable here because users may evaluate arbitrary invariants over the lifetime of a process

That's ok, not all functions in the mpmath are using some caching mechanism.

mpmath/functions/elliptic.py
HeskethGD: I removed this paragraph because it describes a performance workaround that this change makes unnecessary. After the first call, repeated evaluations with the same `g2, g3` reuse the converted half-periods, so users no longer need to perform that conversion manually, therefore neither of those listed methods is faster than the other. If on the other hand `g2,g3` are changing, the user would always need to calculate `omega1, omega2` after every change anyway whether it is done inside or outside the elliptic functions
Therefore, I do not think there is otherwise a universally preferred parameterisation to document as `g2, g3`, `omega1, omega2`, and `tau` represent different forms in which the caller may naturally have the lattice. The preceding paragraph already documents the accepted forms.

HeskethGD:
> > targets the common pattern of many evaluations with the same invariants
> 
> Could you elaborate this a bit more?
> 
> What's wrong with some bookkeeping of invariants, if they will be required later?

The case I have in mind and use a lot is identifying `g2,g3` from some integrable system, then plotting the $\wp, \zeta, \sigma$ over some  range of `z` values, often multiplying several of those functions together. While I could have done that by calculating `omega1, omega2` once first and reusing the calculated values, it presents a bit of an issue when I start to think about how these functions might be implemented in SymPy. There I'd likely only provide one parameter option and I'd prefer to use `g2, g3` because they uniquely parameterise the functions without any basis choice ambiguity, and because `g2,g3` are algebraic functions of system parameters whereas `omega1, omega2` are transcendental functions of system parameters. It would be good to implement `g2,g3` in Sympy without thinking that the choice was numerically suboptimal.

> > but is unsuitable here because users may evaluate arbitrary invariants over the lifetime of a process
> 
> That's ok, not all functions in the mpmath are using some caching mechanism.

Irrespective of whether `memoize` or `memoize_last` is used, the key observation I found was that implementing either as direct functions was quite verbose and that it required additional cache management logic. I found this because I initially did not use a decorator approach in this [now closed PR](https://github.com/mpmath/mpmath/pull/1162) and ended up with code like:
```
def _omega1omega2from(ctx, q=None, m=None, k=None, tau=None, qbar=None,
                      g2=None, g3=None, omega1=None, omega2=None):
                      #... the main function logic

@defun
def omega1omega2from(ctx, q=None, m=None, k=None, tau=None, qbar=None,
                     g2=None, g3=None, omega1=None, omega2=None):
    try:
        cached = ctx._omega1omega2from_cached
    except AttributeError:
        def compute(q=None, m=None, k=None, tau=None, qbar=None,
                    g2=None, g3=None, omega1=None, omega2=None):
            return _omega1omega2from(
                ctx, q, m, k, tau, qbar, g2, g3, omega1, omega2)
        cached = ctx.memoize_last(compute)
        ctx._omega1omega2from_cached = cached
    return cached(q, m, k, tau, qbar, g2, g3, omega1, omega2)


omega1omega2from.__doc__ = _omega1omega2from.__doc__
```
but with this reusable decorator approach in this current PR there is just a one line change to add last-value cache:
```
@defun_memoized_last
def omega1omega2from(ctx, q=None, m=None, k=None, tau=None, qbar=None,
                     g2=None, g3=None, omega1=None, omega2=None):
                     #... the main function logic
```
The decorator method could be used in a few other places in this file to provide speed up, repeating the cache overhead logic everywhere would cumbersome.

While `memoize` could have been used in the decorator, it did not seem required to provide unlimited caching for the use cases I was identifying and simply replacing the tuple was efficient and simple to implement.

mpmath/functions/elliptic.py
skirpichev: > I removed this paragraph because it describes a performance workaround that this change makes unnecessary.

I understand.  But it looks redundant in the master as well.

It describe obvious optimization: save computed values, if you need them again.  Non-trivial part is which parametrization doesn't involve any computations in code, handling parameters.  This might be viewed as an implementation detail.  In that case above paragraph should be just removed.  Or preferred parametrization should be documented.

mpmath/ctx_base.py
```diff
@@ -521,3 +522,50 @@ def f_cached(*args, **kwargs):
...
+            >>> f(3)
+            0.141120008059867
+            >>> f(2)
+            Traceback (most recent call last):
+              ...
+            NoConvergence: maxcalls: function evaluated 2 times
+        """
+        last: tuple[Any, int, Any] | None = None
+
+        def f_cached(*args, **kwargs):
```
skirpichev: That will break signature of the decorated function.  You can fix this, using ``__signature__`` attribute.

skirpichev (commented):
> The case I have in mind and use a lot is identifying g2,g3 from some integrable system, then plotting the ℘ , ζ , σ over some range of z values, often multiplying several of those functions together.

I don't see any issue with this, if you are using mpmath's functions directly.

For SymPy, assuming above workload, you can cache preferred parametrization for numerics and use one in the `_eval_evalf()` helpers.

> this current PR there is just a one line change to add last-value cache

Can you try lru_cache with reasonable limits?  This also address example case and a bit more.  Other functions for parametrization change also can be decorated.

I suspect it might be better just because of C accelerator for this decorator.

mpmath/functions/elliptic.py
HeskethGD: If we are in agreement I will leave it deleted here.

mpmath/ctx_base.py
HeskethGD: Fixed in a follow up commit. I have not squashed the commits while the review is ongoing but would do at the end.

HeskethGD:
> > The case I have in mind and use a lot is identifying g2,g3 from some integrable system, then plotting the ℘ , ζ , σ over some range of z values, often multiplying several of those functions together.
> 
> I don't see any issue with this, if you are using mpmath's functions directly.

Even independently of SymPy, allowing these special functions to be called directly with the available invariants provides a simpler interface. Requiring users to precompute and retain the periods introduces an additional step, assumes prior knowledge of the performance implications, and creates opportunities for incorrect or stale values, particularly in interactive notebooks. Since the repeated conversion can be avoided internally with a small, bounded cache, users need not manage this optimization themselves.

> For SymPy, assuming above workload, you can cache preferred parametrization for numerics and use one in the `_eval_evalf()` helpers.
> 

It seems unnecessary to move this numerical optimization downstream when it can be handled simply within mpmath. Implementing it in SymPy’s `_eval_evalf()` methods would require coordinating cached periods across separate evaluations and Weierstrass functions while also accounting for changes in precision. Keeping the optimization in `omega1omega2from` avoids duplicating mpmath’s parameter-conversion logic and benefits both direct mpmath users and downstream callers.

> > this current PR there is just a one line change to add last-value cache
> 
> Can you try lru_cache with reasonable limits? This also address example case and a bit more. Other functions for parametrization change also can be decorated.
> 
> I suspect it might be better just because of C accelerator for this decorator.

I explored the `lru_cache` approach and created a draft PR for review https://github.com/mpmath/mpmath/pull/1165. Unlike the existing uses of `lru_cache` in mpmath, it cannot be applied directly as a decorator here because the result depends on the context’s working precision which can change. The implementation therefore uses a small wrapper that maintains a separate cache per context, includes the precision in the cache key, and preserves the public function signature.

The C-accelerated LRU provides somewhat faster cache hits than the last-value cache. A forced miss increased the inexpensive explicit-period path from approximately 6.3 µs to 8.2 µs because it adds key construction, lookup, insertion, and possible eviction around the original calculation.

If the `lru_cache` version is preferred I'd close this in favour of that.

mpmath/ctx_base.py
```diff
@@ -521,3 +523,54 @@ def f_cached(*args, **kwargs):
...
+                    return +cvalue
+            value = f(*args, **kwargs)
+            last = key, prec, value
+            return value
+        f_cached.__name__ = f.__name__
+        f_cached.__doc__ = f.__doc__
+        try:
+            f_cached.__signature__ = inspect.signature(f)
+        except (TypeError, ValueError):  # pragma: no cover
+            pass
```
skirpichev: Do we have any example, that might trigger this?  Signatures should be available for pure-Python code.

mpmath/ctx_base.py
```diff
@@ -521,3 +523,54 @@ def f_cached(*args, **kwargs):
...
+        def f_cached(*args, **kwargs):
+            nonlocal last
+            if kwargs:
+                key = args, tuple(kwargs.items())
+            else:
+                key = args
+            prec = ctx.prec
+            if last is not None:
+                ckey, cprec, cvalue = last
+                if key == ckey and cprec >= prec:
```
skirpichev: I suspect this is more tricky.  Contexts have also other options, that affect evaluation: rounding and trap_complex.

E.g., currently this suffers from double rounding, like:
```pycon
>>> mp.prec = 64
>>> x = rand()
>>> x
0.690452739991698760777
>>> mp.prec = 53
>>> mpf(x._mpf_, rounding='n')
0.69045273999169876
>>> mpf(x._mpf_, rounding='d')
0.69045273999169876
>>> mpf(x._mpf_, rounding='u')  # XXX
0.69045273999169887
```

BTW, this decorator also assumes results are correctly rounded.  That's [not true](https://mpmath.readthedocs.io/en/latest/technical.html#correctness-guarantees) for most mpmath's functions, in general.  Decorated function will not be a pure anymore.

skirpichev:
> Even independently of SymPy, allowing these special functions to be called directly with the available invariants provides a simpler interface.

Not necessary.  People might not expect this optimization at all.

While using well known language capabilities - is something much more natural.

> I explored the lru_cache approach and created a draft PR for review https://github.com/mpmath/mpmath/pull/1165. Unlike the existing uses of lru_cache in mpmath, it cannot be applied directly as a decorator here because the result depends on the context’s working precision which can change.

Yes, I think such bounded cache is better than just caching the last value.  I'm convinced that using small sets of parameter values is something typical for elliptic functions.

> If the lru_cache version is preferred I'd close this in favour of that.

I would prefer if you change this pr.

HeskethGD:
The two CLI failures appear unrelated to this PR. They occur during IPython startup, before any mpmath calculation, because recent IPython versions no longer expose `IPython.terminal` as a side effect of `import IPython`.

mpmath/ctx_base.py
HeskethGD: This is no longer in the code.

HeskethGD:
> Yes, I think such bounded cache is better than just caching the last value. I'm convinced that using small sets of parameter values is something typical for elliptic functions.
> 
> > If the lru_cache version is preferred I'd close this in favour of that.
> 
> I would prefer if you change this pr.

I have changed this PR to use the `lru_cache` approach and updated the description and title. I have set it to maxsize=16 currently for the `omega1, omega2` to `g2, g3` conversion.

mpmath/ctx_base.py
HeskethGD: While the comment was for `memoize_last` it still applies to the `lru_cache` approach and so I have fixed it in a dedicated commit on this PR so that cache keys account for `rounding` and `trap_complex`.

mpmath/tests/test_elliptic.py
```diff
@@ -11,7 +11,9 @@
 
 """
 
+import inspect
 import random
+from typing import Any
```
skirpichev: Why you need typing support?

mpmath/tests/test_functions.py
```diff
@@ -15,6 +16,7 @@
                     nthroot, phi, pi, power, powm1, radians, rand, re, root,
                     sec, sech, sign, sin, sinc, sincpi, sinh, sinpi, sqrt, tan,
                     tanh, twinprime, unitroots)
+from mpmath.functions.functions import _ctx_lru_cache
```
skirpichev: Could you test instead the defun_lru_cache decorator?

skirpichev (commented):
> I have changed this PR to use the lru_cache approach and updated the description and title. I have set it to maxsize=16 currently for the omega1, omega2 to g2, g3 conversion.

Thanks.  Perhaps, this make sense for other parameter conversion functions?

> The two CLI failures appear unrelated to this PR.

Yes, I pinned IPython.  Looks as a bug for me.

mpmath/tests/test_elliptic.py
HeskethGD: I was trying to fix some pylance issue that did not recognise functions on ctx. Admittedly, this is not something that is done consistently and `Any` typing does not add anything beyond making a red squiggle go away so this has been removed.

mpmath/tests/test_functions.py
HeskethGD: This has been done in a new commit.

HeskethGD:
> > I have changed this PR to use the lru_cache approach and updated the description and title. I have set it to maxsize=16 currently for the omega1, omega2 to g2, g3 conversion.
> 
> Thanks. Perhaps, this make sense for other parameter conversion functions?

Yes, the intention was always to use it in several other places but to treat this instance as the canary for the approach. I have now added it to the `g2g3from` method too and updated description and title to record the results. 

There would be a bit more work to do to make it a drop in replacement for the existing `foofrom` methods because they are using `defun_wrapped` and that decorator is doing a bit more so we'd likely need to modify one of the two decorators to combine roles. Something to look at in a new PR probably.

mpmath/functions/functions.py
```diff
@@ -1,3 +1,37 @@
...
+    has_rounding = hasattr(ctx, 'rounding')
+    has_trap_complex = hasattr(ctx, 'trap_complex')
+
+    @lru_cache(maxsize=maxsize)
+    def cached(_ctx_state, /, *args, **kwargs):
+        # lru_cache includes the context state in its key, preventing a value
+        # computed with one numerical setting from being reused with another.
+        # The calculation reads the same state from ctx, so it need not
+        # otherwise use this argument.
+        _ = _ctx_state
```
skirpichev:
```suggestion
```

mpmath/functions/functions.py
```diff
@@ -1,3 +1,37 @@
...
+        return f(*args, **kwargs)
+
+    @wraps(f)
+    def f_cached(*args, **kwargs):
+        # Fixed-precision and interval contexts do not expose rounding or
+        # trap_complex, so use a stable placeholder for those contexts.
+        rounding = ctx.rounding if has_rounding else None
+        trap_complex = ctx.trap_complex if has_trap_complex else None
```
skirpichev: Is this really faster than `getattr(ctx, 'foo', None)`?

mpmath/functions/functions.py
```diff
@@ -1,3 +1,37 @@
...
+        ctx_state = (ctx.prec, rounding, trap_complex)
+        return cached(ctx_state, *args, **kwargs)
+
+    # This wrapper is stored directly on a context instance, so wraps alone
+    # would expose the unbound function's leading ctx parameter.
+    cast(Any, f_cached).__signature__ = signature(f)
```
skirpichev:
```suggestion
    f_cached.__signature__ = signature(f)
```
what wrong with this?

mpmath/functions/functions.py
```diff
@@ -1,3 +1,37 @@
...
+from typing import Any, cast
+
+
+def _ctx_lru_cache(ctx, f, maxsize):
+    """Cache *f* by its arguments and the context's numerical state."""
+    has_rounding = hasattr(ctx, 'rounding')
+    has_trap_complex = hasattr(ctx, 'trap_complex')
+
+    @lru_cache(maxsize=maxsize)
+    def cached(_ctx_state, /, *args, **kwargs):
```
skirpichev, line 8: Can you expose cache_info/cache_clear for decorated function?  Per #929, I think it's a good idea to have some control of caches.

mpmath/tests/test_elliptic.py
```diff
@@ -1202,6 +1203,110 @@ def test_weierstrass_conversions_with_weierp():
     assert mpc_ae(weierp(z, g2=g2, g3=g3),
                   weierp(z, omega1=omega1, omega2=omega2), eps=eps*1000)
 
+
+def test_g2g3from_lru_cache(monkeypatch):
```
skirpichev: Assuming you test the new decorator (in test_functions.py), I think those tests are redundant.

mpmath/tests/test_functions.py
```diff
@@ -27,6 +29,53 @@ def mpc_ae(a, b, eps=eps):
...
+    def _test_lru_cache(ctx, x):
+        calls.append((ctx, x))
+        return len(calls)
+
+    name = _test_lru_cache.__name__
+    assert SpecialFunctions.lru_cache_functions[name] == 2
+    context_classes = (type(mp), type(fp))
+    try:
+        ctx = mp.clone()
+        cached = vars(ctx)['_test_lru_cache']
```
skirpichev:
```suggestion
        cached = ctx._test_lru_cache
```

mpmath/tests/test_functions.py
```diff
@@ -27,6 +29,53 @@ def mpc_ae(a, b, eps=eps):
...
+        expected_signature = inspect.signature(
+            _test_lru_cache.__get__(ctx, type(ctx)))
```
skirpichev:
```suggestion
        expected_signature = inspect.signature(cached.__wrapped__)
```
no?

mpmath/tests/test_functions.py
```diff
@@ -27,6 +29,53 @@ def mpc_ae(a, b, eps=eps):
...
+
+        ctx.prec -= 10
+        assert cached(2) == cached(2) == 5
+        ctx.rounding = 'd'
+        assert cached(2) == cached(2) == 6
+        ctx.trap_complex = True
+        assert cached(2) == cached(2) == 7
+
+        fixed_ctx = type(fp)()
+        fixed_cached = vars(fixed_ctx)['_test_lru_cache']
```
skirpichev:
```suggestion
        fixed_ctx = type(fp)()
        fixed_cached = fixed_ctx._test_lru_cache
```

mpmath/tests/test_functions.py
```diff
@@ -27,6 +29,53 @@ def mpc_ae(a, b, eps=eps):
...
+        assert fixed_cached(1) == fixed_cached(1) == 8
+    finally:
+        for context_class in context_classes:
+            if name in vars(context_class):
+                delattr(context_class, name)
```
skirpichev:
```suggestion
        del type(mp)._test_lru_cache
        del type(fp)._test_lru_cache
```

skirpichev (commented):
> There would be a bit more work to do to make it a drop in replacement for the existing foofrom methods because they are using defun_wrapped and that decorator is doing a bit more so we'd likely need to modify one of the two decorators to combine roles.

Maybe we shouldn't mix cache with defun?  I.e. make possible to combine decorators.

HeskethGD:
> Make context LRU caching composable

Done. The `defun_lru_cache` decorator is replaced with the `ctx_lru_cache` decorator which can be composed with either `defun` or `defun_wrapped`. Tests are added to confirm composability but I have purposely not used it everywhere yet to keep the PR targeted.

mpmath/functions/functions.py
HeskethGD: Done.

mpmath/functions/functions.py
HeskethGD: Replaced with suggested `getattr`.

mpmath/functions/functions.py
HeskethGD: Done.

mpmath/tests/test_elliptic.py
HeskethGD: Reduced some that were duplicitous, retained where specific integration was being tested.

mpmath/tests/test_functions.py
HeskethGD: Done

mpmath/tests/test_functions.py
HeskethGD: Done

mpmath/tests/test_functions.py
HeskethGD: Done.

mpmath/tests/test_functions.py
HeskethGD: Done.

mpmath/tests/test_elliptic.py
```diff
@@ -1202,6 +1202,19 @@ def test_weierstrass_conversions_with_weierp():
     assert mpc_ae(weierp(z, g2=g2, g3=g3),
                   weierp(z, omega1=omega1, omega2=omega2), eps=eps*1000)
 
+
+def test_omega1omega2from_lru_cache_trap_complex():
```
skirpichev: I think you can test this in test_ctx_lru_cache().  Use something like `sqrt(x)` in the `_test_lru_cache()` body.

mpmath/functions/functions.py
```diff
@@ -61,6 +97,13 @@ def defun(f):
     SpecialFunctions.defined_functions[f.__name__] = f, False
     return f
 
+def ctx_lru_cache(maxsize):
```
skirpichev, line 100: Why not `lru_cache`?

mpmath/tests/test_functions.py
```diff
@@ -27,6 +30,83 @@ def mpc_ae(a, b, eps=eps):
...
+
+    name = _test_lru_cache.__name__
+    assert SpecialFunctions.lru_cache_functions[name] == 2
+    assert SpecialFunctions.defined_functions[name][1] is False
+    wrapped_name = _test_wrapped_lru_cache.__name__
+    assert SpecialFunctions.lru_cache_functions[wrapped_name] == 2
+    assert SpecialFunctions.defined_functions[wrapped_name][1] is True
+    try:
+        ctx = mp.clone()
+        fixed_ctx = type(fp)()
```
skirpichev: Move this down, to fp tests.

skirpichev (commented):
> The defun_lru_cache decorator is replaced with the ctx_lru_cache decorator which can be composed with either defun or defun_wrapped.

I think now you can add new decorator also for other `from*` functions.  I would appreciate benchmarks, of course.

HeskethGD:
> I think now you can add new decorator also for other `from*` functions. I would appreciate benchmarks, of course.

Added on all `*from` param converter functions. The PR description is updated with a new section including benchmarks and new script. Jacobi elliptic functions are about 1.24× faster when evaluated at multiple `z` values using the same `m`, bringing performance approximately in line with supplying `q` directly.

mpmath/tests/test_elliptic.py
HeskethGD: This has been removed and replaced with a sqrt test in test_functions.py as suggested.

mpmath/tests/test_functions.py
HeskethGD: This was replaced with a test_fp_ctx_lru_cache test in test_fp.py

mpmath/functions/functions.py
HeskethGD: `lru_cache` is the name of the method in Python and while it would be possible to alias and avoid a clash it seemed that there could be potential for developer confusion down the road. Hence the `ctx_` prefix was included to indicate the way it differs from the default `lru_cache` by being context aware.

skirpichev: Ok, it's an internal helper, anyway.

mpmath/tests/test_fp.py
```diff
@@ -24,13 +24,40 @@
...
+from mpmath.functions.functions import SpecialFunctions, ctx_lru_cache, defun
 
 
 def ae(x, y, tol=1e-12):
     if x == y:
         return True
     return abs(x-y) <= tol*abs(y)
 
+
+def test_fp_ctx_lru_cache(monkeypatch):
```
skirpichev: I would prefer to keep all tests for new decorator in the test_functions.py.

mpmath/tests/test_elliptic.py
```diff
@@ -1202,6 +1202,7 @@ def test_weierstrass_conversions_with_weierp():
     assert mpc_ae(weierp(z, g2=g2, g3=g3),
                   weierp(z, omega1=omega1, omega2=omega2), eps=eps*1000)
 
+
```
skirpichev:
```suggestion
```

mpmath/functions/functions.py
```diff
@@ -1,3 +1,34 @@
...
+        # otherwise use this argument.
+        return f(*args, **kwargs)
+
+    @wraps(f)
+    def f_cached(*args, **kwargs):
+        # Fixed-precision and interval contexts do not expose rounding or
+        # trap_complex, so use a stable placeholder for those contexts.
+        rounding = getattr(ctx, 'rounding', None)
+        trap_complex = getattr(ctx, 'trap_complex', None)
+        ctx_state = (ctx.prec, rounding, trap_complex)
```
skirpichev: Nit:
```suggestion
        ctx_state = ctx.prec, rounding, trap_complex
```

skirpichev (changes requested):
Few final comments, thanks for a patience.

I would appreciate if you provide a draft for the commit message.

mpmath/functions/functions.py
HeskethGD: Done

mpmath/tests/test_elliptic.py
HeskethGD: Done

mpmath/tests/test_fp.py
HeskethGD: Done

HeskethGD:
> Few final comments, thanks for a patience.
> 
> I would appreciate if you provide a draft for the commit message.

Draft commit message:

Cache elliptic parameter conversions

Add a context-aware LRU cache that can be combined with defun or
defun_wrapped. Each context owns a separate 16-entry cache, with
precision, rounding mode and trap_complex included in the cache key.

Apply the cache to the elliptic parameter conversion functions to avoid
repeating conversions when elliptic functions are evaluated at multiple
points with the same parameters. Preserve public signatures and expose
cache_info() and cache_clear() for cache inspection and control.

mpmath/functions/functions.py
skirpichev: BTW, note that subscribed people will get email on each you "Done" message.

It doesn't have much sense and makes pr thread noisy.  Next time, if you are agree with suggestion - just apply it locally and resolve conversation.

skirpichev (approved):
LGTM

Lets merge this tomorrow.

skirpichev:
Thanks.

PR #1174, opened 2026-09-07
Optimize jtheta on the fundamental domain boundary

HeskethGD:
# Summary

Make `jtheta`’s modular-boundary decision tolerant of finite-precision rounding and preserve that decision during recursive argument reduction. This avoids unnecessary modular transformations for boundary lattices while retaining them for values resolvably outside the fundamental domain.

## Problem addressed

`jtheta` decides whether to apply a modular transformation by reconstructing `tau` from `q` and testing the standard fundamental-domain boundaries:

```text
abs(Re(tau)) <= 1/2
abs(tau) >= 1
```

The unit circle is the natural crossover for inversion `tau' = -1/tau`, since

```text
Im(tau') = Im(tau) / abs(tau)**2.
```

Inside the circle, inversion improves convergence by increasing `Im(tau)` and reducing `abs(q)`; outside it, it does the opposite. On the circle, both representations have identical convergence, so transformation only adds overhead.

Finite-precision period arithmetic can move a mathematically boundary value slightly across the circle. For example, with `g2=-4`, `g3=1`, the computed ratio may satisfy

```text
abs(tau)**2 = 1 - 0.5*mp.eps
```

even though the exact ratio lies on `abs(tau)=1`. The previous strict comparison could therefore trigger an unnecessary transformation, with recursive argument reduction making the decision less stable under guard bits.

These cases occur naturally. Real Weierstrass invariants with negative `g2` and negative discriminant produce rhombic lattices whose canonical period ratio lies on `abs(tau)=1`; the square and hexagonal lattices are familiar special cases. Generic rhombic ratios need not remain exactly on the circle after rounding.

### Fix

The new branch treats displacements within the standard `almosteq` tolerance `2**(-p+4)`, for original-call precision `p`, as boundary values and preserves that decision recursively. Values resolvably beyond the boundary still use modular reduction.

This is a rounding-sized tie region around the existing mathematical crossover, not a new performance-tuned threshold. Within it, the two representations have indistinguishable convergence, so the direct form is preferred to avoid transformation overhead. The tolerance shrinks with precision.

This remains safely within the rapidly convergent regime. On the circular boundary of the fundamental domain,

```text
abs(q) <= exp(-pi*sqrt(3)/2) ~= 0.0658,
```

and an `O(2**-p)` extension has negligible effect. The change therefore affects only the choice between equivalent evaluation formulas near the boundary, not the domain of convergence.

### Future considerations

A separately tuned crossover based on measured transformation cost could improve performance further, but that would require broader benchmarking. This change is narrower: it makes the existing mathematical crossover stable under finite-precision arithmetic.

## Direct `jtheta` cases

Benchmarks compare `master` at `08041f8d` with the `jtheta-bug` working tree at the same revision, using 50 decimal digits and reporting time per function call.

Each case evaluates theta indices 1 through 4 at 100 varying complex arguments. Three independent points span the right-hand circular arc, and each has matched points resolvably inside and outside the circle:

| Case | `tau` on the unit circle |
|---|---:|
| circle-01 | `0.1 + sqrt(0.99)j` |
| circle-02 | `0.3 + sqrt(0.91)j` |
| circle-03 | `0.4 + sqrt(0.84)j` |

The inside-circle and outside-strip cases must continue to use modular transformations. The outside-circle cases must continue to use the direct series.

| Case | master | jtheta-bug | Comparison |
|---|---:|---:|---:|
| circle-01 boundary | 327.10 us | 119.32 us | **2.74x faster** |
| circle-02 boundary | 113.33 us | 113.66 us | no significant change |
| circle-03 boundary | 333.11 us | 119.48 us | **2.79x faster** |
| circle-01 inside control | 328.84 us | 317.76 us | 1.03x faster |
| circle-02 inside control | 333.52 us | 317.15 us | 1.05x faster |
| circle-03 inside control | 340.59 us | 317.72 us | 1.07x faster |
| circle-01 outside control | 115.30 us | 115.82 us | no significant change |
| circle-02 outside control | 115.72 us | 115.71 us | no significant change |
| circle-03 outside control | 118.29 us | 116.90 us | 1.01x faster |
| Vertical boundary | 93.16 us | 92.88 us | 1.00x faster |
| Outside vertical-strip control | 285.18 us | 272.01 us | 1.05x faster |

At 50 decimal digits, circle-01 and circle-03 round half an epsilon inside the unit circle. On `master`, they consequently cost essentially the same as the genuinely-inside controls. Circle-02 lands exactly on the boundary and is
already fast on `master`. All three boundary cases cost essentially the same as their outside-circle controls on the branch. The modular decisions for all inside and outside controls remain unchanged.

## Weierstrass boundary lattices

Each case evaluates 100 varying arguments for three distinct real-invariant lattices whose period ratios lie at different positions on the unit circle:

| Case | `g2` | `g3` | Approximate `tau` |
|---|---:|---:|---:|
| lattice-01 | `-10` | `3` | `0.121587 + 0.992581j` |
| lattice-02 | `-4` | `1` | `0.153519 + 0.988146j` |
| lattice-03 | `-1` | `1` | `0.397761 + 0.917489j` |

| Case | Function | master | jtheta-bug | Improvement |
|---|---|---:|---:|---:|
| lattice-01 | `weierp` | 853.13 us | 391.45 us | **2.18x faster** |
| lattice-01 | `weierpprime` | 1673.16 us | 747.15 us | **2.24x faster** |
| lattice-01 | `weiersigma` | 452.64 us | 229.26 us | **1.97x faster** |
| lattice-01 | `weierzeta` | 1275.37 us | 496.08 us | **2.57x faster** |
| lattice-02 | `weierp` | 855.09 us | 380.84 us | **2.25x faster** |
| lattice-02 | `weierpprime` | 1639.94 us | 745.26 us | **2.20x faster** |
| lattice-02 | `weiersigma` | 452.48 us | 225.20 us | **2.01x faster** |
| lattice-02 | `weierzeta` | 1258.43 us | 486.72 us | **2.59x faster** |
| lattice-03 | `weierp` | 710.09 us | 383.74 us | **1.85x faster** |
| lattice-03 | `weierpprime` | 1373.19 us | 741.47 us | **1.85x faster** |
| lattice-03 | `weiersigma` | 383.73 us | 228.90 us | **1.68x faster** |
| lattice-03 | `weierzeta` | 1162.32 us | 500.12 us | **2.32x faster** |

Instrumentation outside the timed benchmark confirmed that the branch makes zero modular transformations during a 100-point `weierp` sweep on lattice-02. The same sweep previously made 200 unnecessary transformations.

<details>

Save the following as `benchmark_jtheta_boundaries_pyperf.py` in the repository root and run it once on `master` and once on this branch:

```python
#!/usr/bin/env python3
"""Benchmark jtheta modular boundaries and boundary-lattice functions."""

import subprocess
import time

import pyperf

import mpmath
from mpmath import mp


mp.dps = 50
POINT_COUNT = 100
GOLDEN_RATIO_CONJUGATE = (mp.sqrt(5) - 1) / 2
BOUNDARY_DELTA = mp.ldexp(1, -20)


def theta_points():
    return [
        mp.mpf("0.1") + mp.mpf("0.8") * (
            index + mp.mpf("0.5")
        ) / POINT_COUNT + mp.j * mp.mpf("0.1") * mp.frac(
            (index + mp.mpf("0.5")) * GOLDEN_RATIO_CONJUGATE
        )
        for index in range(POINT_COUNT)
    ]


def lattice_points(omega1, omega2):
    points = []
    for index in range(POINT_COUNT):
        a = mp.mpf("0.1") + mp.mpf("0.8") * (
            index + mp.mpf("0.5")
        ) / POINT_COUNT
        b = mp.mpf("0.1") + mp.mpf("0.8") * mp.frac(
            (index + mp.mpf("0.5")) * GOLDEN_RATIO_CONJUGATE
        )
        points.append(2 * (a * omega1 + b * omega2))
    return points


def bench_jtheta(loops, points, q):
    start = time.perf_counter()
    for _ in range(loops):
        for z in points:
            for n in range(1, 5):
                mp.jtheta(n, z, q)
    return time.perf_counter() - start


def bench_weierstrass(loops, function, points, g2, g3):
    start = time.perf_counter()
    for _ in range(loops):
        for z in points:
            function(z, g2=g2, g3=g3)
    return time.perf_counter() - start


circle_taus = (
    ("circle-01", mp.mpf("0.1") + mp.j * mp.sqrt(mp.mpf("0.99"))),
    ("circle-02", mp.mpf("0.3") + mp.j * mp.sqrt(mp.mpf("0.91"))),
    ("circle-03", mp.mpf("0.4") + mp.j * mp.sqrt(mp.mpf("0.84"))),
)
vertical_tau = mp.mpf("0.5") + mp.j * mp.mpf("1.2")

THETA_CASES = tuple(
    ("%s boundary" % name, tau)
    for name, tau in circle_taus
) + tuple(
    ("%s inside control" % name, tau * (1 - BOUNDARY_DELTA))
    for name, tau in circle_taus
) + tuple(
    ("%s outside control" % name, tau * (1 + BOUNDARY_DELTA))
    for name, tau in circle_taus
) + (
    ("vertical boundary", vertical_tau),
    ("outside vertical-strip control", vertical_tau + BOUNDARY_DELTA),
)

WEIERSTRASS_CASES = (
    ("lattice-01", mp.mpf(-10), mp.mpf(3)),
    ("lattice-02", mp.mpf(-4), mp.mpf(1)),
    ("lattice-03", mp.mpf(-1), mp.mpf(1)),
)

    ["git", "rev-parse", "--short", "HEAD"], text=True
).strip()
runner = pyperf.Runner(metadata={
    "mpmath_revision": revision,
    "mpmath_file": mpmath.__file__,
    "mpmath_dps": mp.dps,
    "points_per_parameter": POINT_COUNT,
})

points = theta_points()
for case_name, tau in THETA_CASES:
    runner.bench_time_func(
        "jtheta 1-4: %s" % case_name,
        bench_jtheta,
        points,
        mp.qfrom(tau=tau),
        inner_loops=4 * POINT_COUNT,
    )

for case_name, g2, g3 in WEIERSTRASS_CASES:
    omega1, omega2 = mp.omega1omega2from(g2=g2, g3=g3)
    points = lattice_points(omega1, omega2)
    for function_name, function in (
        ("weierp", mp.weierp),
        ("weierpprime", mp.weierpprime),
        ("weiersigma", mp.weiersigma),
        ("weierzeta", mp.weierzeta),
    ):
        runner.bench_time_func(
            "%s: %s unit-circle boundary" % (function_name, case_name),
            bench_weierstrass,
            function,
            points,
            g2,
            g3,
            inner_loops=POINT_COUNT,
        )
```

To run it:

```console
git switch master
python benchmark_jtheta_boundaries_pyperf.py \
    -o jtheta-boundaries-master.json

git switch jtheta-bug
python benchmark_jtheta_boundaries_pyperf.py \
    -o jtheta-boundaries-jtheta-bug.json

python -m pyperf compare_to \
    jtheta-boundaries-master.json \
    jtheta-boundaries-jtheta-bug.json \
    --table --table-format md
```

</details>

Pyperf reported instability warnings for some cases under its strict sub-1% variation criterion. The affected boundary improvements are much larger than the recorded variation. No control case regressed significantly in this run; their changes ranged from no significant difference to 7% faster.

## AI use declaration

OpenAI Codex was used to assist with implementation, benchmark design and execution. I identified the need and the opportunity, designed the experiment, reviewed the code and benchmark methodology, and reported results.


mpmath/functions/theta.py
```diff
@@ -455,7 +455,11 @@ def _jtheta_needs_modular(ctx, z, q):
         return False
     tau = ctx.taufrom(q=q)
     assert abs(q) < 1 and tau.imag > 0
-    return abs(tau.real) > 0.5 or tau.real**2 + tau.imag**2 < 1
+    real = abs(tau.real)
+    if real > ctx.one/2 and not ctx.almosteq(real, ctx.one/2):
```
skirpichev:
```suggestion
    if real > 0.5 and not ctx.almosteq(real, 0.5):
```

0.5 is exactly representable in binary.

mpmath/functions/theta.py
```diff
@@ -455,7 +455,11 @@ def _jtheta_needs_modular(ctx, z, q):
...
+        return True
+    norm = tau.real**2 + tau.imag**2
+    return norm < ctx.one and not ctx.almosteq(norm, ctx.one)
```
skirpichev:
```suggestion
    return norm < 1 and not ctx.almosteq(norm, 1)
```

mpmath/functions/theta.py
```diff
@@ -479,31 +483,22 @@ def terms():
...
-    z = ctx.convert(z)
-    q = ctx.convert(q)
-    nd = int(derivative)
-
-    if n not in range(1, 5):
-        raise ValueError("First argument expected to be 1, 2, 3 or 4")
-    if abs(q) >= 1:
-        raise ValueError(f"abs(q) >= 1")
-
+def _jtheta(ctx, n, z, q, nd, tau_reduced=False):
```
skirpichev: I don't see a good reason for new parameter, could you explain?  Is this just for cache?

HeskethGD: The parameter was intended to prevent recursive evaluation from repeating the modular-reduction decision under a different guard precision, rather than for caching. I’ve now removed it by separating the public dispatch from the internal post-dispatch evaluator, which preserves the same behaviour and performance without requiring an explicit state parameter.

skirpichev: I would prefer to separate this from the main change and be benchmarked.

HeskethGD: If by “this” you mean extracting the post-dispatch evaluation into `_jtheta`, I considered that part of the boundary fix rather than a separate optimization. The problem occurs because recursive evaluation re-enters public `jtheta` under guard precision and repeats the modular-reduction decision. Splitting the public dispatcher from the internal evaluator allows that decision to be made once at the caller’s precision.

I also implemented a more local alternative without `_jtheta`, available on the [`jtheta-boundary-minimal` branch](https://github.com/HeskethGD/mpmath/tree/jtheta-boundary-minimal). It compensates for the guard-precision recursion by widening the boundary tolerance, coupling that tolerance to the number of guard bits. Recursive calls also repeat public argument conversion, validation, reconstruction of `tau` from `q`, and the modular-boundary check.

I benchmarked the two implementations directly:

| Workload | Cases | Local alternative | `_jtheta` version | `_jtheta` comparison |
|---|---:|---:|---:|---:|
| Exact circle boundaries | 3 | 116.85 us | 116.07 us | 1.01x faster |
| Inside-circle controls | 3 | 336.11 us | 317.76 us | 1.06x faster |
| Outside-circle controls | 3 | 116.60 us | 116.91 us | no change |
| Boundary Weierstrass functions | 12 | 445.04 us | 423.83 us | 1.05x faster |
| All benchmarks | 23 | 278.10 us | 267.64 us | **1.04x faster** |

Both implementations produce the intended behaviour for the tested boundary cases, but the `_jtheta` version is approximately 4% faster overall. I therefore favoured retaining it because it represents the single modular-reduction decision structurally, avoids coupling the comparison tolerance to an internal guard-precision value, and avoids repeated work during recursive evaluation.

HeskethGD:
I am closing this for now as there may be a better solution and this is only an optimisation not a wrong value bug.

skirpichev:
@HeskethGD, sorry for being unresponsive.  If you think there is a bug, please open an issue.

PR #1173, opened 2026-09-06
Ensure `acos(h)`/`asin(h)` respect special cases from applicable standards

mdhaber:
As discussed in gh-1172, this PR resolves the disagreement between `acos`'s current behavior and that specified in the [array API standard](https://data-apis.org/array-api/2025.12/API_specification/generated/array_api.acos.html).

Because it is my first contribution, this PR is intended to be minimal and follow nearby patterns as closely as possible. Let me know what you'd prefer to see in the future and I'll make it so.

A more complete version, which tests _all_ the special cases from the standard - not just the ones that were failing - can be seen at:
https://github.com/mdhaber/mpmath/pull/new/fix_all_acos_special_cases
I'd prefer to go that route with this and future PRs (for other trig, hyperbolic trig, etc.), but it is more to review, so I decided to submit the minimal version. Let me know which direction you'd prefer. I've marked this as draft in the meantime.

When working on SciPy or MPArray, I never let AI draft my code. I did here because the changes are very mechanical. Of course, it (GPT 5.6 Sol, medium effort) made many mistakes, but I think I fixed it all up so it is compact and consistent with the existing code.

Issue #1172, opened 2026-09-06
Adjustments toward Python array API standard compatibility


mdhaber:
While developing `scipy.stats`, `scipy.special`, and other SciPy sub-packages, we often use `mpmath` to generate reference values, especially when we're working on improving our calculations in some tricky corner of the domain. Thanks, `mpmath`! 

Sometimes we can jump through numerical hoops to avoid over/underflow and get good accuracy in `float64`. But in many cases, it would be nice to allow our users to crunch the numbers in higher precision. NumPy object arrays with with `mpf`s and `mpc`s get us most of the way there, but to close the gap, I wrote [MPArray](https://github.com/mdhaber/mparray/tree/main), a [Python array API standard](https://data-apis.org/array-api/latest/)-compatible backend that leverages `mpmath` for real and complex floating point calculations. scipy/scipy#24840 proposes to add it as an officially supported backend in SciPy 2.0.

The library passes most of the [array-api-tests](https://github.com/data-apis/array-api-tests) suite, but there are a lot of special cases specified in the standard that `mpmath` doesn't implement. Would `mpmath` consider PRs to patch these?

As an example of what these PRs might look like, https://github.com/mpmath/mpmath/pull/795 happened to fix some of the special cases I found while working with mpmath 1.3.0. As an example of a disagreement that persists in mpmath 1.4.1, the standard specifies:

> For complex floating-point operands, let `a = real(x_i)`, `b = imag(x_i)`... If `a` is either `+0` or `-0` and `b` NaN, the result [of `acos(x_i)`] is `π/2+ NaN j`.

But currently:
```
import math
from mpmath import mp
x = mp.mpc(0, mp.nan)
mp.acos(x)
# mpc(real='nan', imag='nan')
```
Would you consider additional contributions that adjust these special cases? I could work around these things in `mparray`, but native support would simplify things considerably, and some of these special case prescriptions are objectively more correct than the current behavior.

---

Also, I just created the [logo](https://github.com/mdhaber/mparray/blob/main/docs/logo.svg) today - do you mind the nod to mpmath's identicon? 

-------

TODO:
- [x] `acos(h)`/`asin(h)` // https://github.com/mpmath/mpmath/pull/1173
- [ ] `exp(m1)` // https://github.com/mpmath/mpmath/pull/1177
- [ ] `pow`
- [ ] `log(1p)`
- [ ] `cos(h)`/`sin(h)`
- [ ] `tan(h)`
- [ ] `sqrt`
[...]

PR #1177, opened 2026-09-10
Ensure `exp`/`expm1` respect special cases from applicable standards

mdhaber:
Toward gh-1172

Add special case logic and regression tests to ensure that `exp` and `expm1` respect special cases defined in the C99 standard and in the Python array API standard.

Applies lessons learned from gh-1173:
- only add tests, and add only the tests from the standard not already present
- use `isnan`, `==`, `.ae`, literals, and predefined constants as appropriate
- cite C standard
- prefer minimal, efficient implementation over the obvious, case by case solution

Similar use of GPT-5.6 Sol for drafting; lots of interactive/manual refinement.

### DIFF
--- a/mpmath/tests/test_functions.py
+++ b/mpmath/tests/test_functions.py
@@ -3,10 +3,7 @@ import inspect
 import math
 import random
 
-import numpy as np
 import pytest
-from hypothesis import given, settings
-from hypothesis import strategies as st
 
 from mpmath import (acos, acosh, acot, acoth, acsc, acsch, arange, arg, asec,
                     asech, asin, asinh, atan, atan2, atanh, catalan, cbrt,
@@ -25,14 +22,6 @@ from mpmath.libmp import (MPZ, ComplexResult, from_int, mpf_gt, mpf_lt,
                           mpf_mul, mpf_pow_int, mpf_sqrt, round_ceiling,
                           round_down, round_nearest, round_up)
 from mpmath.libmp.libmpf import mpf_rand
-
-
-_exp_component_values = [
-    0.0, 1.0, -1.0, math.pi/4, -math.pi/4, 3*math.pi/4, -3*math.pi/4,
-    math.inf, -math.inf, math.nan]
-_exp_components = st.sampled_from(_exp_component_values)
-_exp_imag_components = st.one_of(st.none(), _exp_components)
-_exp_case_count = len(_exp_component_values) * (len(_exp_component_values) + 1)
 
 
 def mpc_ae(a, b, eps=eps):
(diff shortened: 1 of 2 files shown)


mpmath/tests/test_functions.py
```diff
@@ -263,6 +263,34 @@ def test_exact_cbrt():
...
+    assert abs(r.real) == inf and isnan(r.imag)
+    r = exp_fun(mpc(nan, 0))
+    assert isnan(r.real) and r.imag == 0
+    r = exp_fun(mpc(nan, 1))
+    assert isnan(r.real) and isnan(r.imag)
+    r = exp_fun(mpc(nan, nan))
+    assert isnan(r.real) and isnan(r.imag)
+
 def test_exp():
     assert exp(0) == 1
```
mdhaber, line 278: Other real special cases are already in `test_special.py::test_functions_special`

mpmath/tests/test_functions.py
```diff
@@ -263,6 +263,34 @@ def test_exact_cbrt():
         A = random.randint(10**(prec//2-2), 10**(prec//2-1))
         assert cbrt(mpf(A*A*A)) == A
 
+def _test_exp_special_cases(exp_fun, m):
```
skirpichev: Lets just inline this code to test_exp and test_expm1.

mdhaber: I had it separate, then intentionally factored it out so we didn't have two versions of almost exactly the same code, and also because the standard only specifies the complex special cases for `exp`; the `expm1` are derived from those. You're asking me to undo that - copy-paste it back into the separate `test_exp` and `test_expm` functions and hard-code the values of `exp_fun` and `m` in each?

skirpichev: > separate test_exp and test_expm functions and hard-code the values of exp_fun and m in each?

Yes.  Lets just keep explicit tests.

skirpichev (dismissed):

mdhaber:
Thanks for the review. Squash merge with original title and commit message is fine with me:

> Ensure `exp`/`expm1` respect special cases from applicable standards
>
> Add special case logic and regression tests to ensure that `exp` and `expm1` respect
> special cases defined in the C99 standard and in the Python array API standard.

mpmath/tests/test_functions.py
```diff
@@ -276,6 +276,32 @@ def test_exp():
     assert exp(ln2 * 10).ae(1024)
     assert exp(2+2j).ae(cmath.exp(2+2j))
 
+    # Complex special cases:
+    # https://en.cppreference.com/c/numeric/complex/cexp
+    assert exp(0j) == 1
+    r = exp(mpc(1, inf))
```
skirpichev: Where test for `exp(1-infj)`?

mpmath/tests/test_functions.py
```diff
@@ -276,6 +276,32 @@ def test_exp():
...
+    assert isnan(r.real) and isnan(r.imag)
+    r = exp(mpc(1, nan))
+    assert isnan(r.real) and isnan(r.imag)
+    assert exp(mpc(inf, 0)) == mpc(inf, 0)
```
skirpichev: What about `exp(-inf+0j)`?

mpmath/tests/test_functions.py
```diff
@@ -276,6 +276,32 @@ def test_exp():
...
+    assert exp(mpc(-inf, 1)) == 0
+    assert exp(mpc(inf, pi/4)) == mpc(inf, inf)
+    assert exp(mpc(inf, 3*pi/4)) == mpc(-inf, inf)
+    assert exp(mpc(inf, -3*pi/4)) == mpc(-inf, -inf)
+    assert exp(mpc(inf, -pi/4)) == mpc(inf, -inf)
+    assert exp(mpc(-inf, inf)) == 0
+    r = exp(mpc(inf, inf))
```
skirpichev: `exp(-inf-infj)` and `exp(inf-infj)`?

mdhaber:
None of those are specified explicitly, so I didn't include them, but I see they can be derived based on the conjugation identity, so I'll add them.

mpmath/tests/test_functions.py
mdhaber: What about `exp(-inf+0.5j)`? I suppose `0j` tests a different code path though. I could go for property-based tests in this sort of situation, but it may be overkill.

mpmath/tests/test_functions.py
```diff
@@ -276,6 +276,38 @@ def test_exp():
...
+    assert exp(mpc(-inf, inf)) == 0
+    assert exp(mpc(-inf, -inf)) == 0
+    r = exp(mpc(inf, inf))
+    assert abs(r.real) == inf and isnan(r.imag)
+    r = exp(mpc(inf, -inf))
+    assert abs(r.real) == inf and isnan(r.imag)
+    assert exp(mpc(-inf, nan)) == 0
+    r = exp(mpc(inf, nan))
+    assert abs(r.real) == inf and isnan(r.imag)
+    r = exp(mpc(nan, 0))
```
skirpichev: `exp(nanj)`? `exp(nan+infj)`?  `exp(nan-infj)`?

mdhaber:
[3b3246c](https://github.com/mpmath/mpmath/pull/1177/commits/3b3246c901659cbd5dcdcf4d337fba78ee542a40) adds only the requested additional cases.
[8d11ba1](https://github.com/mpmath/mpmath/pull/1177/commits/8d11ba1577b5f95dd2bdb193a4a6f3c894afd37b) adds all the remaining combinations of +/- versions of common arguments explicitly.
529816f switches to property-based testing against NumPy. 
Any of these can be used as the starting point for additional review.

skirpichev (changes requested):
Please return to [3b3246c](https://github.com/mpmath/mpmath/pull/1177/commits/3b3246c901659cbd5dcdcf4d337fba78ee542a40).

Lets just test special cases from the C standard.  BTW, numpy is optional dependency.

skirpichev:
> switches to property-based testing against NumPy.

Maybe using the fp context for reference values can be considered.  This definitely will simplify review and coding.  Also, we will test some consistency among contexts.

I see some issues, though:
1. we are in mercy of platform libm's, this might break tests
2. probably we can't use equality tests.

mpmath/tests/test_functions.py
mdhaber: https://github.com/mpmath/mpmath/pull/1177/changes/587998f56d5c5f3e3049d1a2a4a7f8f197ffca47..d1b723f507d47d5d048cad498d4c941c7ac49cb1

mpmath/tests/test_functions.py
skirpichev: Maybe it worth to make a comment.

skirpichev (approved):
LGTM, one minor comment

mpmath/tests/test_functions.py
skirpichev, line 268:
```suggestion
    # other special real cases are in test_special.py::test_functions_special()
    assert exp(0) == 1
```

skirpichev:
Thanks.

PR #1178, opened 2026-09-10
Ensure `log`/`log1p`/`atan2` respect special cases from applicable standards

mdhaber:
Toward https://github.com/mpmath/mpmath/issues/1172

Add special case logic and regression tests to ensure that `log` and `log1p` respect special cases defined in the C99 standard and in the Python array API standard. `log` is calculated using `atan2`, so we fix/test that here, too.


mpmath/libmp/libmpc.py
```diff
@@ -431,7 +431,10 @@ def mpc_exp(z, prec, rnd=round_down):
     return re, im
 
 def mpc_ln(z, prec, rnd=round_down):
-    re = mpf_log_hypot(z[0], z[1], prec, rnd)
+    a, b = z
+    if (a in _infs and b == fnan) or (a == fnan and b in _infs):
+        return finf, fnan
```
skirpichev: I think that mpf_log_hypot() should be fixed instead.  Right now we have:
```pycon
>>> log(mpc(-inf, nan))
(nan + nanj)
>>> fp.hypot(-inf, nan)  # same does make sense for log of hypot
inf
>>> libmp.libelefun.mpf_log_hypot((-inf)._mpf_, nan._mpf_, 53, 'd')
(0, mpz(0), -123, -1)
>>> mpf(_)
nan
```

mpmath/libmp/libelefun.py
```diff
@@ -765,10 +765,10 @@ def mpf_log_hypot(a, b, prec, rnd):
         if not b[1]:
             if a == b == fzero:
                 return fninf
-            if fnan in (a, b):
-                return fnan
-            # at least one term is (+/- inf)^2
-            return finf
+            if a in (finf, fninf) or b in (finf, fninf):
+                # at least one term is (+/- inf)^2
+                return finf
```
skirpichev:
```suggestion
                return finf  # at least one term is (+/- inf)^2
```

mpmath/libmp/libelefun.py
```diff
@@ -898,6 +898,8 @@ def mpf_atan2(y, x, prec, rnd=round_down):
                 return fzero
             return mpf_pi(prec, rnd)
         if y in (finf, fninf):
+            if x == fnan:
+                return fnan
```
skirpichev:
```suggestion
```
Looks redundant for me, given L920.

mdhaber: K, now this uses 920.

mpmath/tests/test_functions.py
```diff
@@ -600,6 +630,10 @@ def test_atan2():
     assert isnan(atan2(nan,3))
     assert isnan(atan2(0,nan))
     assert isnan(atan2(nan,0))
+    assert isnan(atan2(inf,nan))
+    assert isnan(atan2(-inf,nan))
+    assert isnan(atan2(nan,inf))
+    assert isnan(atan2(nan,-inf))
```
skirpichev, line 637: Probably, you should adjust pr title/description.  This part looks unrelated.

mpmath/functions/functions.py
```diff
@@ -212,7 +212,7 @@ def log1p(ctx, x):
     cmag = ctx.mag(c)
     a, b = c.real, c.imag
     wp = ctx.prec
-    if cmag >= -wp:
+    if cmag >= -wp or ctx.isnan(cmag):
```
skirpichev: Hmm, we can revert branches to avoid this additional check.  BTW, probably having "small c" case first is more logical.

mdhaber: Reversed the order of the branches.

skirpichev (approved):

mpmath/tests/test_functions.py
mdhaber: `log` is calculated using `atan2`, and we adjusted `atan2` to fix the `NaN ± inf j` behavior. This tests the `atan2` changes directly.

mdhaber:
Changed the title + description and added one missing test. I think the others are covered either nearby or in `test_arg_sign`.

Issue #142, opened 2008-11-19
Enhanced polynomial support


fredrik-johansson:
bc.. I propose to reverse the order of polynomial coefficients used by polyval
etc. It was changed to the present order for compatibility with Matlab and
SciPy, but I find this convention annoying every time I use it.

First, the Matlab absolutely makes no sense when working with polynomials
as truncations of power series. Second, having entry i be the coefficient
of x^i and nothing else is just more convenient and less error-prone.

p. Original issue for "#142":https://github.com/fredrik-johansson/mpmath/issues/142: "http://code.google.com/p/mpmath/issues/detail?id=102":http://code.google.com/p/mpmath/issues/detail?id=102

p. Original author: "https://code.google.com/u/111502149103757882156/":https://code.google.com/u/111502149103757882156/

p. Original owner: "https://code.google.com/u/111502149103757882156/":https://code.google.com/u/111502149103757882156/

skirpichev:
FYI: numpy now deprecate poly1d with the mpmath-like coefficient ordering.  I would welcome same change in the mpmath.

skirpichev:
Lets add a new kwarg, e.g. ``desc=True`` (or ``reversed=True``) for polyval/polyroots functions to describe the current convention of using *descending* order of coefficients (largest degree - first) with a deprecation warning for ``True`` value.  Then, after a release, we could change defaults to ``desc=False`` (ascending order) and deprecate the option.

skirpichev:
https://github.com/mpmath/mpmath/pull/779 deprecate descending order of coefficients in polyval() and friends.  In v1.5 we can make asc=True - a new default.

skirpichev:
With #1057, asc=True is default for 1.5.

PR #1182, opened 2026-09-14
Add Riemann theta functions

### LINKED ISSUE #325: Riemann Theta Function
This is a counterpart for an issue opened for sympy at https://github.com/sympy/sympy/issues/10881 to cover an implementation for numerical values. 

This is a feature request for an implementation of the Riemann Theta function in mpmath.

There is a paper on how to implement the function at http://www.ams.org/journals/mcom/2004-73-247/S0025-5718-03-01609-0/S0025-5718-03-01609-0.pdf . It links to code in Maple here http://www.math.fsu.edu/~hoeij/RiemannTheta/RiemannTheta . The function is in Maple 8+ and is documented here http://www.maplesoft.com/support/help/Maple/view.aspx?path=RiemannTheta . In Mathematica the function is called SiegelTheta and is documented here https://reference.wolfram.com/language/ref/SiegelTheta.html .

In Python there is https://github.com/abelfunctions/abelfunctions/wiki/Getting%20Started and also an old issue open for sage at http://trac.sagemath.org/ticket/6371 .

HeskethGD:
# Summary

Add arbitrary-precision Riemann theta values, partial derivatives, and complete derivative jets to mpmath:

```python
rtheta(z, tau, characteristic=None, derivative=0)
rtheta_jet(z, tau, order, characteristic=None)
```

This generalizes the classical Jacobi theta functions from genus one to multidimensional period matrices. The implementation supports arbitrary genus in principle and is optimized for the low-genus, arbitrary-precision problems that arise in integrable systems, algebraic geometry, and Abelian functions. Genus two and three at precisions up to approximately 100 decimal digits are the principal target.

The public functionality includes:

- arbitrary real characteristics;
- ordinary partial derivatives selected by a multi-index;
- complete derivative jets through a requested total order;
- arbitrary mpmath working precision; and
- validation of the period matrix and input dimensions.

`rtheta_jet` returns every partial derivative through a chosen total order in one evaluation. This is useful in typical integrable-system calculations, where several derivatives at the same argument and period matrix are combined into ratios, products, and logarithmic derivatives.

The change also adds a dedicated Abelian-functions documentation page with definitions, conventions, worked examples, and plots. Focused tests cover classical identities, characteristics, quasi-periodicity, derivatives, modular transformations, higher precision and genus, and independent oracle values.

The result is a pure-Python implementation that integrates with mpmath's contexts and precision model while remaining competitive with established specialized implementations. The comparison suite shows:

- accuracy agrees with FLINT's certified complex balls in every tested case;
- Wolfram scaled differences are at most `7.18e-30` at 30 digits and `2.66e-100` at 100 digits, while genus-one comparisons with `jtheta` scale from `1.99e-13` at 15 digits (see note below) to `1.49e-98` at 100 digits;
- value evaluation is faster than Wolfram's `SiegelTheta` in geometric mean on the tested workloads, with comparable observed accuracy and a direct derivative and jet API;
- near parity with the specialized genus-one `jtheta` code for repeated evaluations at fixed `tau`;
- particularly close performance to compiled FLINT for mixed second and third derivatives; and
- strong results against the established Sage implementations, especially when a period matrix benefits from automatic reduction, together with full mpmath arbitrary-precision behaviour.

## Related issue

Closes #325

## Method

The direct evaluator starts from Deconinck, Heil, Bobenko, van Hoeij and Schmies, [*Computing Riemann Theta Functions*](https://doi.org/10.1090/S0025-5718-03-01578-8), Mathematics of Computation 73 (2004), 1417--1442. It combines ellipsoidal lattice truncation with cost-selected partial Siegel reduction, recurrence-based summation, and precision-aware caching. The existing `jtheta` implementation is unchanged.

## Accuracy

The accuracy suite combines three kinds of comparison. FLINT supplies a particularly strong oracle because its results are returned as certified complex balls. Wolfram provides an independent implementation for values and, through numerical differentiation, an independent composite check of selected derivatives. In genus one, the established mpmath `jtheta` implementation gives a direct comparison for all four half-integer characteristics.

| Reference | Scope | Result |
| --- | --- | --- |
| Python-FLINT 0.9.0 / FLINT 3.6.0 | 84 value and derivative workloads at 30 and 50 decimal digits | Every mpmath result was enclosed by the corresponding FLINT complex ball. The largest scaled midpoint differences were `3.53e-30` at 30 digits and `4.11e-49` at 50 digits. |
| Wolfram `SiegelTheta` | Genus 1--3 values at 30 and 100 digits, plus selected composite derivative checks | The largest scaled value differences were `7.18e-30` at 30 digits and `2.66e-100` at 100 digits. Selected first-derivative differences were at most `1.02e-31` and `7.80e-102` respectively. |
| mpmath `jtheta` | All four genus-one half-integer characteristics, including first and second derivatives | The largest scaled differences were `1.99e-13`, `7.67e-49`, and `1.49e-98` at 15, 50, and 100 digits respectively. The largest 15-digit discrepancy occurred in a deliberately strongly unreduced derivative case. It retained about 12.7 relative decimal digits. |

The FLINT checks cover zero and nonzero characteristics, values, first derivatives, mixed second derivatives, third derivatives, dimensions up to genus five, and both well-damped and deliberately unreduced period matrices. FLINT's Taylor coefficients are converted to ordinary partial derivatives before comparison.

The Wolfram derivative comparison is not a direct comparison with a native `SiegelTheta` derivative interface. It constructs derivatives using `NumericalCalculus`'s numerical differentiation machinery, so it is useful as an independent accuracy check but not as a derivative-performance benchmark.

### A demanding genus-one transformation

The largest 15-digit `jtheta` difference, `1.99e-13`, comes from a deliberately weakly damped derivative case with `tau = 0.17 + 0.003i`. Automatic reduction makes this case practical while retaining about 12.7 relative decimal digits. The same comparison scales to `7.67e-49` and `1.49e-98` at 50 and 100 digits.

### Accuracy near zeros

Near a theta zero, fixed absolute accuracy implies fewer relative digits. A targeted genus-two regression retains absolute error around `1e-34` at 30 digits. FLINT's certified balls and Wolfram's precision tracking show the same expected transition from relative to absolute accuracy as the value approaches zero. Applications requiring rigorous enclosures can use FLINT's ball-valued interface.

## Performance

**On the benchmark suite, `rtheta` evaluates values faster than Wolfram's `SiegelTheta` in geometric mean, with comparable observed accuracy and the additional convenience of direct partial derivatives and complete jets. As a pure-Python arbitrary-precision implementation, it is also competitive with compiled FLINT, specialized genus-one `jtheta`, and established Sage implementations.**

All ratios below are

```text
mpmath elapsed time / comparison implementation elapsed time
```

so a ratio below 1 means that mpmath was faster. Geometric means give equal weight to each benchmark row and reduce the distortion caused by a few very slow cases.

### Impact of recurrence-based summation

The evaluator computes one exponential at the start of each lattice row and generates adjacent terms by multiplicative recurrence, restarting at every row to limit error accumulation. In the same-run 50- and 100-digit comparison this reduced mean runtime by 42.1%; value workloads improved by 59.7% and 67.5% respectively, and derivative workloads by 34.3% and 36.4%. The geometric-mean mpmath/Wolfram value ratio moved from 1.43x before this change to 0.42x after it, with no measurable change in the external accuracy results.

### Wolfram

These are warmed, fixed-period-matrix, varying-argument value workloads. Process startup and output formatting are excluded from the timed region.

| Precision | Geometric mean ratio | Interpretation |
| ---: | ---: | --- |
| 30 digits | 0.46x | mpmath was about 2.2 times faster |
| 100 digits | 0.38x | mpmath was about 2.6 times faster |
| Combined | **0.42x** | mpmath was about 2.4 times faster |

The individual ratios ranged from 0.03x to 1.19x. For already well-damped matrices—those whose imaginary part makes successive lattice terms decay quickly—the two implementations were approximately level in genus two and three. On deliberately unreduced matrices, where the same function is presented in a basis that would otherwise produce a much longer lattice sum, mpmath's automatic partial reduction was especially effective. The bounded suite excludes 100-digit genus-three composite numerical derivatives; these are outside the value-performance headline.

### Validation and performance against FLINT

FLINT provides a state-of-the-art C implementation with rigorous complex-ball error bounds, designed for much higher precisions and a broader algorithmic scope than this contribution. It is therefore both a strong accuracy oracle and a demanding performance reference for mpmath's Python implementation.

| Workload | 30 digits | 50 digits | Combined |
| --- | ---: | ---: | ---: |
| Pointwise values | 1.69x | 2.44x | 2.03x |
| Pointwise derivatives | 1.60x | 1.50x | **1.55x** |
| Mixed second and third derivatives | 1.21x | 1.12x | **1.16x** |
| All pointwise workloads | 1.65x | 1.99x | **1.81x** |
| Repeated fixed-`tau`, varying-`z` workloads | 1.97x | 2.44x | **2.19x** |

Within the targeted low-genus range, mpmath comes particularly close to FLINT for higher derivatives: the combined geometric mean for mixed second and third derivatives is 1.16x, with lower mpmath timings in 8 of the 16 pointwise rows. Across all pointwise workloads, FLINT's compiled and extensively optimized implementation leads by a geometric mean factor of 1.81x.

The public interfaces perform somewhat different amounts of work: FLINT computes all half-integer characteristics together, whereas mpmath evaluates a selected characteristic. The table compares the natural public call in each library rather than attempting to normalize their internal workloads.

### Genus one: mpmath `jtheta`

The first call for a period matrix prepares reduction and lattice data that can be reused at the active precision. Repeated calls with fixed `tau`, as in plots and trajectories for integrable systems, therefore show the most relevant comparison with `jtheta`.

| Repeated 50-point sweep | 15 digits | 50 digits | 100 digits |
| --- | ---: | ---: | ---: |
| Values | 1.08x | 1.14x | 1.15x |
| First and second derivatives | 0.96x | 1.01x | 0.98x |

Across the broader pointwise suite, warmed-call geometric means were 1.17x, 1.22x, and 1.21x at 15, 50, and 100 digits. A first call, including preparation of reusable multidimensional data, measured 8.53x, 8.41x, and 7.63x respectively.

`rtheta` leads in some generic and transformed cases, while the specialized `jtheta` series retains an advantage when its nome is already small. This pull request keeps the mature `jtheta` implementation unchanged and adds a complementary multidimensional API.

### Sage implementations

The Sage comparison covers values, derivatives, and jets in genera two and three. RiemannTheta is a strong specialized comparator: its Cython MPFR/MPC kernel reuses a prepared `RiemannTheta(tau)` object and shares work between requested derivatives. The package also provides a standalone multiprecision `siegel_reduction` routine, although `RiemannTheta(tau)` does not invoke it automatically. The benchmark compares direct evaluation of the same `tau` and `z`: RiemannTheta construction is excluded, while mpmath uses its warmed cache and may transparently apply a cost-selected reduction. “Coupled” cases are already inexpensive to sum; “unreduced” cases exercise this difference in the public evaluation workflows rather than comparing the two reduction algorithms directly.

The following ratios are mpmath elapsed time divided by RiemannTheta elapsed time, so values below one favour mpmath.

Within the low-genus, 30--100-digit range targeted here, RiemannTheta is arguably the fastest specialized evaluator after comparable explicit Siegel reduction. FLINT has a broader rigorous design intended to extend to much higher precision, where it would be the more natural performance reference. These same-input direct-call results are included to show that this mpmath PR remains competitive as a pure-Python implementation, not to claim overall superiority.

| Precision | Scope | Geometric mean | mpmath wins |
| --- | --- | ---: | ---: |
| 30 digits | Matched direct calls | 1.19x | 9 / 20 |
| 50 digits | Matched direct calls | **0.78x** | 10 / 20 |
| 100 digits | Matched direct calls | **0.52x** | 10 / 20 |
| 30--100 digits | Matched direct calls | **0.79x** | 29 / 60 |
| 30--100 digits | Well-damped direct calls | 2.92x | 0 / 30 |
| 30--100 digits | Unreduced direct calls | **0.21x** | 29 / 30 |

The measured split is complementary: RiemannTheta's compiled evaluator is 2.92x faster in geometric mean on well-damped matrices, while mpmath's integrated reduction path makes its direct call 4.71x faster on the unreduced inputs. Across these 60 same-input direct calls, mpmath is 1.27x faster.

RiemannTheta 1.1.1 also confirms its arbitrary-precision design after [a precision-coercion issue](https://github.com/nbruin/RiemannTheta/issues/6) was fixed promptly. The original 333-bit reproducer now agrees with the defining series to about `1e-100`; the wider suite delivers 28--30 digits at 30 dps, 48--50 at 50 dps, and at least 100 at 100 dps, with no fixed-precision plateau.

Abelfunctions instead provides a fast machine-double NumPy/C path. At 15 digits, it led the well-damped subset by 1.57x, while mpmath led the unreduced subset by 2.44x and the combined comparison by 1.25x.

## Tests and sources

The dedicated test module exercises the public value and jet APIs together with the numerical machinery behind them. It covers all four classical genus-one characteristics, diagonal and block factorization, parity, quasi-periodicity, first through third derivatives, arbitrary real characteristics, modular transformations, cache behaviour, validation, and precision-doubling checks through 100 digits. Strongly damped genus-10 and genus-20 cases provide bounded smoke tests of the dimension-independent API.

Mathematical identities and transformation laws are cited directly to [DLMF Chapter 21](https://dlmf.nist.gov/21), while the lattice truncation starts from [Deconinck et al. (2004)](https://doi.org/10.1090/S0025-5718-03-01578-8). Independent numerical fixtures include Wolfram Engine 14.3 values generated at up to 120 digits and certified FLINT 3.6.0 derivative enclosures generated at 130 digits. The implementations and test suites of [abelfunctions](https://github.com/abelfunctions/abelfunctions), [RiemannTheta](https://github.com/nbruin/RiemannTheta), and [FLINT](https://flintlib.org/doc/acb_theta.html) informed the review of useful coverage, but the tests and benchmark drivers here were written independently for mpmath.

Run the focused suite with:

```bash
python -m pytest mpmath/tests/test_riemann_theta.py
```

## Reproducing the headline comparisons

The two self-contained scripts below can be copied directly from this document and saved in the root of an mpmath checkout. Run them with the Python interpreter used for mpmath development. Timings reported above were collected on macOS; absolute timings and some ratios will vary by processor and Python build.

<details>
<summary>Embedded benchmark: <code>rtheta</code> versus <code>jtheta</code></summary>

Save as `benchmark_rtheta_jtheta.py` and run:

```bash
python benchmark_rtheta_jtheta.py
```

This covers all four genus-one half-characteristics, first and second derivatives, reusable fixed-`tau` calls, first calls, and 50-point fixed-`tau` sweeps. Ratios below one favour `rtheta`.

```python
from statistics import geometric_mean, median
from timeit import repeat

from mpmath import mp


TAUS = (
    ("comfortable", "0.20", "1.20"),
    ("generic", "0.35", "0.75"),
    ("near-boundary", "0.20", "0.30"),
    ("strongly-unreduced", "0.17", "0.003"),
)
POINTS = (
    ("0", "0"), ("0.11", "0.03"), ("-0.23", "0.07"),
    ("0.37", "-0.09"), ("-0.41", "-0.12"), ("0.52", "0.16"),
)
WORKLOADS = (
    ("theta1", 1, 0, ((0.5,), (0.5,)), -1),
    ("theta2", 2, 0, ((0.5,), (0,)), 1),
    ("theta3", 3, 0, ((0,), (0,)), 1),
    ("theta4", 4, 0, ((0,), (0.5,)), 1),
    ("theta3 d1", 3, 1, ((0,), (0,)), 1),
    ("theta3 d2", 3, 2, ((0,), (0,)), 1),
)


def clear_rtheta_caches():
    mp._rtheta_tau_data.cache_clear()
    mp._rtheta_reduction_data.cache_clear()
    mp._rtheta_derivative_radius.cache_clear()


def time_per_call(function, calls, samples):
    return median(repeat(function, number=1, repeat=samples)) / calls


def sweep_points(count=50):
    golden = (mp.sqrt(5) - 1) / 2
    return tuple(
        mp.mpf("0.1") + mp.mpf("0.8") * (k + mp.mpf("0.5")) / count
        + mp.j * mp.mpf("0.1") * mp.frac((k + mp.mpf("0.5")) * golden)
        for k in range(count)
    )


for dps in (15, 50, 100):
    with mp.workdps(dps):
        warm_ratios = []
        first_ratios = []
        sweep_value_ratios = []
        sweep_derivative_ratios = []
        largest_error = mp.zero

        for unused_name, tau_real, tau_imag in TAUS:
            tau = mp.mpc(tau_real, tau_imag)
            tau_matrix = [[tau]]
            q = mp.expjpi(tau)
            points = tuple(mp.mpc(real, imag) for real, imag in POINTS)

            for unused_label, index, derivative, char, sign in WORKLOADS:
                def rtheta_batch():
                    for w in points:
                        mp.rtheta([w / mp.pi], tau_matrix, char,
                                  derivative=derivative)

                def first_rtheta_batch():
                    for w in points:
                        clear_rtheta_caches()
                        mp.rtheta([w / mp.pi], tau_matrix, char,
                                  derivative=derivative)

                def jtheta_batch():
                    for w in points:
                        sign * mp.pi ** derivative * mp.jtheta(
                            index, w, q, derivative=derivative)

                rtheta_batch()
                rt = time_per_call(rtheta_batch, len(points), 5)
                cold = time_per_call(first_rtheta_batch, len(points), 5)
                jt = time_per_call(jtheta_batch, len(points), 5)
                warm_ratios.append(rt / jt)
                first_ratios.append(cold / jt)

                with mp.extradps(30):
                    reference_q = mp.expjpi(mp.mpc(tau_real, tau_imag))
                    references = [
                        sign * mp.pi ** derivative * mp.jtheta(
                            index, mp.mpc(real, imag), reference_q,
                            derivative=derivative,
                        )
                        for real, imag in POINTS
                    ]
                actual = [
                    mp.rtheta([w / mp.pi], tau_matrix, char,
                              derivative=derivative)
                    for w in points
                ]
                largest_error = max(
                    largest_error,
                    *(abs(a - b) / max(1, abs(b))
                      for a, b in zip(actual, references)),
                )

                sweep = sweep_points()
                rtheta_sweep = tuple([w / mp.pi] for w in sweep)

                def rtheta_sweep_batch():
                    for z in rtheta_sweep:
                        mp.rtheta(z, tau_matrix, char,
                                  derivative=derivative)

                def jtheta_sweep_batch():
                    for w in sweep:
                        sign * mp.pi ** derivative * mp.jtheta(
                            index, w, q, derivative=derivative)

                rtheta_sweep_batch()
                ratio = (
                    time_per_call(rtheta_sweep_batch, len(sweep), 3)
                    / time_per_call(jtheta_sweep_batch, len(sweep), 3)
                )
                target = (sweep_derivative_ratios if derivative
                          else sweep_value_ratios)
                target.append(ratio)

        print(
            dps,
            "warm", f"{geometric_mean(warm_ratios):.2f}x",
            "first", f"{geometric_mean(first_ratios):.2f}x",
            "sweep-values", f"{geometric_mean(sweep_value_ratios):.2f}x",
            "sweep-derivatives",
            f"{geometric_mean(sweep_derivative_ratios):.2f}x",
            "max-error", mp.nstr(largest_error, 8),
        )
```

</details>

<details>
<summary>Embedded benchmark: <code>rtheta</code> versus Wolfram <code>SiegelTheta</code></summary>

This comparison requires Wolfram Engine. Set `WOLFRAM_KERNEL` to the local kernel executable, save the script as `benchmark_rtheta_wolfram.py`, and run:

```bash
WOLFRAM_KERNEL=/path/to/WolframKernel python benchmark_rtheta_wolfram.py
```

Each row runs in an isolated kernel with a 30-second timeout. Wolfram's `RepeatedTiming` excludes kernel startup and result formatting from the measured time. The script covers the value workloads underlying the headline geometric means, including zero, half-integer, and real characteristics. Ratios below one favour mpmath.

```python
import ast
import os
import re
import statistics
import subprocess
from timeit import repeat

from mpmath import mp


def n(real, imag="0"):
    return real, imag


POINTS = (
    (n("0"), n("0"), n("0")),
    (n("0.11", "0.03"), n("-0.07", "0.02"), n("0.05", "-0.01")),
    (n("-0.18", "0.06"), n("0.13", "-0.04"), n("-0.09", "0.02")),
    (n("0.27", "-0.08"), n("-0.16", "0.05"), n("0.12", "0.04")),
)
CASES = (
    ("g1", ((n("0.20", "1.20"),),), ("zero", "half"), POINTS),
    ("g2-diagonal", ((n("0.20", "1"), n("0")),
                     (n("0"), n("-0.15", "1.20"))),
     ("zero", "half"), POINTS),
    ("g2-coupled", ((n("0.10", "0.90"), n("-0.20", "0.12")),
                    (n("-0.20", "0.12"), n("0.05", "1.10"))),
     ("zero", "half"), POINTS),
    ("g3-diagonal", ((n("0.10", "0.90"), n("0"), n("0")),
                     (n("0"), n("-0.20", "1.10"), n("0")),
                     (n("0"), n("0"), n("0.05", "1.25"))),
     ("zero", "half"), POINTS),
    ("g3-coupled", ((n("0.10", "1"), n("-0.12", "0.08"), n("0.05", "0.04")),
                    (n("-0.12", "0.08"), n("0.20", "1.15"), n("0.08", "0.06")),
                    (n("0.05", "0.04"), n("0.08", "0.06"), n("-0.15", "0.90"))),
     ("zero", "half"), POINTS),
    ("g2-near-boundary", ((n("0.20", "0.92"), n("-0.12", "0.06")),
                          (n("-0.12", "0.06"), n("-0.18", "1.08"))),
     ("zero", "half"), POINTS),
    ("g2-unreduced", ((n("0.20", "0.15"), n("0.12", "0.03")),
                      (n("0.12", "0.03"), n("-0.10", "0.20"))),
     ("zero", "half", "real"),
     ((n("0.13", "0.07"), n("-0.21", "0.04")),
      (n("2.31", "-1.14"), n("-1.72", "0.83")))),
    ("g3-unreduced", ((n("0.30", "0.12"), n("0.17", "0.03"), n("0", "0.02")),
                      (n("0.17", "0.03"), n("-0.20", "0.16"), n("0", "0.02")),
                      (n("0", "0.02"), n("0", "0.02"), n("0.10", "0.20"))),
     ("zero", "half", "real"), POINTS),
    ("g3-block", ((n("0.15", "0.85"), n("0"), n("0")),
                  (n("0"), n("0.10", "0.95"), n("-0.07", "0.04")),
                  (n("0"), n("-0.07", "0.04"), n("-0.12", "1.15"))),
     ("zero", "half"), POINTS),
)


def characteristic(genus, kind):
    if kind == "zero":
        return None
    if kind == "half":
        return ([0.5 if k % 2 == 0 else 0 for k in range(genus)],
                [0.5 if k == genus - 1 else 0 for k in range(genus)])
    return ([0.3, -0.2, 0.1][:genus], [0.1, 0.4, -0.15][:genus])


def mp_number(value):
    return mp.mpc(mp.mpf(value[0]), mp.mpf(value[1]))


def wl_number(value, digits):
    real, imag = value
    r = f"Rationalize[{real}`{digits},0]"
    if imag == "0":
        return r
    return f"({r}+I Rationalize[{imag}`{digits},0])"


def wl_vector(values, digits):
    return "{" + ",".join(wl_number(x, digits) for x in values) + "}"


def wl_matrix(rows, digits):
    return "{" + ",".join(wl_vector(row, digits) for row in rows) + "}"


def wl_characteristic(char, digits):
    if char is None:
        return None
    return "{" + ",".join(
        "{" + ",".join(f"{x}`{digits}" for x in part) + "}"
        for part in char
    ) + "}"


def geometric(values):
    return mp.exp(mp.fsum(mp.log(x) for x in values) / len(values))


kernel = os.environ["WOLFRAM_KERNEL"]
rows = []
for digits in (30, 100):
    mp.dps = digits
    for name, tau_data, kinds, point_data in CASES:
        genus = len(tau_data)
        for kind in kinds:
            char = characteristic(genus, kind)
            tau_code = wl_matrix(tau_data, digits)
            zs_code = "{" + ",".join(
                wl_vector(point[:genus], digits) for point in point_data
            ) + "}"
            if char is None:
                function = "SiegelTheta[tau,#]&"
            else:
                function = (
                    f"SiegelTheta[{wl_characteristic(char, digits)},tau,#]&"
                )
            code = (
                f'tau=N[{tau_code},{digits}];zs=N[{zs_code},{digits}];'
                f'f={function};values=N[f/@zs,{digits}];f/@zs;'
                'elapsed=First@RepeatedTiming[Scan[f,zs],.2]/Length[zs];'
                'Print[StringRiffle[Join[{ToString[elapsed,InputForm]},'
                '(ToString[#,InputForm]&/@Flatten[{{Re[#],Im[#]}&/@values}])],'
                '"\\t"]];Exit[]'
            )
            try:
                process = subprocess.run(
                    [kernel, "-noprompt", "-run", code],
                    text=True, capture_output=True, check=True, timeout=30,
                )
            except subprocess.TimeoutExpired:
                print(digits, name, kind, "timed out")
                continue
            output = process.stdout.strip()
            if output.startswith('"'):
                output = ast.literal_eval(output)
            fields = output.split("\t")
            wolfram_time = float(fields[0].replace("*^", "e"))
            wolfram_values = [
                mp.mpf(re.sub(r"`[^*]*", "", x).replace("*^", "e"))
                for x in fields[1:]
            ]

            tau = [[mp_number(x) for x in row] for row in tau_data]
            zs = [[mp_number(x) for x in point[:genus]]
                  for point in point_data]

            def evaluate():
                for z in zs:
                    mp.rtheta(z, tau, char)

            evaluate()
            mpmath_time = statistics.median(
                repeat(evaluate, number=1, repeat=3)
            ) / len(zs)
            actual = [mp.rtheta(z, tau, char) for z in zs]
            expected = [
                mp.mpc(wolfram_values[2*k], wolfram_values[2*k + 1])
                for k in range(len(zs))
            ]
            error = max(
                abs(a - b) / max(1, abs(b))
                for a, b in zip(actual, expected)
            )
            rows.append((digits, name, kind,
                         mpmath_time / wolfram_time, error))
            print(digits, name, kind,
                  f"{mpmath_time / wolfram_time:.2f}x", mp.nstr(error, 6))

for digits in (30, 100):
    selected = [row for row in rows if row[0] == digits]
    print(digits, "geometric mean",
          mp.nstr(geometric([row[3] for row in selected]), 4),
          "largest error", mp.nstr(max(row[4] for row in selected), 6))
print("combined geometric mean",
      mp.nstr(geometric([row[3] for row in rows]), 4))
```

The Wolfram process timeout is intentionally short. Increase it only if a known-slow case is worth pursuing; an incomplete row is preferable to an unbounded benchmark run.

</details>

## AI use declaration

OpenAI Codex was used extensively to assist with reviewing literature, implementation, benchmark design and execution. I designed several experiments, reviewed profile outputs and numerous benchmarks, examined readability of code, setup environments and experiments with several alternative tools, and iterated and prompted extensively.


skirpichev (commented):
This will take some time to review.  Few observations so far.

It still looks as AI-generated code for me.  In some places it seems you are using tuples instead of matrices and reinvent existing mpmath capabilities, e.g. ctx.norm().

Some parts, probably, could be factored out to simplify review and provide some new capabilities for the mpmath.  For example, LLL reduction.  Or polynomial manipulation (see #142).

Edit:
Perhaps for genus=1 you should fallback to jtheta?

HeskethGD:
Just in case you were waiting for my edits here, I plan to address above once the LLL has been reviewed so that I can work it in.

PR #1183, opened 2026-09-15
Increase wp for mpf_exp() to return correctly rounded results

### LINKED ISSUE #1181: iv.exp can return an interval that does not contain exp(x) (enclosure guarantee violated)
`iv.exp` can return an interval whose upper endpoint is below the true `exp(x)`, so the true value is not contained. The docs state the guarantee is `f(v) ⊆ f̂(v)`, but for some exact dyadic inputs the returned interval fails to enclose `exp(x)`.

The upper endpoint is computed by `mpi_exp` via `mpf_exp(sb, prec, round_ceiling)` (`mpmath/libmp/libmpi.py`), which is intended to round outward (up). But `mpf_exp` is not correctly rounded: when the true `exp` lies just above a representable grid point, the ceiling-rounded approximation can still land on that lower grid point, one ulp below where a correct outward-rounded ceiling would be — and just below the true value.

## Reproducer (standard library only)

```python
from decimal import Decimal, getcontext
from mpmath import iv, mp, mpf
getcontext().prec = 80

def upper_exact(Y):                      # exact Decimal of the interval's upper endpoint
    sign, man, exp, bc = Y._mpi_[1]      # mpf tuple = (-1)**sign * man * 2**exp
    v = Decimal(man) * (Decimal(2) ** exp)
    return -v if sign else v

mp.prec = 60
for num, den, p in [(3893924827815215, 1125899906842624, 53),
                    (1381846513289487, 70368744177664, 113)]:
    iv.prec = p
    x = Decimal(num) / Decimal(den)      # x is an exact dyadic (den is a power of two)
    true_exp = x.exp()                   # correctly rounded to 80 digits
    Y = iv.exp(iv.mpf(mpf(num) / mpf(den)))
    ub = upper_exact(Y)
    print(f"x = {num}/{den}  (iv.prec={p})")
    print(f"  iv.exp(x) upper bound = {ub}")
    print(f"  true exp(x)           = {true_exp}")
    print(f"  upper >= true ?         {ub >= true_exp}   <-- must be True for a valid enclosure")
    print(f"  true - upper          = {true_exp - ub}\n")
```

Output:

```
x = 3893924827815215/1125899906842624  (iv.prec=53)
  iv.exp(x) upper bound = 31.769286826103925847064601839520037174224853515625
  true exp(x)           = 31.769286826103925847314551452083032333640559839473033909976919555310968064082700
  upper >= true ?         False   <-- must be True for a valid enclosure
  true - upper          = 2.499...E-19

x = 1381846513289487/70368744177664  (iv.prec=113)
  iv.exp(x) upper bound = 337548593.639065733181444072201225156049566...
  true exp(x)           = 337548593.639065733181444072201225156056764...
  upper >= true ?         False
  true - upper          = 7.197...E-30
```

`decimal.exp` is correctly rounded, so `true_exp` is the true value to well within the reported gaps. I also confirmed the true value independently with Arb (via `python-flint`, rigorous ball arithmetic), which agrees.

## Scope

- Reproduces on 1.4.1 and on `master` (`1.5.0a2.dev35+g972270f01`).
- Independent of `mp.prec` (checked 53/113/237/500) — the endpoints here are exactly representable at these precisions.
- Rare: a scan of 1416 integer point inputs across `sin/cos/exp/log/sqrt/tan` × 4 precisions found no under-enclosure; these two dyadic inputs to `exp` are the counterexamples I have. The magnitude is small — the true value sits well under one ulp above the returned endpoint — but the interval still fails to contain it.
- I understand interval support is documented as experimental; filing because it is a specific, mechanism-pinpointed counterexample to the stated enclosure guarantee rather than a general "iv is imprecise" report.

## Two possible resolutions

1. Add the special function's own error bound into the directed rounding, so the ceiling endpoint is provably ≥ the true value (i.e., widen by the guaranteed `mpf_exp` error before rounding up).
2. If tight enclosure of transcendentals is out of scope, document that `iv` transcendental functions are not guaranteed to enclose to the last representable grid point.

Happy to provide the search generator or more counterexamples if useful.

skirpichev:
Closes #1181


repowazdogz-droid:
At p=53, `prec+25` still returns `1-2**-53` for `exp(-2**-53)` under `round_up` and `round_ceiling`; the correct result is 1.

```python
from mpmath.libmp import mpf_exp, round_up, round_ceiling
x = (1, 1, -53, 1); expected = (0, 1, 0, 1)
for rnd in (round_up, round_ceiling):
    observed = mpf_exp(x, 53, rnd)
    print(rnd, "observed:", observed, "expected:", expected)
```

A separate defect is in `mpf_pow_int`: a negative exponent rounds the positive power at `prec+5` before reciprocal. `mpf_exp` reaches this path for integer x when p>600. Retained failures are x=-172 at p=601 and p=602, plus x=-1000 at p=2000. Boundary controls -172/p600, -172/p603, -171/p601, and -999/p2000 pass. #1183's wider outer precision masks these `mpf_exp` cases, but direct `mpf_pow_int` callers retain the path.

Changing `prec+5` to `prec+14` fixes six/380 power comparisons and adds no disagreements over 2,435 frozen inputs × five modes. Timings are flat except a repeatable +16% at -172/p2000. This is empirical, not a sufficiency proof. I can push either tests-only or tests-plus-guard onto #1183, or follow up, as preferred; all counterexamples were checked against an independent rigorous enclosure.

Reproducers: https://gist.github.com/repowazdogz-droid/6ee497ca5e18378406830b7fa2f45dc9

skirpichev:
> At p=53, prec+25 still returns 1-2**-53 for exp(-2**-53) under round_up and round_ceiling; the correct result is 1.

Indeed, now I'm less sure if this is a right approach (you could increase wp by 60 to "fix" that case).

PR #1185, opened 2026-09-17
Add LLL reduction

HeskethGD:
# Add `lll_gram` for lattice reduction from Gram matrices

Adds a public LLL reduction function for real symmetric positive-definite Gram matrices. This extracts and develops the lattice-reduction work from the [`rtheta` PR (#1182)](https://github.com/mpmath/mpmath/pull/1182) into a reusable linear-algebra operation.

This function is needed for computing reduced period matrices in the `rtheta` implementation, and is also useful for lattice-based applications in cryptography, computational number theory, and optimization.

## API

```python
lll_gram(Y, delta=0.75, *, maxsteps=10000)
```

Returns an integer unimodular transformation `U` with determinant +1 such that `U.T * Y * U` is LLL-reduced. For a basis `B` stored in columns, the reduced basis is `B * U`.

- Supports `mp` and `fp` contexts.
- Returns a tuple of tuples of Python integers, preserving arbitrarily large coefficients without rounding.
- Requires `1/4 < delta < 1` and enforces the size-reduction bound `eta = 0.5`.
- Invalid inputs raise `ValueError`; exhaustion of reduction work limits raises `NoConvergence`.

The reduction guarantees apply to the values after input conversion into the chosen context. Converting the returned integers to an mpmath matrix makes subsequent arithmetic subject to working precision.

The implementation combines fraction-free exact reduction for small inputs with numerical reduction and exact output verification. Method selection is internal. Documentation is under **Matrices → Linear algebra → Lattice reduction**.

## Benchmarks

Geometric means across nine integer lattices: six generated cases of dimensions 2–24 and three shear cases. Each timing is the median of seven calls after warmup; engine startup, input construction and independent output validation are excluded. mpmath's internal verification is included. Measurements use Python 3.14.7 on macOS ARM64, SageMath 10.8 and Wolfram Engine 14.3.

| Implementation                      | Geometric-mean time |
| ----------------------------------- | ------------------: |
| mpmath `mp`, 15 digits, GMP         |            0.257 ms |
| mpmath `mp`, 15 digits, pure Python |            0.261 ms |
| mpmath `fp`, GMP                    |            0.205 ms |
| mpmath `fp`, pure Python            |            0.207 ms |
| Sage `LLL_gram` / PARI              |            0.085 ms |
| Sage basis LLL / exact NTL          |            0.094 ms |
| Wolfram `LatticeReduce`             |            0.112 ms |

mpmath uses `(delta, eta) = (0.99, 0.5)` here. Sage's Gram interface uses `(0.99, 0.51)`; the NTL comparison uses `delta=0.99`, strict size reduction and returns the transformation. Wolfram does not expose these parameters in its documented API. Sage Gram and mpmath receive `B * B.T`; Sage NTL and Wolfram receive the row basis `B`. These compare related public operations, not identical algorithms. Reference-tool timings above are from the GMP run; disabling GMP affects mpmath only.

A separate comparison of numerical-only, exact-only and automatic selection covers **387 inputs per backend**, including eight inputs per dimension/coefficient-size combination and theta-related matrices at 15, 50 and 100 digits. Automatic selection gives geometric-mean speedups of **1.06× / 1.16×** on dense inputs and **1.51× / 2.16×** on the existing theta matrices, with/without GMP respectively. Some dense inputs regress; the largest observed dispatcher slowdowns were 1.33× and 1.42×. These are LLL timings, not end-to-end `rtheta` speedups.

<details>
<summary>Standalone mpmath / Wolfram timing comparison</summary>

Save as `compare_lll_wolfram.py` and run from the checkout with an environment containing this implementation. Requires `wolframscript` on `PATH` and an activated Wolfram Engine. Set `WolframKernel` if needed for your installation.

```sh
python compare_lll_wolfram.py
MPMATH_NOGMPY=1 python compare_lll_wolfram.py
```

This timing example recreates the nine workloads above. The full benchmark additionally verifies lattice preservation and reduction conditions independently.

```python
import json
import random
import statistics
import subprocess
import tempfile
import time
from pathlib import Path

from mpmath import fp, mp
from mpmath.libmp.backend import BACKEND

cases = []
for n in (2, 4, 8, 12, 16, 24):
    rng = random.Random(1182 + n)
    B = [[rng.randint(8, 20) if i == j else
          rng.randint(-5, 5) if j > i else 0
          for j in range(n)] for i in range(n)]
    for _ in range(2*n):
        i, j = rng.sample(range(n), 2)
        q = rng.choice((-2, -1, 1, 2))
        B[i] = [x + q*y for x, y in zip(B[i], B[j])]
    cases.append({"name": f"random-{n}", "basis": B})
for n, shear in ((3, 1000), (5, 1000), (8, 10)):
    B = [[int(i == j) + (shear if i == j + 1 else 0)
          for j in range(n)] for i in range(n)]
    cases.append({"name": f"shear-{n}-{shear}", "basis": B})

with tempfile.TemporaryDirectory() as directory:
    payload = Path(directory) / "cases.json"
    payload.write_text(json.dumps(cases))
    worker = Path(directory) / "benchmark.wls"
    worker.write_text('cases = Import[' + json.dumps(str(payload)) +
                      ', "RawJSON"];\n' + r'''
rows = Map[Function[c, Module[{b = c["basis"], samples, result},
  LatticeReduce[b];
  samples = Table[
    First[AbsoluteTiming[result = LatticeReduce[b];]], {7}];
  <|"name" -> c["name"], "seconds" -> Median[samples]|>
]], cases];
Print[ExportString[rows, "RawJSON", "Compact" -> True]];
''')
    result = subprocess.run(
        ["wolframscript", "-file", str(worker)],
        check=True, capture_output=True, text=True, timeout=120)
    wolfram = {r["name"]: r["seconds"] for r in json.loads(result.stdout)}

mp.dps = 15
samples_by_engine = {"mp": [], "fp": [], "Wolfram": []}
print("mpmath backend:", BACKEND)
print("case                 mp ms     fp ms    Wolfram ms")
for case in cases:
    B = case["basis"]
    G = [[sum(x*y for x, y in zip(a, b)) for b in B] for a in B]
    times = []
    for label, ctx in (("mp", mp), ("fp", fp)):
        matrix = ctx.matrix(G)
        delta = ctx.mpf("0.99")
        ctx.lll_gram(matrix, delta=delta)
        measurements = []
        for _ in range(7):
            start = time.perf_counter()
            ctx.lll_gram(matrix, delta=delta)
            measurements.append(time.perf_counter() - start)
        elapsed = statistics.median(measurements)
        samples_by_engine[label].append(elapsed)
        times.append(1000*elapsed)
    samples_by_engine["Wolfram"].append(wolfram[case["name"]])
    times.append(1000*wolfram[case["name"]])
    print(f'{case["name"]:20}', *(f'{x:10.3f}' for x in times))
print(f'{"Geometric mean":20}', *(
    f'{1000*statistics.geometric_mean(values):10.3f}'
    for values in samples_by_engine.values()))
```

</details>

## Validation

- 245 focused tests passed, with one expected failure; 211 checks passed with the pure-Python backend. LLL branch coverage is 100%.
- Stress testing verified 1,258 reductions across both backends, plus 20 invalid-input and four work-limit checks. Cases include dimensions through 64, extreme binary scales and 7,001-bit transformation entries.
- All 2,322 method/input combinations in the expanded benchmark passed independent exact verification.
- Lint and HTML documentation checks passed.

## AI Declaration

AI tools (Codex and Claude) assisted with implementation, tests, benchmarking and this draft.


mpmath/matrices/lattice.py
```diff
@@ -0,0 +1,376 @@
...
+        **References**
+
+        A. K. Lenstra, H. W. Lenstra Jr. and L. Lovasz,
+        "Factoring polynomials with rational coefficients", Mathematische
+        Annalen 261 (1982), 515-534.
+        """
+        try:
+            maxsteps = index(maxsteps)
+        except TypeError:
+            raise ValueError("maxsteps must be a positive integer") from None
```
skirpichev: Why you need to catch this exception?

mpmath/matrices/lattice.py
```diff
@@ -0,0 +1,376 @@
...
+        if maxsteps <= 0:
+            raise ValueError("maxsteps must be a positive integer")
+        try:
+            delta = ctx.convert(delta)
```
skirpichev: You could use ctx.mpf, right?

mpmath/matrices/lattice.py
```diff
@@ -0,0 +1,376 @@
...
+            if not ctx.isfinite(delta) or ctx.im(delta):
+                raise ValueError
+            delta = delta.real
+            if not ctx.mpf('0.25') < delta < 1:
```
skirpichev: And then just check if delta is in range, no need for try-except block:

```suggestion
            if not 0.25 < delta < 1:
```

mpmath/matrices/lattice.py
```diff
@@ -0,0 +1,376 @@
...
+                raise ValueError
+        except (TypeError, ValueError):
+            raise ValueError("delta must be real and satisfy 0.25 < delta < 1") from None
+        if isinstance(Y, (list, tuple)) and any(
+                not isinstance(row, (list, tuple)) or len(row) != len(Y) for row in Y):
+            raise ValueError("Y must be a nonempty square matrix")
+        try:
+            Y = ctx.matrix(Y)
```
skirpichev: Why just not use ctx.matrix(Y)?  That will handle list/tuple input.

mpmath/matrices/lattice.py
```diff
@@ -0,0 +1,376 @@
...
+        except (TypeError, ValueError, IndexError):
+            raise ValueError("Y must be a nonempty square matrix") from None
+        n = Y.rows
+        if not n or n != Y.cols:
+            raise ValueError("Y must be a nonempty square matrix")
+        if any(not ctx.isfinite(x) or ctx.im(x) for x in Y):
+            raise ValueError("Y entries must be finite and real")
+        # Avoid ctx.re here: unary operations can round an existing mpf.
```
skirpichev: Assuming you are using ctx.matrix above, you *already* triggered rounding in the matrix constructor.

mpmath/matrices/lattice.py
```diff
@@ -0,0 +1,376 @@
...
+# Algorithm-specific helpers stay with their owning implementation below.
+
+
+class LatticeMethods:
+    def lll_gram(ctx, Y, delta=0.75, *, maxsteps=10000):
+        r"""
+        Return an LLL-reducing change of basis for the Gram matrix *Y*.
+
+        *Y* must be a nonempty, real, symmetric positive-definite matrix or
+        nested sequence. Symmetry is required exactly, rather than to a
```
skirpichev: You already told above, that *Y* is a Gram matrix.  Then it's real and symmetric.  Only new thing is that it's required to be a positive-definite, not semi-positive.  (Or you can relax this requirement?)

PR #1188, opened 2026-09-22
Fix randomized fft test failure

baranwalayush:
- adjusted abs_eps from the default (~1.77e-15) to 1e-14.
- 1e-14 provides a safe margin over the worst-case numerical noise observed (~ 2.51e-15).
- set the `deadline=None` for the `test_fft_randomized_complex()`


skirpichev:
Apparently you didn't take look on test failure.

skirpichev:
Please take look on the test failure.

baranwalayush:
The root cause of the test failure in master is Hypothesis generated complex values with extremely small components such as 1e-294, processing those become slow and hits the 600-second pytest timeout. 
- Added the `deadline=None` settings for this test in the last commit. This should fix the error.

Apologies for the previous assumption, though that is needed too. Now, this new failure here is something related to mpmath version, will look into it.

baranwalayush:
All good now, these changed should fix the failure.

skirpichev:
I think it could be related to the PyPy version.  I see no problem on the previous stable release locally.  You can try v8.0.

baranwalayush:
Ok, I believe this is because of the pytest-timeout. In pyproject.toml, the timelimit is set to 600 seconds (10 minutes), so even if hyothesis doesn't raise error because of the @settings(deadline=None) decorator, this pytest-timout could still fail. 
So, I guess we can either reduce the number of examples from the default value of 100 to something like 10 or 20 or if there is an option to increase the pytest-timeout for this specific case, we can do that too.

skirpichev:
> Ok, I believe this is because of the pytest-timeout.

Good.  Next time, please start from reading logs.

> I guess we can either reduce the number of examples from the default value of 100 to something like 10 or 20 or if there is an option to increase the pytest-timeout for this specific case

The timeout value 600 (10 minutes) is already a lot.  The problem is not pytest-timeout, but your new code: now this test looks to be more than 10x slower on PyPy with gmp and ~5x slower on CPython 3.14.

Let me know if you are interested to finish this pr.

baranwalayush:
> Good. Next time, please start from reading logs.

sure

> > I guess we can either reduce the number of examples from the default value of 100 to something like 10 or 20 or if there is an option to increase the pytest-timeout for this specific case
> 
> The timeout value 600 (10 minutes) is already a lot. The problem is not pytest-timeout, but your new code: now this test looks to be more than 10x slower on PyPy with gmp and ~5x slower on CPython 3.14.

true, 10 minutes should be enough. But as I pointed in the earlier comment that hypothesis is generating random complex values with extremely small components such as 1e-294, and I guess processing these are expensive.

> Let me know if you are interested to finish this pr.

I'll try some workaround and let you know when I find some solution. I'd also welcome if you have any suggestion on fixing this.

